PR #431, opened 2015-02-21
Fixing multi-labels for Adam: fixture object

azaroth42:
Added fixture 40, with multiple labels in metadata. Used as validation test for the validator.


azaroth42:
Fixed multiple sequences to actually have the external sequence! @zimeon

PR #497, opened 2015-07-07
470 semver

### LINKED ISSUE #470: [Meta] Clarify understanding of SemVer requirements
From #425, we need to be clearer in our understanding of the Semantic Versioning requirements, with respect to what is a backwards incompatible (breaking) change.

In particular, the question has arisen as to whether a new requirement at a minor level is a breaking change, because a server that was (e.g.) level 2 compliant and does not implement the new requirement, suddenly is not level 2 compliant.  This may be an unintended side effect of levels only being versioned at the major level.

azaroth42:
Closes #470 

(Can go straight to master as SemVer and Notes are not subject to versioning)


source/api/annex/notes/scheduled-changes.md
```diff
@@ -0,0 +1,19 @@
...
+
+# Introduction
+
+This document records scheduled changes that are intended to be made in the future, but must wait for new major releases as they're backwards incompatible.
+
+# Image API
+
+* Require support for the `square` region at compliance level 1.
+* Clarify that height and width are for calculating aspect ratio only, as there is no 'source' image.
+
```
jpstroop: I wonder if it's worth adding links to the relevant Github issues for each of the bullets? I recognize that this is a deeper view into our process than we'd probably want to publish most of the time, but in this case there's a conversation/rationale that supports each item and I think that should be surfaced somehow.

azaroth42: I considered it.  Transparency is good, getting to be part of the process is good ... but not sure if this is too much sausage making? We can put them in, and take them out if it's too weird ... or we can put them in with a subsequent PR.

Issue #478, opened 2015-06-08
Deprecate pct: syntax


jpstroop:
What does it get us? You have to get the info.json to know the size to calculate the percentage anyway, and clients may use different decimal precision which is bad for caches. Could note that is is deprecated in 2.1 for exclusion in 3.0

azaroth42:
Discuss in Princeton at Eds' meeting.

azaroth42:
Eds agree. Deprecation warnings added to spec in 2.1; @jpstroop to send email to iiif-discuss to seek objections.

jpstroop:
https://groups.google.com/d/msg/iiif-discuss/CkYj3vjse5Y/Qo6MQ9swr28J

azaroth42:
For 10/7 call, I propose to remove deprecation warnings. The use cases in the discussion linked are convincing to me at least :)

zimeon:
Editors agree: remove deprecation warning from 2.1, revisit for 3.0 based on additional experience

jpstroop:
Note also that the size `pct` syntax is required for level 1, but the region `pct` syntax is only required for level 2. I'd support making it all level2 or greater.

robcast:
I think we should split this issue as using `pct` for size or for region are very different issues.

Using `pct` for size does not seem to make much sense to me as I have to know the size of the image on the server to know what size I'll get. If I know the size already I can just do the multiplication myself and use an absolute size.

Using `pct` for a region is an important feature because it is the only way to specify an image region without knowing the absolute image size. 

I think having a relative measure like `pct` is important for image annotations or other uses where a region gets saved separate from the original image because with the relative measure the region can be reconstructed even if the image resolution changes. (Another way would be to save the size of the original image with the reference... I'll need to talk to the annotation people about that idea.)

jpstroop:
@IIIF/editors propose: move all `pct` requirements to level 2.

jpstroop:
Closed with f89824768c917d4956f8460a21cf1d764cac4f00

zimeon:
Noting use of `pct:` in Wikidata: https://www.wikidata.org/wiki/Property:P2677

PR #567, opened 2015-10-08
closes #566; closes #565; closes #564; closes #560; closes #556; ...

### LINKED ISSUE #556: Clarify context link header for info.json
That it should be returned whenever the application/json media type is used, not when the ld+json media type is requested in Accept.

### LINKED ISSUE #560: Clarify canonical syntax for square
pixel region or just "square"?

"square" is not as precise as "full", so even though both are shorthand for something that could be expressed as pixels, the fact that it's up to the server to decide what square you'll get makes me feel like it should be the pixel region--this also means that if you don't like the square you got, you _could_ adjust it without getting the info.json.

### LINKED ISSUE #564: Kill Image API section 8.1
[It's just 8.](http://image-21.iiif.io/api/image/2.1/#introduction-1)

### LINKED ISSUE #565: URI template parameters table should mention both host and port?
Maybe the table that lists some of the parameters for the URI template should mention that the server parameter can include both host and port? Small clarification.

https://github.com/IIIF/iiif.io/blob/ca22d7fd669e7bc1ef4c56e24d829eb58255ac9b/source/api/image/2.0/index.md#2-uri-syntax

### LINKED ISSUE #566: Use full URI template instead of base_url parameter
It would be clearer if the spec used the full URI template instead of using a "base_url" parameter in: "The full URL for an image with “default” quality in “jpg” format would be: {base_url}/{identifier}/full/{width},{height}/0/default.jpg"

This appears to be the only place base_url is used and it does not seem to be defined anywhere.

https://github.com/IIIF/iiif.io/blob/ca22d7fd669e7bc1ef4c56e24d829eb58255ac9b/source/api/image/2.0/index.md#51-image-information

azaroth42:
Also removes deprecation warning -- defers #478

### DIFF
--- a/source/api/image/2.1/index.md
+++ b/source/api/image/2.1/index.md
@@ -7,7 +7,7 @@ tags: [specifications, image-api]
 major: 2
 minor: 1
 patch: 0
-pre: final
+pre: draft2
 ---
 
 ## Status of this Document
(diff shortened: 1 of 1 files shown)


source/api/image/2.1/index.md
```diff
@@ -813,6 +794,7 @@ Early sanity checking of URIs (lengths, trailing GET, invalid characters, out-of
   * For use cases that enable the saving of the image, it is _RECOMMENDED_ to use the HTTP `Content-Disposition` header ([RFC6266][rfc-6266]) to provide a convenient filename that distinguishes the image, based on the identifier and parameters provided.
   * This specification makes no assertion about the rights status of requested images or any other descriptive metadata, whether or not authentication has been accomplished. Please see the [IIIF Presentation API][prezi-api] for rights and other information.
   * Additional [Apache HTTP Server implementation notes][apache-notes] are available.
+  * Linked data implementations may construct the info.json response using the frame supplied in the [JSON-LD framing implementation note][annex-frames].
```
jpstroop: :+1:

PR #573, opened 2015-10-13
Add Universal Viewer event to Ghent for Friday

tomcrane:
There is no Eventbrite mini-event for this. I'm not sure if there needs to be or if it's part of the Unconference.


azaroth42:
Let's make a mini-event for it. Not sure what will happen with people who have already registered tho? I guess they can make changes?  Should we wait to merge for that update?

tomcrane:
Can we synchronise (ish) merging/depolying with updating the EventBrite listing? Is that possible? I guess the eventBrite should be live before this PR.

zimeon:
It seems that this is not listed in the eventbrite https://iiif-winter-2015.eventbrite.com/ so shouldn't merge yet. Who is on the hook for the eventbrite update??

blalbrit:
Eventbrite is updated. Please confirm that it looks good and then merge away.

zimeon:
There is no "more info" link on the UV entry on the eventbrite menu but that isn't a showstopper. The details are still there lower on the page. Will merge.

zimeon:
Looks good: <http://iiif.io/event/2015/ghent.html#universal-viewer-and-manifesto>

blalbrit:
Thanks for catching that "more info" bit - fixed.

zimeon:
'tis beautiful @blalbrit . All done I think.

PR #574, opened 2015-10-15
[DO NOT MERGE] image-21 to master?

jpstroop:
_In theory_ nothing here should break existing specs, and at this point it seems odd to be pointing an increasing number of people to a branch. 

Better, I think, to merge this in, and add the w3c-style WIP warning at the top, [similar to what we've done with auth](http://iiif.io/api/auth/0.9/).

If we get a quorum of :+1:s I volunteer to do the merge / rebase and add the warning.


jpstroop:
Oops, had the branches reversed...I'll come back

PR #575, opened 2015-10-15
[DO NOT MERGE] Image 21

jpstroop:
For real this time... _In theory_ nothing here should break existing specs, and at this point it seems odd to be pointing an increasing number of people to a branch. 

Better, I think, to merge this in, and add the w3c-style WIP warning at the top, [similar to what we've done with auth](http://iiif.io/api/auth/0.9/).

If we get a quorum of :+1:s I volunteer to do the merge / rebase and add the warning.


zimeon:
I agree that as soon as we have draft/beta versions that get announced on iiif-discusss for general comment, they should be up on the main site with the large beta warning (in the same way as [auth](http://iiif.io/api/auth/0.9/)). So :+1:

jpstroop:
Any more :+1: s?

azaroth42:
source/api/image/2.1/authentication.md and its referenced images should be
deleted, no?

jpstroop:
Yup, that would be a part of the merge.

There are some other Image-21 branches (`-clean`, `-524`). Are those anything? Can they be wiped out?

jpstroop:
Closing in favor of #576

Issue #579, opened 2015-10-23
Move Framing/Compaction to implementation note


azaroth42:
The last paragraph of 4.2 in Prezi 2.1 discusses @graph, which can trivially be removed using JSON-LD compaction and/or framing.  This is an implementation detail and thus should be in an implementation note, rather than the main text of the spec.

zimeon:
Change to remove discussion of @graph looks good

Issue #578, opened 2015-10-23
Prezi recommends context link header incorrectly


azaroth42:
As per image discussion, the presentation API should not recommend returning the context in the link header, as it MUST be ignored anyway due to potential for conflict with @context in the representation.

Proposal: Delete from "If the regular " down to the end of the following example from section 4.2

zimeon:
Corresponding change in image api was discussed in #556 in changed commit 87571a8. Agree change for this issue is good.

Issue #554, opened 2015-09-16
Clarify @context can be array


azaroth42:
To allow extensions to actually be semantic, rather than implementations dropping the information when they're not resolved by the context.  The specification currently suggests that it needs to be in the encapsulating object, but this doesn't take into account that the extension may be at the Manifest level.

Action is to update the contexts section with more coherent description:
http://iiif.io/api/presentation/2.0/#linked-data-context-and-extensions

zimeon:
Example of array at end of 4.3.5 is missing closing square bracket, and closing brace is italics

zimeon:
Otherwise, :+1: for text

azaroth42:
Fixed ] in https://github.com/IIIF/iiif.io/commit/97a24e64784265f623911e7f956db0f2b8ed61f6

zimeon:
:+1:

Issue #541, opened 2015-08-03
Clarify 8.2 regarding retrieval of Layers


azaroth42:
More than one sentence needed regarding how to dereference Layers.

azaroth42:
in the same way as a they are  --> delete "a"

zimeon:
Eds meeting: agree good (modulo minor typo)

zimeon:
Typo fixed, :+1:

PR #580, opened 2015-10-23
Prezi 21 clean

azaroth42:
Add Presentation 2.1 to master after fixing all the issues.


zimeon:
Have gone though #579, #578, #554, #541, #540, #539, #537, #533 and I think this is ready to merge... which I see @jpstroop just did

Issue #524, opened 2015-07-27
Update image service example to use auth


azaroth42:
... once we have a more solid auth spec.

azaroth42:
Include merging 8.1 into 8. unless there's additional structure needed.

zimeon:
Perhaps greatest clarity would be to add another service example, using the auth service, into section 8. 

+1 to merging 8.1 into 8

jpstroop:
+1

azaroth42:
Blocked by #570.

PR #583, opened 2015-10-23
Add @context; Closes #570

### LINKED ISSUE #570: Auth spec does not mention @context
The Auth spec can be used for both image and presentation metadata, and hence is its own separate API.  However it doesn't have any mention of @context.  Surely it needs it?

I was going to fix #524 but I think we need to discuss.

azaroth42:
Unblocks #524 but pushes auth to 0.9.1


azaroth42:
Only needed in login, closing to fix.

PR #584, opened 2015-10-23
Auth context

azaroth42:
Re-opening after fixing text


azaroth42:
Not happy with process, need a change-log etc.

PR #586, opened 2015-10-23
Remove draft dates except for current date, closes #585

### LINKED ISSUE #585: Drafts should not have entries in change-logs
As they meaninglessly expose internal milestones, rather than actual specification delivery dates.  They should be deleted out of image and presentation API changelog appendices.

For currently draft specs, the most recent date can be added but only the final date should be present by the end of the process.

zimeon:


azaroth42:
:+1:

Issue #591, opened 2015-11-02
Recommend against fragment URIs for Canvas


azaroth42:
As we recommend using #xywh= to identify a region of a Canvas, we should recommend against fragment URIs for Canvases.  Otherwise, you would end up with nonsense like:

`http://example.org/iiif/book1/manifest.json#canvas1#xywh=1,2,3,4`

azaroth42:
Closing, in master :)

zimeon:
Change in 9269382 looks good, agree done

PR #596, opened 2015-11-09
Closes #593, #591

### LINKED ISSUE #593: Clarify not(collections, manifests) is discouraged
An unintended interpretation of the discouragement for completely empty collections was that a collection without collections was discouraged, which is not the intent.  The sentence in 2.1's section 7.3 should be clarified.

azaroth42:


source/api/presentation/2.1/index.md
```diff
@@ -1227,7 +1227,7 @@ collections
 manifests
 : References to manifests contained within the current collection. Each referenced manifest _MUST_ have the appropriate @id, @type and label.
 
-At least one of `collections` and `manifests` _SHOULD_ be present.  Empty collections with neither are allowed but discouraged.
+At least one of `collections` and `manifests` _SHOULD_ be present in the response.  Responses with neither of `collections` or `manifests`, describing collections that have no members, are allowed but discouraged.
```
jpstroop: "Responses with neither of" -> "Responses without ...". Maybe "i.e.," before "describing collections ... "?

source/api/presentation/2.1/index.md
```diff
@@ -629,7 +629,7 @@ Recommended URI pattern:
 ```
 {: .urltemplate}
 
-The canvas represents an individual page or view and acts as a central point for laying out the different content resources that make up the display. Canvases _MUST_ be identified by a URI and it _MUST_ be an HTTP(s) URI. If following the recommended URI pattern, the {name} parameter _MUST_ uniquely distinguish the canvas from all other canvases in the object. As with sequences, the name _SHOULD NOT_ begin with a number. Suggested patterns are "f1r" or "p1".
+The canvas represents an individual page or view and acts as a central point for laying out the different content resources that make up the display. Canvases _MUST_ be identified by a URI and it _MUST_ be an HTTP(s) URI. If following the recommended URI pattern, the {name} parameter _MUST_ uniquely distinguish the canvas from all other canvases in the object. As with sequences, the name _SHOULD NOT_ begin with a number. Suggested patterns are "f1r" or "p1". The URI of the canvas _SHOULD NOT_ contain a fragment (a `#` followed by further characters), as this would make it impossible to refer to segment of the canvas's area using the `#xywh=` syntax.
```
jpstroop: Do issues w/ # URIs surface elsewhere, warranting a move of this new _SHOULD NOT_ sentence to a more general area of the spec?

Also, "to refer to segment" -> "to refer to __a__ segment"

source/api/presentation/2.1/index.md
tomcrane: "An empty collection, with neither `collections` or `manifests`, is allowed but discouraged."

source/api/presentation/2.1/index.md
azaroth42: Anywhere that a fragment should be used to describe a segment of a resource, it will conflict with a use to identify a semantic construct.  I think that's only Canvas though?

jpstroop: I think you're right. Just the typo then.

source/api/presentation/2.1/index.md
zimeon: +1 to @tomcrane's formulation

jpstroop: :+1:

azaroth42:
Line notes addressed in latest commit, GTG?

azaroth42:
Ping?

PR #605, opened 2015-11-12
Add Cogapp and Qatar National Library to the institutions.yml file

andyc-cogapp:
We are actively using IIIF and part of the bi-weekly calls, we'd like to be listed. We helped the Qatar National Library implement IIIF for their digital library http://www.qdl.qa/en.


azaroth42:
:+1:

PR #607, opened 2015-11-12
Warning to explain the sizes issue, fixes #544

### LINKED ISSUE #544: Make description of `sizes` compatible with canonical URI syntax
The current description of [canonical URI syntax](http://image-21.iiif.io/api/image/2.1/#canonical-uri-syntax) says:

Parameter | Canonical value
--- | ---
size | “full” if the default size is requested, the w, syntax for images that should be scaled maintaining the aspect ratio, and the w,h syntax for explicit sizes that change the aspect ratio.

and the description of `sizes` in [info.json](http://image-21.iiif.io/api/image/2.1/#technical-properties) says:

Technical Property | Required? | Description
--- | --- | ---
sizes | Optional | A set of height and width pairs the client should use in the size parameter to request complete images at different sizes that the server has available. This may be used to let a client know the sizes that are available when the server does not support requests for arbitrary sizes, or simply as a hint that requesting an image of this size may result in a faster response. _A request constructed with the w,h syntax using these sizes must be supported by the server_, even if arbitrary width and height are not.

I see two issues:

  1. Saying that _a request constructed with the w,h syntax using these sizes must be supported by the server_ for the w,h given in `sizes` is at odds with the idea of using `w,` for aspect ratio preserving sizes in the canonical URI syntax
  2. I think we have assumed that `sizes` is for aspect ratio preserving sizes but we don't say that. If that is the intention then it should be explicit, if not then example should probably include a distorting w,h.

My feeling is that for 2) we should allow distorting w,h, but that to address 1) we should say that for aspect ratio preserving sizes, only the width parameter should be given. And thus the requirement for support of requests is that _A request constructed using these sizes, with the w, syntax where width alone is given or w,h syntax where width and height are given, must be supported by the server_

zimeon:
I guess I should actually make the PR... see #544 for discussion

### DIFF
--- a/source/api/image/2.1/index.md
+++ b/source/api/image/2.1/index.md
@@ -491,6 +491,10 @@ A recipe for enabling these behaviors is provided in the [Apache HTTP Server Imp
 
 The objects in the `sizes` list have the properties in the following table. Images requested using these sizes _SHOULD_ have a region parameter of "full" and rotation of "0".  The size _SHOULD_ be requested using the canonical syntax of `w,`. Thus, the full URL for an image with "default" quality in "jpg" format would be: `{scheme}://{server}/{prefix}/{identifier}/full/{width},/0/default.jpg`
 
+__Warning__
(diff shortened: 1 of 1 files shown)


azaroth42:
:+1: PR implements the decision from the call, shipping it.

PR #612, opened 2015-11-24
Ghent update

azaroth42:
Update agenda, add security announcement:
  http://ghent_update.iiif.io/event/2015/ghent.html


azaroth42:
Gracias :)

PR #616, opened 2015-11-30
Ghent schedule

azaroth42:
http://ghent_schedule.iiif.io/event/2015/ghent.html

Optimistic schedule adding Matt and Emmanuelle, links to restaurants and sched.org.


azaroth42:
Will re open after getting new name from Emmanuelle.

PR #622, opened 2015-12-03
Prezi issues

### LINKED ISSUE #610: Clarify non text/plain formats okay for annotations
The only format mentioned in the spec is text/plain. It should be clearer that this is not a requirement, and text/html text/markdown or any other is possible. A recommendation for text/plain or text/html might be useful, however?

From: https://github.com/IIIF/mirador/issues/585

### LINKED ISSUE #611: Hotspot linking
A use case from National Library of Norway: There are areas in the canvas that should be linked to either other canvases or external documents.  For example, linking from "continued on page 5" to the region on page 5 where the article is continued, or linking a reference to something to a description of it.

Open Annotation has a "linking" motivation that seems appropriate and could be added as a subsection in section 6 of prezi/2.1

```
{
  "@id":"http://www.example.org/iiif/book1/annotation/anno1",
  "@type":"oa:Annotation",
  "motivation":"oa:linking",
  "resource": {
    "@id":"http://www.example.org/page-to-go-to.html",
    "@type":"dctypes:Text",
    "format":"text/html"
  },
  "on":"http://www.example.org/iiif/book1/canvas/p1#xywh=500,500,150,30"
}
```

azaroth42:
Closes #611 ; Closes #610

### DIFF
--- a/source/api/presentation/2.1/index.md
+++ b/source/api/presentation/2.1/index.md
@@ -1070,6 +1070,26 @@ Other resources may also have comments made about them, including Manifests (com
 }
 {% endhighlight %}
 
+
+### 6.8 Hotspot Linking
+
+It is also possible to use Annotations to create links between resources, by using a `motivation` of "oa:linking". The region of the canvas that, which clicked, should trigger the link is specified in the `on` field in the same way as other Annotations, and the resource that is linked to it is in the `resource` field.  The linked resource _MAY_ also be another canvas or region of a canvas.  The user experience of whether the linked resource is opened in a new tab, new window or by replacing the current view is up to the implementation.
+
+{% highlight json %}
+{
+  "@context": "http://iiif.io/api/presentation/2/context.json",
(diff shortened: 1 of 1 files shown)


jpstroop:
I think there are sufficient :+1:s from @mikeapp, @zimeon, and me in the issues. :ok_hand: :question:

jpstroop:
Can I get one more +1 to merge?

jpstroop:
@IIIF/editors can we get +1s or comments on the two open PRs before the meeting tomorrow?

jpstroop:
This is going stale...

zimeon:
:+1:

jpstroop:
@azaroth42 I think this is one of the branches that had it's history rewritten following mergegate? I'll merge if you want, but you should be on hand in case the site breaks.

mikeapp:
:+1:

jpstroop:
Seems to have gone :ok:

PR #635, opened 2015-12-08
no more variables in contexts

### LINKED ISSUE #633: Do not use variable interpolation for contexts
Old ones should hang around; we don't want them to change when we change the major version (someday).

jpstroop:
Closes #633

### DIFF
--- a/source/api/annex/openannotation/context.json
+++ b/source/api/annex/openannotation/context.json
@@ -1,13 +1,26 @@
----
----
 {
-	"@context" : {
-		"iiif":    "http://iiif.io/api/image/{{ site.image_api.latest.major}}#",
+  "@context": {
+    "iiif": "http://iiif.io/api/image/2#",
 
-		"region" : {"@id" : "iiif:regionValue", "@type": "xsd:string"},
-		"size" : {"@id" : "iiif:sizeValue", "@type": "xsd:string"},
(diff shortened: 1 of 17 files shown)


jpstroop:
I did this w/ grep, sed, and `python -m json.tool` to make sure the results were still well-formed json (and because the files had a mix of tabs and spaces and formatting was generally less nice than it could be). 

Hope the reformatting is :ok:

azaroth42:
I find the one key/value per line much harder to read for the context docs, personally, as you can't just scan the lines for the key. The values are all similar so the indentation of the value makes the consistency clearer. But I accept that the PR is more consistent.  

Overall +0 from me.

zimeon:
I share Rob's view that one-per-line formatting of the definitions is better. However, I think using standardized formatting available automatically from `python -m json.tool ` trumps this, and few people will actually have to read these documents. So :+1:

jpstroop:
That's my feeling too...I'd rather it be tidy in some way that can be automated than having to do it all by hand. @tomcrane @mikeapp ?

tomcrane:
I do read the presentation context doc quite a lot, usually by following the link at the top of the nearest manifest - but Chrome formats it it to look like the new version anyway, so I don't mind this change, :+1:

mikeapp:
:+1: I find this easily readable, so don't have any objection.

jpstroop:
that's (more than) enough +1s to merge if someone wants to pull the trigger

PR #654, opened 2015-12-13
clarify rendering to support linking to a viewer

### LINKED ISSUE #648: Should there be a property in image and prezi APIs to link to a preferred viewer instance?
This question came up in discovery discussions at [Ghent](http://iiif.io/event/2015/ghent.html) that we might have a registry allowing discovery of sets of IIIF resources via either IIIF Collections or via a sitemap of HTML pages that have microdata (#557) linking to the IIIF resource. In the second case the discovery system will already know the "preferred" access page which might be presented to discover system users (along with drag and drop options) which may give a particularly good UX tailored to the resource, and likely make the host institution happy for extra traffic. If starting from an IIIF Collection and finding Manifests or Image Services, there is currently no way to find an access point.

tomcrane:
Resolves #648 (if not required in Image API too)

### DIFF
--- a/source/api/presentation/2.1/index.md
+++ b/source/api/presentation/2.1/index.md
@@ -282,7 +282,7 @@ related
     * Any resource _MAY_ have one or more external resources related to it.
 
 rendering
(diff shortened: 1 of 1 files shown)


jpstroop:
Change "must" in "...rendering resource must be supplied..." back to a big RFC _MUST_ (:tm:) and then I'm :+1:.

azaroth42:
:+1: to text and need for __MUST__

azaroth42:
Fixed the __MUST__, good to go from me.

jpstroop:
:+1: after a rebase.

jpstroop:
:+1: @zimeon @mikeapp ?

mikeapp:
:+1:

zimeon:
:+1:

PR #655, opened 2015-12-13
Add context for auth; closes #637

### LINKED ISSUE #637: Publish auth context
We reference one in the spec, but it doesn't resolve.

azaroth42:
I minted an auth NS, and included image and prezi NS even though they're not needed directly.
Otherwise the only three keys needed are label, service and profile.

### DIFF
--- a/source/api/auth/0/context.json
+++ b/source/api/auth/0/context.json
@@ -1,0 +1,22 @@
+{
+  "@context": {
+    "iiif": "http://iiif.io/api/image/2#",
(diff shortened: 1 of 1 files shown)


jpstroop:
:+1:

zimeon:
:+1: , JSON validates ;-)

azaroth42:
I think we can merge without Tom or Mike as it's non normative (otherwise it would be a breaking change). Merge if you agree :)

PR #656, opened 2015-12-14
Range layer link

### LINKED ISSUE #645: Allow link between Range and Layer
Some content types, such as newspapers, have both physical and very different logical/textual structures.  Articles can start on some part of the first page (canvas1) and continue on some part of the fifth page (canvas5).  

Presentation Use Cases:
  * A viewer might wish to display all of the OCR'd / transcribed text of the article, regardless of which canvas is in the primary view. 
  * A viewer might wish to synchronize the display of the image with the text that is being scrolled, and hence jump from canvas1 to canvas5
  * For articles that are part of the same canvas, but split between columns, the viewer may wish to synchronize the display and refocus within the same canvas when the article text is scrolled across the columns
  * For export, it would be valuable to associate the metadata of the article (range1) and the content (the annotations), for example in an export to PDF

Modeling requirements:

A page is a canvas, an article is a range (that can span canvases), and the content is linked via annotations.  Those annotations can be in one or more annotation lists.

Issue:

There is currently no way to associate the content (annotations) with the article (range), other than a lot of math to calculate the overlap of all annotations that target the canvas with the area of the canvas that is part of the range.  This is not necessarily actually correct, as there may be commentary or other annotations that are NOT part of the article. It would be impossibly expensive to do this in the client at run time.

Proposal:

Allow a link between the Range that represents the article, and a Layer that collects together one or more AnnotationLists, where each AnnotationList contains the annotations that target a particular canvas, or contiguous region within the canvas.  This would allow the UA to follow the link from the article to the layer to the annotationlists to the annotations without complicated math. 

Thus, we add a range specific relationship (name TBD) to point to the layer.

newspaper-iiif-model

azaroth42:
Closes #645

### DIFF
--- a/source/api/presentation/2.1/index.md
+++ b/source/api/presentation/2.1/index.md
@@ -321,6 +321,14 @@ startCanvas
     * A sequence or a range _MAY_ have exactly one canvas as its start canvas.
     * Other resources _MUST NOT_ have a start canvas.
 
+contentLayer
+:   A link from a range to a layer that includes the annotations of content resources for that range.  Clients might use this to present content to the user from a different canvas when interacting with the range, or to jump to the next part of the range within the same canvas.  
+
+    Usage:
+    {: .usage}
+    * A range _MAY_ have exactly one layer as its content layer.
+    * Other resources _MUST NOT_ have a content layer.
+
 ####  3.5. Paging Properties
 
 first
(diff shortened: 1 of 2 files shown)


azaroth42:
Note that merging will add the contentLayer property to the live, current context even though it's defined in the future version 2.1.

Not sure how our process can deal with this otherwise though? Suggestions welcome.

jpstroop:
Hmmm...if we're not breaking the context, I don't think it's a problem.

mikeapp:
Agree, for now not a problem.

zimeon:
:+1:, also agree that addition to context is harmless

azaroth42:
Normative text change, so need :+1:s from everyone please.

jpstroop:
:+1:

mikeapp:
:+1:

jpstroop:
> Normative text change, so need :+1:s from everyone please.

According to the [Editoral Process draft](http://editors.iiif.io/api/annex/notes/editors.html#acceptance-criteria-for-merging-changes) we need all but one, so pulling the trigger.

tomcrane:
(would have been a +1)

PR #657, opened 2015-12-14
Change to dcterms:rights; closes #636

### LINKED ISSUE #636: Change "license" to map to dct:rights
@kestlund, @anarchivist, @aisaac 
e.g. rightsstatements.org

azaroth42:
If we merge this PR, then the live, current context will change. 
Confirm agreement that this doesn't need to be delayed until 2.1 is released please.

### DIFF
--- a/source/api/presentation/2/context.json
+++ b/source/api/presentation/2/context.json
@@ -16,7 +16,7 @@
     "as": "http://www.w3.org/ns/activitystreams#",
     "license": {
       "@type": "@id",
-      "@id": "dcterms:license"
+      "@id": "dcterms:rights"
     },
     "service": {
       "@type": "@id",


jpstroop:
This is the right thing to do, even though it breaks our rules. I don't like it, but practically speaking, it makes sense and no one will likely notice.

jpstroop:
Shrugs. Should there at least be a note to IIIF discuss?

azaroth42:
Note to IIIF-Discuss seems good course of action. :+1:s on PR with a note describing the change?

tomcrane:
:+1: , and maybe tweak the description in the Presentation API spec to convey that viewer implementers don't have to use the label "license"  when they display the link to a user; that label could be "license/rights" or something.

zimeon:
:+1: to going ahead. I think we are invoking the _in practice_ clause in semver... if a change is technically breaking but nobody is using the feature that is broken, the change is not breaking. Also agree on a note to IIIF-Discuss and a change to suggest the use or _rights_ or _license/rights_ in UI.

mikeapp:
:+1: Agreed, and on note to IIIF-Discuss and suggested label change.

PR #670, opened 2015-12-17
Add mapping for rendering, navdate; closes #666

### LINKED ISSUE #666: rendering and navDate need predicates
The new 2.1 properties `navDate` and `rendering` do not currently have RDF predicates and hence can't be in the context until they do.  If they're in our own namespace (hopefully not) then we also need to include them in the ontology.

I think that `rendering` could be `dcterms:hasFormat`:

> A related resource that is substantially the same as the pre-existing described resource, but in another format.

The semantics of navDate (being not semantic but just about navigation) I think mean we should probably just include it in shared canvas. Otherwise dc:date is pretty vague and might cover it:

> A point or period of time associated with an event in the lifecycle of the resource.

azaroth42:


jpstroop:
:+1:

zimeon:
:+1:

mikeapp:
:+1:

tomcrane:
:+1:

PR #671, opened 2015-12-17
Add maxWidth/maxHeight/maxArea, clarify compliance

### LINKED ISSUE #620: Limiting max height and width that can be requested
See the [thread started by Glen yesterday](https://groups.google.com/forum/#!topic/iiif-discuss/j-tg5F7J6VM). We've heard it before, and I was about to add my own slightly different use case, but @tomcrane has already done so. I'm perfectly happy to give you `full` of anything, but from a practical, systems perspective, I can't do it because the full image is just too large (i.e. scrolls, and some other huge prints we have) and the server will time out.

Yes, practically speaking, if you can get the tiles you can recreate the whole image. However, we've heard enough times now that it would make people comfortable and further adoption if we could find a solution.

The cheap solution would be to add a predicate like `fullAvailable: true|false`, and force clients to work with `sizes` to determine what they can get. The better solution, in my opinion, is something like the discussion around #277.

zimeon:
Add maxWidth/maxHeight/maxArea, clarify compliance document regarding support for combinations of parameters, number compliance document sections following form of Image API, give Level 0 its own section. Fixes #620

### DIFF
--- a/source/api/image/2.1/compliance.md
+++ b/source/api/image/2.1/compliance.md
@@ -64,7 +64,7 @@ See also note under [Size][size] about combinations of Size and Region that may 
 |        | `sizeAboveFull`  | ![optional][icon-opt] | ![optional][icon-opt] | ![optional][icon-opt] |
 {: .api-table}
 
-Note that servers may express limits on the sizes available for an image with the optional `maxWidth`, `maxHeight` and/or `maxArea` [Profile Description properties][profile]. Servers are compliant provided they support the forms of the Size parameter shown above for image sizes up to the limits specified. Clients _SHOULD NOT_ assume that Region and Size parameter combinations such as `/full/full/` will be supported. 
+Note that servers may express limits on the sizes available for an image with the optional `maxWidth`, `maxHeight` and/or `maxArea` [Profile Description properties][profile]. Servers are compliant provided they support the forms of the Size parameter shown above for image sizes up to the limits specified. Clients should not assume that Region and Size parameter combinations such as `/full/full/` will be supported. 
 
 ### 3.3 Rotation
 
--- a/source/api/image/2.1/compliance.md
+++ b/source/api/image/2.1/compliance.md
@@ -14,12 +14,11 @@ pre: draft
 
 __This version:__ {{ page.major }}.{{ page.minor }}.{{ page.patch }}{% if page.pre != 'final' %}-{{ page.pre }}{% endif %}
 
-__Latest stable version:__ {{ site.image_api.latest.major }}.{{ site.image_api.latest.minor }}.{{ site.image_api.latest.patch }}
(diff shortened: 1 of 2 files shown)


zimeon:
Not sure what is up with the auto-website: http://maxwidth.iiif.io/api/image/2.1/ and http://maxwidth.iiif.io/api/image/2.1/compliance.html do not show the changes in this branch

zimeon:
Auto-website now happy, see http://maxwidth.iiif.io/api/image/2.1/ and http://maxwidth.iiif.io/api/image/2.1/compliance.html

azaroth42:
Can we add maxWidth (at least, as most common use case) to the example?

zimeon:
Good point @azaroth42, added `maxWidth` in profile example

azaroth42:
I think it's clear that if you specify both area and maxW/H that both apply, but is it worth making it explicit? Either way :+1: from me

source/api/image/2.1/compliance.md
```diff
@@ -46,21 +48,25 @@ In the tables below "![required][icon-req]" indicates that support is _REQUIRED_
...
+| ,h     | `sizeByH`        | ![optional][icon-opt] | ![required][icon-req] | ![required][icon-req] |
+| pct:x  | `sizeByPct`      | ![optional][icon-opt] | ![required][icon-req] | ![required][icon-req] |
+| w,h    | `sizeByForcedWh` | ![optional][icon-opt] | ![optional][icon-opt] | ![required][icon-req] |
+| !w,h   | `sizeByWh`       | ![optional][icon-opt] | ![optional][icon-opt] | ![required][icon-req] |
+|        | `sizeAboveFull`  | ![optional][icon-opt] | ![optional][icon-opt] | ![optional][icon-opt] |
 {: .api-table}
 
-### Rotation
+Note that servers may express limits on the sizes available for an image with the optional `maxWidth`, `maxHeight` and/or `maxArea` [Profile Description properties][profile]. Servers are compliant provided they support the forms of the Size parameter shown above for image sizes up to the limits specified. Clients _SHOULD NOT_ assume that Region and Size parameter combinations such as `/full/full/` will be supported. 
+
```
jpstroop: Small thing, but why SHOULD NOT as opposed to MUST NOT?

jpstroop: And should we be telling clients what to do at all? Maybe we leave the warning text in place but remove the RFC _SHOULD NOT_?

zimeon: I agree, I think the correct thing is to make this a plain English "should not"

jpstroop:
:+1:

azaroth42:
:+1: (post deRFCification) There's no test or warning possible, so SHOULD NOT bears no real weight.

zimeon:
@mikeapp @tomcrane - are we good on this one?

tomcrane:
:+1:

mikeapp:
:+1:

PR #672, opened 2015-12-18
Allow multiple logos in image and presentation; closes #664

### LINKED ISSUE #664: Multiple logos?
Currently Prezi says that: 

> Any resource may have exactly one logo associated with it.

Is there a use case for allowing multiple logos? How about the manifest that refers to canvases from multiple institutions, or has metadata contributions from multiple institutions?  Should we instead say one or more?

azaroth42:


azaroth42:
Closes #664, including for both image and presentation. Intentionally does not address choice of logo.

jpstroop:
:+1:

zimeon:
I'm nervous that we a cannibalizing the original use-case of having a logo displayed in order to support multiple. However, maybe we have to go this way.

I think the _NB_ portion should be an intro to the section and we need to soften the phrases like "Text that _MUST_ be shown when the resource it is associated with is displayed or used" because that really isn't the requirement anymore.

azaroth42:
@zimeon I think that is still the requirement, it's just that the content provider can require multiple attributions or logos be displayed. Happy to move the note to an intro.

mikeapp:
:+1:

tomcrane:
The mandatory display of rights properties has generated a fair amount of heat with more than one client, so maybe a slightly fuller introduction? If viewer implementers read the spec ( :) ) then they need something like this - the "client" might be a simple image carousel, or not much more than a basic OSD canvas:

The following properties ensure that the interests of the owning or publishing institutions are conveyed regardless of the viewing environment, and a client _MUST_ make these properties clearly available to the user. Given the wide variation of potential client user interfaces, it will not always be possible to display all or any of the properties to the user in the client's initial state. If initially hidden, the method of revealing them _MUST_ be obvious, such as a button or scroll bars.

tomcrane:
(I'll commit this if it's going in the right direction)

zimeon:
@tomcrane - I like your intro, I've been waiting to find time to suggest something like this. I think you manage to convey both the reason/importance and the understanding that it may be hidden initially.

jpstroop:
:+1: I like @tomcrane's formulation as well.

azaroth42:
:+1: from me too (4 = good to go)

On Mon, Dec 21, 2015 at 5:37 AM, Jon Stroop <notifications@github.com>
wrote:

> [image: :+1:] I like @tomcrane <https://github.com/tomcrane>'s
> formulation as well.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/IIIF/iiif.io/pull/672#issuecomment-166307792>.
>



-- 
Rob Sanderson
Information Standards Advocate
Digital Library Systems and Services
Stanford, CA 94305

tomcrane:
I had better commit it then... hold on a minute...

azaroth42:
@zimeon if you're good with the text, want to hit merge? :)

zimeon:
I guess the merge was a :+1: from me...

PR #673, opened 2015-12-21
Holiday logo and message

azaroth42:
Holiday themed IIIF :)


azaroth42:
Okay, magic is in place to swap back to the real values on site-rebuild after Jan 4th.

azaroth42:
Have tried many things, can't get it to look any better without massively reworking the CSS :(

azaroth42:
Closing, can't get the image to look good at the right size. At least learnt a trick for liquid variables :D

PR #679, opened 2015-12-22
Rename size/tile, closes #677

### LINKED ISSUE #677: Rename iiif:size in image context
In writing the JSON-LD implementation note, I realized we can get ourselves out of the inadvertent size expansion issue by simply renaming the RDF predicate `iiif:size` to `iiif:hasSize`.  As we don't consider the RDF mapping to be subject to semver (per the rights/license discussion), we can change it in 2.1

The issue:  `sizes` maps to `iiif:size`.  We have features like: `iiif:sizeByW`, so the compaction algorithm sees that if can make a short term by using those together:  `sizes:ByW` has the same expansion as `iiif:sizeByW` with fewer characters.

Thus by renaming to `iiif:hasSize` we avoid the unexpected compaction.  Note that the Ruby RDF library from @gkellogg has a workaround in it already to avoid this, but it's not part of the official JSON-LD 1.0 expansion algorithm.

azaroth42:
I don't believe we have this problem in Prezi. Changed tile as well as size for consistency.
Closes #677 (edit to make this link)

### DIFF
--- a/source/api/image/2/context.json
+++ b/source/api/image/2/context.json
@@ -103,11 +103,11 @@
       "@type": "xsd:string"
     },
     "sizes": {
-      "@id": "iiif:size",
+      "@id": "iiif:hasSize",
       "@type": "@id"
     },
     "tiles": {
-      "@id": "iiif:tile",
+      "@id": "iiif:hasTile",
       "@type": "@id"
     },
     "profile": {


jpstroop:
:ok: :+1:

mikeapp:
:+1:

tomcrane:
:+1:

PR #680, opened 2015-12-22
Jsonld note, closes #668

### LINKED ISSUE #668: service key collides with service URI scheme in JSON-LD expansion
__Issue:__

JSON-LD expansion is greedy when it comes to looking for terms in the context that match the beginning of a string followed by a colon.  Thus in the worst case, if we defined a key called 'http' with a definition of "http://example.com/steal/things/", then the value "http://iiif.io/" would be expanded to "http://example.com/steal/things///iiif.io/" :(

At the moment we are vulnerable to one such unintended expansion: service.  There is a URI scheme for services defined in RFC2609, and corresponding protocol in RFC2608. So if someone were to use a service uri it would be incorrectly expanded.
From the RFC, an example is:
`service:device-drivers:tftp://x2.bean.org/vol3/disk/drivers.drv;driver=scsi;platform=sys3.2-rs3000`

Which would expand to:
`http://rdfs.org/sioc/services#has_service:device-drivers:tftp://x2.bean.org/vol3/disk/drivers.drv;driver=scsi;platform=sys3.2-rs3000`

Not a big deal (clearly) but perhaps we should write an implementation note recording the issue?

__References:__
 * https://github.com/jasnell/w3c-socialwg-activitystreams/issues/273
 * http://tools.ietf.org/html/rfc2609
 * http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

azaroth42:
Closes #668 
Expands JSON-LD frames note out to all our JSON-LD issues, and adds python code to check contexts for the URI scheme expansion bug (tho not to Travis yet)


azaroth42:
Ping?

zimeon:
To save the next person having to look up the URL to check: <http://jsonld_note.iiif.io/api/annex/notes/jsonld.html>

zimeon:
Should the code sample have a working manifest URI so that it actually runs? Could use <http://iiif.io/api/presentation/2.0/example/fixtures/1/manifest.json> instead of <http://iiif.io/api/presentation/2/example/manifest1.json> (which is 404)

zimeon:
:+1: with fix to python example to have a good manifest URI

azaroth42:
Changed URI as suggested by @zimeon. GTG?

zimeon:
:+1: GTG, code passes the cut-paste-run test

mikeapp:
:+1:

tomcrane:
:+1:

jpstroop:
:+1:

PR #682, opened 2015-12-22
676 frontpage

### LINKED ISSUE #676: Replace front page demo with carousel of showcase tiles
The try it demo on the front page is horribly out of date (v1.1) and while it was great when IIIF was more technically focused, we've outgrown it as a community.

As a replacement, I propose a carousel of the showcase tiles.  We should also link to showcase.iiif.io in the header navigation bar, and solicit more tiles.

azaroth42:
Closes #676


jpstroop:
:+1: following a squash

tomcrane:
I have updated the UV info - but it depends on a PR on showcase.iiif.io as well.

mikeapp:
:+1: overall but would suggest adjusting the `wrapper` class width to 800px (i.e., exactly two tiles wide)

zimeon:
:+1: but agree with @mikeapp that exaclty two tiles displayed would be good

azaroth42:
Agreed, will fix tile width and squash

azaroth42:
Responsiveness fixed, squashed commits. GTG?

zimeon:
:+1: http://676_frontpage.iiif.io/ is good to go to me... responsive 1,2,3 tiles, yum

jpstroop:
:+1:

Issue #678, opened 2015-12-22
Drop "full" size 


jpstroop:
See https://github.com/IIIF/iiif.io/issues/663#issuecomment-166312822

azaroth42:
:+1: to killing full in 3.0

mikeapp:
:+1: to killing `full`

tomcrane:
:+1:

zimeon:
:+1: to deprecating `full` as of 2.1 and killing in 3.0

zimeon:
Discussion at The Hague, syntax will change from `/full/full/0/default` to '/full/max/0/default`, servers might well redirect anyway. Might be useful to note the history in the 3.0 spec

jpstroop:
Consensus on 31 Jan 2018 community call is that this is still OK. This + resolution of #693 solves #1370

azaroth42:
closed by #1422

stevewh:
How can one be assured to get the actual size of the original image so that bounding box coordinates (as well as cropping paths) can work with the original image file? 

This likely also speaks to the immutability issue of URLs, where a system is built with a dependency on a URL such that the content of the URL returns a constant value (in this case an image with immutable dimensions).

jpstroop:
If I understand the wish question: the width and height that reported in the `info.json` are there so that clients can do the math they need to do to do to make the appropriate request.

Also, there is no notion of an "original" image in the Image API's semantics, e.g. just because a server reports a specific width and height, sizes of tiles, etc, does not mean that the image exists (it could be generated dynamically) or the the server allows retrieval of the entire image--have a look at [the explanation of `max`](https://iiif.io/api/image/3.0/#42-size) which was introduced in 2.1.1 and will completely replace `full` in 3.0.

Regarding URIs and immutability, I believe the section on the [canonical syntax](https://iiif.io/api/image/3.0/#47-canonical-uri-syntax) should solve any concerns there?

jpstroop:
(apologies for typos, answering from my phone)

stevewh:
Thanks Jon for the quick response. Perhaps this thread needs a new issue or
is related to a different issue.

I have read these parts of the specification and understood the replacement
of full with max. I also understand the json.info request.

I have an image annotation web app and would like to conform to IIIF.
Images in the system can be stored as URLs which includes IIIF urls. The
URL is loaded into an <img> element with auto for height and width so that
the "full" image is loaded and the coordinates are captured. The system
[...]

PR #687, opened 2016-01-14
Add `max` size keyword, deprecate `full` size keyword

### LINKED ISSUE #663: When we have maxWidth/maxHeight/maxArea, should there be a way to ask for the largest size for a given region?
Separating this from #620, comments so far:

@tomcrane wrote: Given the discussion on what this issue means a server should do in response to a size parameter that it can't support, is there an argument for an additional size parameter "best" (or similar)? `/full/best/0/default.jpg` or `/100,100,10000,10000/best/0/default.jpg`.  This would have definite utility - the UV offers a "download high res" option that currently requires out of band knowledge of the server's abilities - all a viewer need do is ask for `/full/best/`... (or "max")

@azaroth42 wrote: Or ... "default" :wink:

@glenrobson wrote: Regarding @tomcrane comment on `/full/best/0/default.jpg` I would like to be able to support a URL that will blindly get the best image I can provide without working it out from the info.json. I thought it may be required for the EDM Web Resource id but in the end we used the thumbnail URL. It still may come up that someone would want to use this URL. ... For public domain images I would probably look at disabling the maxwidth/area restriction for these. I'm not sure how the discussions will go but it could be that we don't give the full resolution for these images and so a `/full/best/0/default.jpg` would be useful.

@mattmcgrattan wrote: +1 for a 'best' option.

@mikeapp wrote: best or max could return a size anywhere from `/full/` down to the tile size on the long edge, its going to be difficult to use in a consistent way. Would it be OK if the user clicked 'download high res' and got a 350x512 image? I think you'd need to check the info.json and not offer the option if the server doesn't support a reasonably high resolution download. Also, if someone codes a client to display images using /best/ in order to work around tile size limitations, it could have unfortunate consequences when hitting a server that can return a full-scale image. Given the proposed size restrictions I'd really like the option to indicate that I'll accept a smaller size, e.g., `/0,0,2000,2000/<1000,/` for 'up to 1000 wide' when creating links without loading info.json. But that would add even more complexity on the server side :frowning:

jpstroop:
:arrow_up: that. Advances #678, closes #663.

Seems like the deploy script is broken. Changes would be [visible here](http://mad_max.iiif.io/api/image/2.1/#size).


azaroth42:
I would prefer a warning slightly less definite in case we change our minds. Something like:

The size keyword `full` is likely to be removed in favor of only supporting `max` in Version 3.0 and thus should be considered as deprecated.

And we might want to define deprecated in our semantic versioning doc?

jpstroop:
I'd rather speak with a little more conviction. What do you think the of [this revision](http://mad_max.iiif.io/api/image/2.1/#full-dep)?

tomcrane:
I like that revision.

azaroth42:
I'm good with the revision, though maybe there's a meta point we can discuss about whether all such deprecations are signaled.  Will raise an issue

azaroth42:
To be clear:  :+1:

jpstroop:
Added another commit. What is the compliance level for `max` size? It will be 0 in 3.0, but until then? 

New features aren't breaking changes, so it's up to us. That said, I don't think we should raise the bar for getting to 2.1 too much, so I'd say either level 2 or completely optional, with a slight leaning toward level 2. This commit has it marked as completely optional which seemed safest.

azaroth42:
I have cold feet with all of the example changes.
--> issue

jpstroop:
@azaroth42 This is rebased against master and just waiting for an image that demonstrates `max` size.

jpstroop:
:+1: (it looks like the merge might be messy)

azaroth42:
:+1: to merge, needs everyone please. @zimeon @tomcrane @mikeapp

zimeon:
:+1:

mikeapp:
:+1:

azaroth42:
Rebased to allow merge.

jpstroop:
Hah, forgot it was my PR. I guess @azaroth42 and I should recuse ourselves. @tomcrane @zimeon @mikeapp ?

zimeon:
I'm still :+1:

jpstroop:
Various iterations have been :+1: by everyone, and the changes since have been non-normative examples. Pull the trigger?

PR #696, opened 2016-01-21
Add implementation note for slash parsing

### LINKED ISSUE #555: Solve the identifier with slashes problem
To allow implementations to parse from the right, and if prefix is known use the rest as the identifier.
Most implementations do it already somehow anyway, so we should bless and normalize the behavior.

azaroth42:
Closes #555.


zimeon:
:+1: to http://555_slashnote.iiif.io/api/image/2.1/#a-implementation-notes

PR #708, opened 2016-02-03
Link fixes - WIP, DO NOT MERGE UNTIL DEPLOY SCRIPT IS FIXED

jpstroop:
These were all the result of the Jekyll update. Will get a bunch o' redirects in place before merge.


jpstroop:
Also, only about 3/4 of the broken links were due to Jekyll; the tests hadn't been running properly for a long time. Adding `set -ev` to test.sh (so that the script exits as soon as any suite fails) was key.

azaroth42:
:+1:

PR #712, opened 2016-02-05
Adds a cache for external link checking

jpstroop:
Only check external links every two weeks; speeds up CI considerably.


azaroth42:
:+1:

PR #715, opened 2016-02-08
Add implementation note for slash parsing

azaroth42:
Rebased to new master

### DIFF
--- a/source/api/image/2.1/index.md
+++ b/source/api/image/2.1/index.md
@@ -812,7 +812,8 @@ Early sanity checking of URIs (lengths, trailing GET, invalid characters, out-of
 
 ### A. Implementation Notes
 
-  * For use cases that enable the saving of the image, it is recommended to use the HTTP `Content-Disposition` header ([RFC6266][rfc-6266]) to provide a convenient filename that distinguishes the image, based on the identifier and parameters provided.
+  * For use cases that enable the saving of the image, it is _RECOMMENDED_ to use the HTTP `Content-Disposition` header ([RFC6266][rfc-6266]) to provide a convenient filename that distinguishes the image, based on the identifier and parameters provided.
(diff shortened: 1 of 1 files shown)


azaroth42:
Any thoughts @jpstroop @zimeon @tomcrane @mikeapp ?

azaroth42:
(New PR was from rebasing) @zimeon was in favor here: https://github.com/IIIF/iiif.io/pull/696#issuecomment-173717573

jpstroop:
Propose this:

Server implementations may rely on frameworks that unescape the URI path, as is the case with the [Python Web Server Gateway Interface](https://www.python.org/dev/peps/pep-0333/). In such situations, request path for a given prefix can be parsed from the right in order to handle identifiers possibly containing slashes, given the knowledge of the API parameters and the prefix for which the server handles requests. Responses for missing identifiers containing slashes should result in a 404 response, rather than a 400 response for incorrect parameter values if this parsing approach is taken.

jpstroop:
I'm not sure that I understand 

>  Responses for missing identifiers containing slashes may result in a 404 response, rather than a 400 response for incorrect parameter values if this parsing approach is taken.

Is this to account for the fact that it's still impossible to distinguish between a bad request (400) and a base URI that contains slashes and needs to be redirected to `/info.json`?

sdellis:
I noticed from the link to #696 that it applies to the Image API and the server, but I have been wondering about client side parsing of IDs with slashes (it's helpful in many instances to use the shortened ID on the client rather than the full URI).  If there is a method for determining the ID (parsing from the right?) from the URI, it would be helpful to provide an example with this note.

zimeon:
I'm not sure whether the 400/404 comment adds very much. In practice it doesn't seem to matter much as I don't imagine clients doing anything different for these different codes. I'm :+1: with the current PR but would also be fine removing the 400/404 sentence, or trying to clarify it with something more long-winded:

> If this parsing approach is used then the response code may be 404 in the case of a bad or missing identifier containing slashes, when it might otherwise be a 400 response if the path were parsed from the left.

jpstroop:
I think we need to mention WSGI to make it clear why we're talking about this at all. I agree the 400/404 sentence could go.

At a minimum, the sentence that ends with "the prefix which the server handles requests for." needs to be fixed

jpstroop:
:+1:

mikeapp:
:+1:

zimeon:
:+1:

PR #724, opened 2016-02-17
Add Members

### LINKED ISSUE #716: Collection / Range membership
There are requirements (see #697 and #646) to order collections and manifests within a collection, and ranges and canvases within a range.  Currently they exist only in two separate, ordered lists, making interleaving impossible.

The proposed solution is to allow a `members` property that allows either, with the additional restriction that `@id`, `@type`, and `label` MUST be present and `viewingHint` SHOULD be present for collections and manifests.

If a client sees a `members` property, then it should use it even if the regular properties exist.  Publishing systems should be aware that if they use members and do not provide a fall back for the split properties, then 2.0 clients will not produce the expected results.  

The split properties would be deprecated in 3.0

jpstroop:
Closes #716.

PR is not likely ready, but now you can see the diff. First draft is here:

http://membership.iiif.io/api/presentation/2.1/

A few questions:

 * [ ] Should we mention that `members` makes it easier for a client to navigate the resource using `structures`? Do we want to discourage that--and if so, do we do it explicitly or by staying silent?
 * [ ] @azaroth42 seemed to think (based on a discussion in person a few weeks ago) that ranges could now point to manifests as well (as with layers). If that's supposed the case (and I'm skeptical), it could be added with this PR or separately. 7.1 is a little bit of a mess and could use a rewrite to be more consistent w/ 7.3. This would probably push it over the edge, thus warranting a separate ticket.
 * [ ] Should a collection's members _always_ require a `viewingHint`, or just when the member is itself a collection? If always, then we need `viewingHint`s for everything that isn't a collection or an multi-volume work, and that seems awkward.
 * [ ] `ranges`/`canvases` for ranges and `collections`/`manifests` for collections should be deprecated, right? See the deprecation warnings. Very questionable wrt semver, but probably OK, if you squint :eyeglasses:?

Feedback welcome.

### DIFF
--- a/source/api/presentation/2.1/index.md
+++ b/source/api/presentation/2.1/index.md
@@ -1124,9 +1124,11 @@ A range will typically include one or more canvases or, unlike sequences, parts 
 
 Ranges _MAY_ include other ranges.  This is done in a `ranges` property within the range.  The values within the ranges list _MUST_ be strings giving the URIs of ranges in the list in the manifest.
 
-Ranges _MAY_ link to a layer, described in the next section, that has the content of the range.  This allows, for example, the range representing a newspaper article that is split across multiple pages to be linked with the text of the article. Rendering clients might use this to display all of the article text, regardless of which canvas is being viewed. The link is created by adding a `contentLayer` property to the range, with a single value of the layer's URI.  The layer will then have one or more annotation lists, each of which contains annotations that target the areas of canvases within the range, and provide the content resources.
+In cases where the ordering of a set of ranges and canvases is significant, the `members` property can be used. Each member of the `members` list _MUST_ be an object, and that object _MUST_ include `@id`, `@type`, and `label`.
 
-Ranges are linked or embedded within the manifest in a `structures` field.  It is a flat list of objects, even if there is only one range.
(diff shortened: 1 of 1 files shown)


source/api/presentation/2.1/index.md
```diff
@@ -1295,6 +1344,43 @@ An example collection document:
...
+
+  "members": [
+    {
+      "@id": "http://example.org/iiif/collection/part1",
+      "@type": "sc:Collection",
+      "label": "My Multi-volume Set",
+      "viewingHint": "multi-part"
+    },
+    {
+      "viewingHint": "multi-part"
```
mikeapp: missing comma

jpstroop: and shouldn't be here

zimeon:
I'm :+1: to merge this so that it doesn't get in the way of other work. With the understanding that it will need another pass before 2.1 hits the shelves.

mikeapp:
:+1: I'm with Simeon, merge now and revisit later.

azaroth42:
:+1: Merging, then will rebase #738

Issue #720, opened 2016-02-17
Include sizeByWhListed in profiles


mikeapp:
`sizeByWhListed` is required at all compliance levels, so it should be included in the `supports` section of the profile documents.

zimeon:
I agree

azaroth42:
:+1: to the issue described (c.f. comment on #727)

PR #727, opened 2016-02-21
Size and profile clarifications

tomcrane:
Fixes https://github.com/IIIF/iiif.io/issues/720
Question - those profiles are for 2.0 too (because semver) - is that a problem? 

Fixes sizeByForcedWh syntax typo in compliance doc
Question - should this be fixed in the 2.0 compliance doc too?

### DIFF
--- a/source/api/image/2.1/compliance.md
+++ b/source/api/image/2.1/compliance.md
@@ -54,17 +54,18 @@ See also the note under [Size][size] about combinations of Size and Region that 
 
 ### 3.2 Size
 
-| Syntax | Feature Name     | Level 0 | Level 1 | Level 2  |
-|:-------|:-----------------|:-------:|:-------:|:--------:|
-| `full` |                  | ![required][icon-req] | ![required][icon-req] | ![required][icon-req] |
-| `max`  |                  | ![optional][icon-opt] | ![optional][icon-opt] | ![optional][icon-opt] |
-| w,h    | `sizeByWhListed` | ![required][icon-req] | ![required][icon-req] | ![required][icon-req] |
-| w,     | `sizeByW`        | ![optional][icon-opt] | ![required][icon-req] | ![required][icon-req] |
-| ,h     | `sizeByH`        | ![optional][icon-opt] | ![required][icon-req] | ![required][icon-req] |
-| pct:n  | `sizeByPct`      | ![optional][icon-opt] | ![required][icon-req] | ![required][icon-req] |
-| !w,h   | `sizeByForcedWh` | ![optional][icon-opt] | ![optional][icon-opt] | ![required][icon-req] |
-| w,h    | `sizeByWh`       | ![optional][icon-opt] | ![optional][icon-opt] | ![required][icon-req] |
-|        | `sizeAboveFull`  | ![optional][icon-opt] | ![optional][icon-opt] | ![optional][icon-opt] |
+| Syntax | Feature Name        | Level 0 | Level 1 | Level 2  |
+|:-------|:--------------------|:-------:|:-------:|:--------:|
+| `full` |                     | ![required][icon-req] | ![required][icon-req] | ![required][icon-req] |
+| `max`  |                     | ![optional][icon-opt] | ![optional][icon-opt] | ![optional][icon-opt] |
+| w,h    | `sizeByWhListed`    | ![required][icon-req] | ![required][icon-req] | ![required][icon-req] |
+| w,     | `sizeByW`           | ![optional][icon-opt] | ![required][icon-req] | ![required][icon-req] |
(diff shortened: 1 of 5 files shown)


tomcrane:
Hmm, external link validation is failing the build, but the links to http://www.openannotation.org/ return HTTP 200 for me.

mikeapp:
I don't see a semver problem, `sizeByWhListed` is required at all compliance levels even in 2.0.   It shouldn't be a breaking change.

azaroth42:
Agree with the issue in #720, but there is the additional question of which document is correct for the description of `sizeByForceWh`.  My interpretation of "Forced" was that the aspect ratio is forced, and thus compliance is correct and the base spec is wrong.  `!w,h` doesn't "force" anything.

jpstroop:
> My interpretation of "Forced" was that the aspect ratio is forced, and thus compliance is correct and the base spec is wrong. !w,h doesn't "force" anything.

I thought that's what we were saying...but now I see the PR inverts this. To me, forced == distorted.

zimeon:
I agree that to me forced => distorted, and thus is not about `!w,h`. I think we have a spec bug to fix.

mikeapp:
Agree on forced => distorted.  ImageMagick syntax uses `!` for the case where aspect ratio is forced, too bad we're not consistent.

zimeon:
Editors call: Agree that there is a bug and that we haven't yet included a concept that matches `!w,h`. Per  @tomcrane's proposal we should have:

  * `w,h`  `sizeByWh ` - a request that preserves aspect ratio (same as 2.0)
  * `w,h`  `sizeByDistortedWh` - a request that "knowingly" does not preserve aspect ratio (addition)
  * `!w,h` `sizeByConfinedWh` - confine to box w,h without distorting (addition)

and that in 2.1 we deprecate `sizeByForcedWh`:

  * in 2.0 there were inconsistent descriptions of `sizeByForcedWh`
  * deprecated in 2.1 in favor of `sizeByDistortedWh` and `sizeByConfinedWh`
  * `sizeByForcedWh` will be removed in 3.0

zimeon:
Have sent note to [IIIF-Disuss](https://groups.google.com/forum/#!topic/iiif-discuss/47_o2BJjC7c)

tomcrane:
I thought this would be simple to add... but while doing it I came up with these questions:

1. `sizeByConfinedWh` - what happens if the image is smaller than the box `!w,h`? Do we scale up? I think not - it's still confined to the box if the image is only 1000,1000 and you asked for !2000,2000 - it doesn't _have_ to touch the sides.

2. Profile JSON at http://iiif.io/api/image/2/level2.json 
This document needs `sizeByConfinedWh` and `sizeByDistortedWh` for 2.1 - but then it no longer works for 2.0. 

3. Should the compliance document reproduce or link to the "sizes" canonical inconsistency in the image API doc? 

4. `sizeByWhListed` was always a named feature, and after discussion we added it to all compliance level profiles (earlier in this PR), because it must be supported at all levels - if the info.json has 'sizes' then they must be honoured. However, what's the difference between a size that must be served because it's listed under `sizes` and a size that must be served because it's implied by the value of the `tiles` property? I don't think there is a difference - one has the word "size" in the property but they are equivalent concepts. Which means that either:

  a. there's another all-compliance-levels feature under `supports` called `sizeImpliedByTiles` (or something to that effect), or
  b. `sizeByWhListed` isn't an explicit feature at all and should be removed from the `supports` list altogether

azaroth42:
1. Uhhh. Uhhhhhh..... .... raise an issue? :crying_cat_face: 

2. Yes. It needs all three. So long as we don't take out features, it'll work as expected for previous versions.

3. I don't think that compliance doc needs to talk about the canonical format.

4.  I think there's sizeByTiles feature (or whatever). New issue?

jpstroop:
For `#1` above, see #752

azaroth42:
Per call, add periods to end of table descriptions

mikeapp:
:+1:

zimeon:
Regarding question 4, I favor removing `sizeByWhListed` as description of the implicit feature, rather than adding `sizeImpliedByTiles`

azaroth42:
After discussion, I also favor removing sizeByWhListed. 

Also, table in section 5 is missing api-table styling.

zimeon:
Need to be consistent with full-stops in table 5.2, mark `sizeByWhListed` as deprecated in the table.

In compliance document add pithy description of the sizes that must be supported by implication from the `sizes` and `tiles` properties

tomcrane:
> table in section 5 is missing api-table styling

...it now has it - but it gains a grey background from .urltemplate as well.

azaroth42:
:+1:

mikeapp:
:+1:

zimeon:
:+1:

PR #729, opened 2016-02-23
Add IDs for properties

### LINKED ISSUE #728: Facilitate deeper links into the specs
I've wanted to link to viewingHint about twenty times in the past year. Any objections to adding ids to the properties listed in section 3 to make this possible?

There's temptation to think about the difference between the URIs we've minted for our properties and the new hash URIs that would result from this. I don't feel a need to worry about it now...I'd just like to be able to add links to documentation to emails, etc.

jpstroop:
Closes #728 

 * Changes definition lists to `h5`s to get links for free
 * Stops toc plugin from rendering deeper than h4
 * Adds slick hover links on headings.

Example: http://property_ids.iiif.io/api/presentation/2.1/#viewinghint


zimeon:
:+1: looks good and I like the addition of a way to find the anchor URIs with the hover over them

azaroth42:
:+1: non normative, gtg?

mikeapp:
:+1:

PR #733, opened 2016-02-24
Fix http://iiif.io/api/annex/services/ examples to use site image API version

### LINKED ISSUE #731: Fix profile link in annex/services
Missing major version expansion in services, from Chad Mills:

While reading through the documentation on http://iiif.io/api/annex/services/ under 3.1 I noticed this profile values which isn't in step with the rest.

```
      "profile": "http://iiif.io/api/image//2/level2.json"
```

zimeon:
Replaces page.major with site.image_api.latest.major for image API version and fixes an instance with extra explicit `/2/`. Closes #731

### DIFF
--- a/source/api/annex/services/index.md
+++ b/source/api/annex/services/index.md
@@ -105,7 +105,7 @@ The service _MAY_ have additional information embedded from the Image Informatio
       {"width" : 512, "scaleFactors" : [1,2,4,8,16]}
     ],
     "profile" : [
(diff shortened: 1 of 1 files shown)


azaroth42:
Do we want to auto-update or stay pinned to services? I'm fine with the PR, FWIW, but want to make sure the implication is clear.

:+1:

mikeapp:
:+1:

Issue #734, opened 2016-02-24
Rephrase paging properties to be clearer 


azaroth42:
@zimeon:
> "A collection may have exactly one collection as its next sibling in a list of sub-collections."

And similar

azaroth42:
Closed by #738

Issue #710, opened 2016-02-04
Add Collection, Range, Layer to section 3 consistently


azaroth42:
Combining #703 and #709, the property descriptions in section 3 of the Presentation API should consistently have usage entries for Collection, Range and Layer where appropriate. These should thus also be described in the introduction, and perhaps move the full model diagram up.

azaroth42:
Closed by #738

Issue #697, opened 2016-01-21
Where do I put multi-volume items in an iiif:collection?


jpstroop:
Surfaced by @tpendragon:

In 2.1 we defined the viewingHint "multi-part", which is:

> Valid only for collections. Collections with this hint consist of multiple manifests that each form part of a logical whole. Clients might render the collection as a table of contents, rather than with thumbnails. Examples include multi-volume books or a set of journal issues or other serials.

So we seem to agree that the way to describe a multi-volume bookset is with a collection that lists _n_ references to manifests. However, when I have a collection that contains single and multi-volume books, it feels awkward to put the multi-volume books in `collections : [ ]` and the single-volume books in `manifests : [ ]` when they're really meant to collectively comprise a sigle collection regardless of these physical characteristics.

Proposal (not sure I like it, but may as well start somewhere...): recommend using the current properties when you're in a 'mixed-content' situation (i.e. have a collection that includes items AND actual sub-collections), and add a new property "members" that is less opinionated and can contain both. We'd need find or define a predicate for the context as well, and I'm not quite sure how this would fit into the SC model.

tomcrane:
I have this need too, and I get round it currently with a bit of a fudge:

http://wellcomelibrary.org/service/collections/topics/Physiology,%20Pathological/

There are 5 books in this category, so I'd like to show 5 IIIF resources together.

One of these manifests (http://wellcomelibrary.org/iiif/b18031511/manifest) is not really a manifest; the server redirects to http://wellcomelibrary.org/iiif/collection/b18031511. I'm not suggesting this is the answer, it would be solved by a collection that could accommodate both.

The other reason (or excuse...) for this redirect behaviour is to have a consistent guessable URL for an IIIF resource given a Wellcome Library identifier:

http://wellcomelibrary.org/iiif/<bib-id>/manifest always works, but it redirects to a collection for a multi-volume item.

azaroth42:
To restate the issue slightly ...

It is currently possible to have collections and manifests within a collection, but it is not possible for them to be interleaved in a single ordering.  The cause of this is the separation by type (`collections` and `manifests`) rather than by the reason that the resource is included in the collection (hierarchy of collections, versus leaf nodes in that hierarchy regardless of type).  This was not obvious in 2.0, as 2.1 introduced the pattern of using a Collection for a Multi-Volume Work, which should be a leaf node but must go into `collections`.

Yes?

And thus we can (1) have a single ordered pool of members:
```
{
[...]

Issue #646, opened 2015-12-10
How do I know the order of canvases and ranges in a Range?


jpstroop:
If I have a Range that has canvases and ranges, how do I know the order in which to display them? Some of the `canvases` may come before the `ranges` and others after.

Example: a book has covers and fly leaves at the beginning and end. I'd want to have canvases for the upper cover and front fly leaves, and then ranges for the chapters, and then more canvases for the back fly leaves and lower cover.

azaroth42:
Do we consider this in scope for 2.1?  I'm tempted to defer until 2.2 at this stage.

jpstroop:
Seems it's likely result in a breaking change , no?

jpstroop:
@tpendragon is actively developing an authoring client...curious what his thoughts are.

tpendragon:
Oh, yeah, I should comment on this:

There's no way, currently. I either have to do one of two things:

A) Create intermediate ranges with a single canvas and a meaningless label. These'll show up in the viewer and there's no hints to get them to not display.

B) The viewer we're using (universal viewer), doesn't really use range ordering for anything but the first element in the range (and I think it wouldn't even do that if you used startCanvas) - then it uses them as sets. So if the page you're on in the sequence appears in a structure, highlight the range it's in on the left. In that way, it doesn't matter - the structure controls paging, order doesn't matter, canvas labels don't show in the table of contents. It's only an issue if you want "Chapter 2a"'s canvas B to be where the viewer goes to when you click "Chapter 2", but then there's an argument that canvas B should just be in both ranges, which solves that problem.

If structures are meant to be able to be paged through one canvas at a time, then it's an issue and you need something like "elements" rather than "ranges and structures". Otherwise, I think it's fine.

sdellis:
How do you know which Sequence a Structure is referring to?  You would need this if you wanted to extract order for a Range from a Sequence, especially since a Canvas could exist in different order in multiple Sequences.

azaroth42:
@sdellis Ranges don't refer to Sequences, and the Range recursively includes all of the canvases by reference.

The order could be generated by creating a list of all canvases, recursively, in the range. Then order the canvases by the current sequence's order of canvases.

sdellis:
@azaroth42 Ok, I see.  There must always be a "current" sequence that is supplying order to the ranges.  Something still doesn't sit right with me re: this approach, but perhaps my expectations of what one can do with Ranges is inaccurate.  

[...]

PR #738, opened 2016-02-26
Re-structure presentation API doc; closes #723

### LINKED ISSUE #723: "Figure 3. All Resource Types"
Where are figures 1 & 2?

azaroth42:
Overkill on closing #723
http://prezi_refactor.iiif.io/api/presentation/2.1/


jpstroop:
This is basically the outline we were discussing on slack, and I like it. Some nitty-gritty:

 * [ ] [contentLayer](http://prezi_refactor.iiif.io/api/presentation/2.1/#contentlayer) has _Usage:_ above the bullets. None of the other terms do.
 * [ ] The use of bullets in [Section 3](http://prezi_refactor.iiif.io/api/presentation/2.1/#resource-properties) seems a little sporadic. I propose we use bullets exclusively for the normative/RFC statements, and move anything that is not RFCish into the preceeding prose. And the inverse: take any RFC normative/RFC statements out of the prose and make them bullets.
 * [ ] Somewhat related to the previous point, [viewingHint](http://prezi_refactor.iiif.io/api/presentation/2.1/#viewinghint) and [viewingDirection](http://prezi_refactor.iiif.io/api/presentation/2.1/#viewingdirection) have two bulleted lists: one that lists the possible values, and one for the RFC points. I think they're separate `ul`s, but visually they run together. Could we come with some styling that makes the supported values distinct? Maybe they should be linkable too (`h6`)?
 * [ ] Just CSS: The spacing in [Section 2](http://prezi_refactor.iiif.io/api/presentation/2.1/#resource-type-overview) (in particular) gets tight. I wonder if we could try removing the left margin applied to the text under the `h5`s (they were `dl`s until early last week).
 * :+1: to collapsing the `members` examples into the existing `ranges`/`canvases` and `collections`/`manifests` examples
 * [ ] Examples are bulleted for [Segements](http://prezi_refactor.iiif.io/api/presentation/2.1/#segments) and nowhere else that I can see.
 * [ ] Let's see if we can override the HTML `id`s in [Section 2](http://prezi_refactor.iiif.io/api/presentation/2.1/#resource-type-overview) to, e.g. `#manifest-overview`, so that Manifest (currently [5.1](http://prezi_refactor.iiif.io/api/presentation/2.1/#manifest-1)) can be `#manifest` instead of `#manifest-1`.
 * [ ] Should Section 2 contain forward references to the detailed description in Section 5?

jpstroop:
I should say, I'm happy to take a stab at any of this if you'd like. :hocho:

azaroth42:
For the styling in Section 2 ... agree that it's messy. The float right only works for some screen widths, and the images overlap down into the following sections. What I'd like is what we would have used a table for in 1995 (when I learnt HTML) to make sure that everything stays in line.

I've done the other tasks, and redeployed.

jpstroop:
Does getting rid of the left padding/margin really make that much difference wrt that responsive problem? I'd like to see [these lines](https://github.com/IIIF/iiif.io/blob/prezi_refactor/source/css/spec-doc2.scss#L109-L114) go from the CSS.

azaroth42:
I've tried futzing around with the CSS here without success. Can we merge as is and fix the CSS later?

azaroth42:
Closes #734 ; Closes #710 ; Closes #697 ; Closes #646 ; Closes #58

jpstroop:
Talking of saving space over on #739, there a lot of properties in [Section 3](http://prezi_refactor.iiif.io/api/presentation/2.1/#resource-properties) that have a single bullet which, I think, could just as easily be the last sentence of the preceding paragraph.

tomcrane:
http://prezi_refactor.iiif.io/api/presentation/2.1/#thumbnail has six statements about cardinality, whereas http://prezi_refactor.iiif.io/api/presentation/2.1/#attribution has just one - but I think the presentation should be the same in both cases even if it results in a dangling bullet. I _personally_ don't mind that at all in this context, although it annoys me in others!

jpstroop:
A list of one thing...it just bugs me. I'm out of time for the moment, but what about:

__Cardinality:__ One constraint

__Cardinality:__
 * One constraint
 * Two constraint
 * Three constraint

tomcrane:
Some of the bullets combine a statement about cardinality with some additional information that goes with it, e.g., "A content resource may have exactly one format, and if so, it must be the value of the Content-Type header returned when the resource is dereferenced." Which might make a "Cardinality:" prefix slightly misleading. We want those clauses to stay together, but it's not always just about cardinality.

mikeapp:
For me, having consistency within the presentation of the properties helps, even if it results in lone  bullets.

zimeon:
I'm with @tomcrane and @mikeapp ... consistent layout trumps lone bullet issues for me

jpstroop:
OK, fine, but (one last try)--what is the consistency? Until the reader puts it together, it just looks like some RFC assertions are in paragraphs, and some are in bullets. Why? We just got sick of writing text? Maybe "Cardinality" isn't the right word. Maybe "Constraints"? We should explain the intent of those lists.

zimeon:
I could see adding a word at the end of the paragraph before the bullets (could be "Constraints:") but the original argument for removing all the floating "Usage:" lines (#651) was that it was clutter that didn't help readability.

jpstroop:
Bah, I missed #651! I would have argued against.

I think:

__Constraints:__ One constraint

__Constraints:__
 * One constraint
 * Two constraint
 * Three constraint

Would be better.

zimeon:
Hmmm... I'm still very much in favor of #651, I think the 2.1 draft is better than 2.0 for layout.

jpstroop:
Overall the layout is absolutely better. I just don't think the bullets makes sense. _gives up_

azaroth42:
Given that we're failing in master and this fixes it, plus there's a lot of other branches ready to merge once this hits, can we hit the green button? :+1:s requested from @IIIF/editors

jpstroop:
:+1:

zimeon:
:+1:

mikeapp:
:+1:

PR #747, opened 2016-03-04
Initial int; closes #732

### LINKED ISSUE #732: Move SHOULD NOT for initial int to Implementation Note
... in all classes for Presentation API. It doesn't really deserve to be normative in the spec for such a specific technology stack that would have trouble with it.

azaroth42:
Move initial integer recommendations to implementation note. Closes #732 
Link: http://initial_int.iiif.io/api/presentation/2.1/#d-implementation-notes


zimeon:
Two suggested changes:

  1. s/then certain technology stacks may be inconvenienced/then developers using certain technology stacks may be inconvenienced/ --- the developer is inconvenienced, not the technology (which is like honey badger). However, the more I read this advice the more I think the value of the note is to point out to folks using XML/RDF that they'd better not assume they can use a prefix! Unless we were to mandate an alpha prefix, any implementer of a consuming app will have to handle the numeric case (and I'm certainly _NOT_ suggesting such a restriction)
  2. Don't escape < > in the md, these get escaped a second time when converted to HTML

zimeon:
:+1:

jpstroop:
:+1:

mikeapp:
:+1:

tomcrane:
:+1:

PR #748, opened 2016-03-04
Add structural properties table; closes #739

### LINKED ISSUE #739: `members`, `canvases`, `ranges`, `collections` to Linking properties & table
These are only mentioned in later part of the spec. I think they should be treated like any other property, especially `members` since it's used twice.

azaroth42:
http://structural739.iiif.io/api/presentation/2.1/#b-summary-of-metadata-requirements

### DIFF
--- a/source/api/presentation/2.1/index.md
+++ b/source/api/presentation/2.1/index.md
@@ -476,11 +476,11 @@ The manifest response contains sufficient information for the client to initiali
 
 The identifier in `@id` _MUST_ always be able to be dereferenced to retrieve the JSON description of the manifest, and thus _MUST_ use the http(s) URI scheme. 
 
-After the descriptive information, there is then a `sequences` section, which is a list of objects. Each object is a sequence, described in the next section, that represents the order of the views and the first such sequence should be included within the manifest as well as optionally being available from its own URI. Subsequent sequences _MUST_ only be referenced with their identifier (`@id`), class (`@type`) and `label` and thus _MUST_ be dereferenced by clients in order to process them if the user selects to view that sequence.
+Along with the descriptive information, there is a `sequences` section, which is a list of JSON-LD objects. Each object describes a [Sequence][sequence], discussed in the next section, that represents the order of the parts of the work, each represented by a [Canvas][canvas].  The first such sequence _MUST_ be included within the manifest as well as optionally being available from its own URI. Subsequent sequences _MUST_ only be referenced with their identifier (`@id`), class (`@type`) and `label` and thus _MUST_ be dereferenced by clients in order to process them if the user selects to view that sequence.
 
-After the `sequences` section there _MAY_ be a `stuctures` section listing one or more [ranges][range] which describe additional structure. 
+There _MAY_ also be a `stuctures` section listing one or more [Ranges][range] which describe additional structure of the content, such as might be rendered as a table of contents.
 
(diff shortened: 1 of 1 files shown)


zimeon:
Suggest additional text introducing section 5 and linking to be summary tables, perhaps:

> This section provides detailed description of the resource types used in this specification. [Section 2](http://structural739.iiif.io/api/presentation/2.1/#resource-type-overview) provides an overview of the resource types and figures illustrating allowed relationships between them, and [Appendix B](http://structural739.iiif.io/api/presentation/2.1/#b-summary-of-metadata-requirements) provides summary tables of metadata requirements.

Perhaps also in [5.1 Manifest](http://structural739.iiif.io/api/presentation/2.1/#manifest) there should be a forward reference to [5.6 Range[(http://structural739.iiif.io/api/presentation/2.1/#range) with mention of `structures`. Perhaps add to end of para starting "The identifier...":

> After the `sequences` section there _MAY_ be a `stuctures` section listing one or more [Ranges](http://structural739.iiif.io/api/presentation/2.1/#range) which describe additional structure.

azaroth42:
Thanks @zimeon, great comments.  Fixed and redeployed.

source/api/presentation/2.1/index.md
```diff
@@ -472,7 +474,11 @@ Recommended URI pattern:
...
+The identifier in `@id` _MUST_ always be able to be dereferenced to retrieve the JSON description of the manifest, and thus _MUST_ use the http(s) URI scheme. 
+
+After the descriptive information, there is then a `sequences` section, which is a list of objects. Each object is a sequence, described in the next section, that represents the order of the views and the first such sequence should be included within the manifest as well as optionally being available from its own URI. Subsequent sequences _MUST_ only be referenced with their identifier (`@id`), class (`@type`) and `label` and thus _MUST_ be dereferenced by clients in order to process them if the user selects to view that sequence.
```
jpstroop: It's bothering me that this reads as though we're prescribing the order of the parts of the manifest. We have this language in 2.0, but leading off paragraphs with "After the" draws more attention to it. Is this recommended, required, or does it just apply to the example below? I'd really rather stay away from this unless there's a practical reason for it.

jpstroop: Regardless of the first comment above:

"...there is then a `sequences` section, which is a list of objects. Each object is a sequence, described in the next section, that represents the order of the views and the first such sequence should be included within the manifest as well as optionally being available from its own URI."

I had to read this a couple of times to get it to make sense to me. Each object where? Also, what's a "view"? How about:

...there is then a `sequences` section, which is a list of objects. Each object in that list is a [Sequence][forward-reference] that represents the order of the [Canvases][forward-reference] that make up the object. The first such sequence _SHOULD_ be included within the manifest as well as optionally being available from its own URI.

Also, why not RFC that "should" (as I have above)? We have an RFC "must" in the next sentence.

azaroth42:
@jpstroop Otherwise +1 when I fix the above?

jpstroop:
I think so. I expected the first comment to raise controversy--and it should block this issue. If it goes unchanged here I can open another. :+1:

mikeapp:
:+1:

zimeon:
:+1:

tomcrane:
:+1:

PR #793, opened 2016-04-11
Add language algorithm to prezi

### LINKED ISSUE #777: Fix presentation API description multi-language properties to match image API
We have new, clearer text for the image API per #771, this should be copied to the presentation API in http://iiif.io/api/presentation/2.1/#language-of-property-values

azaroth42:
Add carefully crafted language selection algorithm to Prezi 2.1

Closes #777

### DIFF
--- a/source/api/presentation/2.1/index.md
+++ b/source/api/presentation/2.1/index.md
@@ -426,7 +426,7 @@ Note that [RFC 5646][rfc5646] allows the script of the text to be included after
 In the case where multiple values are supplied, clients _MUST_ use the following algorithm to determine which values to display to the user.  
 
 * If none of the values have a language associated with them, the client _MUST_ display all of the values.
-* Else, the client should try to determine the user's language preferences.  If the user's language preferences are unknown, then the client _MUST_ supply a language preference. Then:
+* Else, the client should try to determine the user's language preferences, or failing that use some default language preferences. Then:
   * If any of the values have a language associated with them, the client _MUST_ display all of the values associated with the language that best matches the language preference.
   * If all of the values have a language associated with them, and none match the language preference, the client _MUST_ select a language and display all of the values associated with that language.
   * If some of the values have a language associated with them, but none match the language preference, the client _MUST_ display all of the values that do not have a language associated with them.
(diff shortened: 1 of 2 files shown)


source/api/presentation/2.1/index.md
```diff
@@ -423,6 +423,15 @@ This pattern may be used in `label`, `description`, `attribution` and the `label
 
 Note that [RFC 5646][rfc5646] allows the script of the text to be included after a hyphen, such as `ar-latn`, and clients should be aware of this possibility. This allows for full internationalization of the user interface components described in the response, as the labels as well as values may be translated in this manner; examples are given below.
 
+In the case where multiple values are supplied, clients _MUST_ use the following algorithm to determine which values to display to the user.  
+
+* If none of the values have a language associated with them, the client _MUST_ display all of the values.
+* Else, the client should try to determine the user's language preferences.  If the user's language preferences are unknown, then the client _MUST_ supply a language preference. Then:
```
jpstroop: Should we be more specific about what "the client _MUST_ supply a language preference" means? Do we expect the client to prompt users? Make an arbitrary guess?

zimeon: It does seem a bit odd to say a client must _supply_.  It seems to me we are just saying that the client has to do something, perhaps make an arbitrary guess or, more likely, have a built in default of the language preference of the client author. I think the should in the Else clause should be RFC2119, and then perhaps change the following sentence to something weaker. Perhaps:

  *  Else, the client _SHOULD_ try to determine the user's language preferences in preference to using a default setting. Then:

zimeon: ug... "preferences in preference". Perhaps:

  * Else, the client SHOULD try to determine the user's language preferences, or failing that use some default language preferences. Then:

azaroth42: :+1: to @zimeon's reformulation. Will add for both Image and Prezi

zimeon:
:+1: (and rebased to make CI happy)

azaroth42:
@jpstroop or @tomcrane if you are now okay, feel free to merge?

jpstroop:
I don't love how it reads, but can't make a better suggestion at the moment. I'll merge and open a new PR later in the week if I get time.

Issue #791, opened 2016-04-11
Clean up old meeting notes pages to remove links not expected to persist


zimeon:
@jpstroop wrote in #788: "I had to fix that link on my branch for the scrolling hashes feature as well. I think we need to think about how long it's really necessary to keep links to random restaurants and hotels post-meetings."

I agree, I think we should have a practice of cleaning these from meetings pages and do a pass over the current ones to clean up.

zimeon:
Closing as done, now part of Sheila's process

PR #796, opened 2016-04-11
Wider remote link check interval

jpstroop:
Make the link checker only run every 4 weeks.


azaroth42:
:+1:

zimeon:
:+1: but I also still propose #791 to reduce our long-term problems with pages where the remote links no longer matter

PR #797, opened 2016-04-11
remove double lines

azaroth42:
from Image API 2.0. Checked the other specs and couldn't find any more.


zimeon:
:+1: changed look good

PR #801, opened 2016-04-12
Make q optional in search, confirm required in autocomplete

### LINKED ISSUE #765: Supporting "q" param should not be a mandatory server requirement in Search API
The client does not have to supply a q parameter, but currently a service has to support it if supplied.

This precludes a service that is similar to the level 0 Image API, where a functioning and useful image service can be constructed from entirely static resources (e.g., files on disk).

#764 allows a server to return pre-computed annotation lists for a resource by different granularities.
A service could return all the annotations from a particular user, without also having to support text searching over them.

azaroth42:
Closes #765.

### DIFF
--- a/source/api/index.md
+++ b/source/api/index.md
@@ -21,7 +21,7 @@ cssversion: 2
 | Image API          | [2.1.0][image21] *(Beta draft)* |
 | Presentation API   | [2.1.0][prezi21] *(Beta draft)* |
 | Authentication API | [0.9.1][auth09] *(Beta draft)* |
-| Search API         | [0.9.1][search09] *(Beta draft)* |
+| Search API         | [0.9.3][search09] *(Beta draft)* |
 {: .api-table}
 
 __Feedback Requested__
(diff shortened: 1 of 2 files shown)


mikeapp:
:+1:

zimeon:
:+1:

tomcrane:
Couple of points, I can make a PR:

> The scope of the specification is searching textual annotation content within a single IIIF resource, such as a Manifest or Range.

Maybe remove the word "textual"? I could search for all geo tags made by user X with the API. Resource bodies might use only URIs, no string literals being searched over

How about "a Collection, Manifest or Range" - this shows that you could search for annos across manifests, and could in fact support a text search service across an entire library with the API - with the caveat that your annotation results need to assert what they are `within` to be useful.

zimeon:
I take @tomcrane's suggestion as a "friendly ammendment" which seems easiest to handle in this PR. I thus removed "textual" and added "Collection" -- I kept Manifest first in the order though because that seems the primary and simpler use-case.

I also fixed the title to reflect v0.9.3, and the version in http://iiif.io/api/#draft-specifications

Issue #805, opened 2016-04-14
Use TEI media type in TEI examples


jpstroop:
https://tools.ietf.org/html/rfc6129

azaroth42:
Closed by #806.

PR #806, opened 2016-04-14
Fix typos from Rob Lancefield

### LINKED ISSUE #804: Typos
Thanks to Rob Lancefield for the careful reading!

## Image API

- [x] 4.2. Size First table, row 2, col 2: "This is the same as full if none of these are properties are provided." The first "are" should be deleted. 

- [x] 4.6. Order of Implementation Second paragraph, only sentence: "The parameters should be interpreted as if the the sequence of image manipulations were:" The last "the" should be deleted.

- [x] 4.7. Canonical URI Syntax  First table, row 2, col 2, "Note": "The the size keyword...." The second "the" should be deleted.

- [x] 5.2. Technical Properties Second table, row 2, col 3: "aas an integer" should be "as an integer."

- [x] 5.3. Profile Description Second paragraph, fifth sentence: "Servers must ensure that sizes specified within any sizes or tiles properties are with any size limits expressed." The "with" seems to want to be "within," unless I'm misreading intended sense.

- [x] 7.2. Error Conditions First table, row 1, col 2: UK spelling of "fulfil" may want to become US "fulfill" for consistency in the context of this document.

- [x] Appendix A. Implementation Notes Seventh bullet note: "for a full imagine" should be "for a full image."

## Presentation API

- [x] 2.1. Basic Types First sentence in "Sequence": "of a the object" wants to be "of an object."

- [x] 3.1. Descriptive Properties Last line: "Other resources types" wants to be either "Other resource types" or "A resource of any other type"

- [x] 5.8. Collection First paragraph, last sentence: "Embedded collections should have their own URI from which the description is also made available." May be clearer as "An [or: Each] embedded collection should have its own URI from which the description is also made available."

- [x] 6.1. Segments Second paragraph, first sentence: "Note that if both there are segments of both image and canvas" wants deletion of the first "both."

- [x] 6.3. Choice of Alternative Resources Last paragraph, last sentence: "this can be either done such that the entire image is switched to change state, or only the section of the image" wants to be "this can be done such that either the entire image is switched to change state, or only the section of the image"

## Search API

- [x] 1.1. Use Cases Trivial but noted in passing: Some items in bullet list end with periods, others don't.

- [x] 2. Overview Second paragraph, last sentence: "annotation lists" wants to be "Annotation lists"

- [x] 3. Search First sentence: "uri" wants to be "URI"

- [x] 3.3.1. Simple Lists Second paragraph, last sentence: "derefencable" wants to be "dereferenceable"

- [x] 3.4 Search API Specific Responses First paragraph, only sentence: UK spelling of "fulfil" may want to become US "fulfill" for consistency in the context of this document.

- [x] 3.4 Second paragraph, first sentence: "As these response" wants to be "As these responses"

- [x] 3.4.4. Multi-Annotations Hits First paragraph, last sentence: "on the method that the text and annotations were generated" would read more clearly as "on the method by which the text and annotations were generated"

- [x] 5. Property Definitions Under "count": Since these definitions are scoped to the context of server response to one given query, "A number of times that a term appears" seems to want to be "The number of times that a term appears."

azaroth42:
Closes #804

### DIFF
--- a/source/api/image/2.1/index.md
+++ b/source/api/image/2.1/index.md
@@ -205,7 +205,7 @@ The size parameter determines the dimensions to which the extracted region is to
 | Form   | Description |
 |--------|-----------------------|
 | `full` | The image or region is not scaled, and is returned at its full size. Note [deprecation warning][full-dep]. |
-| `max`  | The image or region is returned at the maximum size available, as indicated by `maxWidth`, `maxHeight`, `maxArea` in the [profile description][profile-description]. This is the same as `full` if none of these are properties are provided. |
+| `max`  | The image or region is returned at the maximum size available, as indicated by `maxWidth`, `maxHeight`, `maxArea` in the [profile description][profile-description]. This is the same as `full` if none of these properties are provided. |
 | w,     | The image or region should be scaled so that its width is exactly equal to w, and the height will be a calculated value that maintains the aspect ratio of the extracted region. |
 | ,h     | The image or region should be scaled so that its height is exactly equal to h, and the width will be a calculated value that maintains the aspect ratio of the extracted region. |
 | pct:n  | The width and height of the returned image is scaled to n% of the width and height of the extracted region. The aspect ratio of the returned image is the same as that of the extracted region. |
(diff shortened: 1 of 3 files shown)


azaroth42:
Also Closes #805.

jpstroop:
:+1:

zimeon:
:+1: and huge thanks to Rob Lancefield!

zimeon:
Three sets of eyes checking non-normative change seems enough, will merge.

PR #809, opened 2016-04-15
Conference schedule tweaks

zimeon:
Perhaps we'll want to add a couple of other tweaks but so far I have:
  * Moved metadata search to Thursday 3pm
  * Shifted MediaChain up to 4pm on Friday
  * Changed "IIIF A/V Update" to "IIIF A/V Update & Discussion" since I don't want to give the impression that this is eds speaking for 90mins!


azaroth42:
:+1:

PR #812, opened 2016-04-20
Update schedule to allow more time for breaks

azaroth42:
Had a call with NYAM today, and we need to have coffee breaks at set times, and to allow 90 minutes for lunch unless we're paying for everything (which we're not).

This PR updates the schedule given these considerations


jpstroop:
:+1:

mikeapp:
:+1:

tomcrane:
:+1:

PR #817, opened 2016-04-22
Clarify type

### LINKED ISSUE #815: Clarify @type requirement only for main resource classes
and not things like Service, Thumbnail and what have you.

Eds agree (4/22 call)

cc @regisrob

azaroth42:
Add a note to the "All ... MUST ..." under @type to clarify what all is being described.

http://clarify_type.iiif.io/api/presentation/2.1/#type

Closes #815.

### DIFF
--- a/source/api/presentation/2.1/index.md
+++ b/source/api/presentation/2.1/index.md
@@ -147,7 +147,7 @@ An ordered list of canvases, and/or further ranges.  Ranges allow canvases, or p
 
 ##  3. Resource Properties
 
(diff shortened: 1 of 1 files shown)


jpstroop:
:+1:

zimeon:
:+1: to content but a question on style: This would add a second "_N.B._" to the document, do we want that? Would it be clearer to instead stick with the more frequent use of "Note...", perhaps even "Note well.." as we have already one other place.

tomcrane:
:+1:  Given that these two sections are short and the NB is short, I think it would read fine without either NB or Note, just

This requirement applies....

This PR (and #820) need to be merged before announcing spec freeze I assume?

mikeapp:
:+1:

zimeon:
:+1: to "This requirement applies...."

azaroth42:
I think a green Note block would be overkill for this. Removing both NBs.

zimeon:
:+1: - CI issue is just the old link problem with Rob's homepage, will go away with merge

PR #818, opened 2016-04-22
Clarify SpecRes, Choice for image annos

### LINKED ISSUE #816: Image annotations could be oa:SpecificResource
http://iiif.io/api/presentation/2.1/#image-resources suggests that the annotation resource '**SHOULD** have an @type of “dctypes:Image”'

One reasonable exception to this is where the image annotations are of @type `oa:SpecificResource` - e.g., to crop out scanner machinery. The spec should warn that some clients might not be able to cope with this (it has to handle the crop).

azaroth42:
http://specific_image.iiif.io/api/presentation/2.1/#image-resources

Add note that just because it SHOULD be an image, doesn't mean that it always will be one, and there are lots of valid situations to test for.

Closes #816

### DIFF
--- a/source/api/presentation/2.1/index.md
+++ b/source/api/presentation/2.1/index.md
@@ -699,9 +699,9 @@ If a [IIIF Image API][image-api] service is available for the image, then a link
 
 Although it seems redundant, the URI of the canvas _MUST_ be repeated in the `on` field of the Annotation. This is to ensure consistency with annotations that target only part of the resource, described in more detail below.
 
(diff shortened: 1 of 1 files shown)


tomcrane:
:+1:

zimeon:
:+1:

jpstroop:
👍

mikeapp:
:+1:

azaroth42:
Jon, can you please deploy master... I don't have a laptop yet, let alone
cap installed!

Thanks!

R

Issue #700, opened 2016-01-23
Add sitemap submission to build script.


jpstroop:
After #699, do this:

http://support.google.com/webmasters/bin/answer.py?hl=en&answer=183669

Tack something like this onto the`publish.sh` script:

```
CURL=/usr/bin/curl

SITEMAP_URI="http%3A%2F%2Ffindingaids.princeton.edu%2Fsitemaps.xml"
GOOGLE_SVC="www.google.com/webmasters/tools/ping?sitemap="
LOG_DIR="../log"
#HEADER_LOG="$LOG_DIR/last_sitemap_submit_response.header"

request_url="$GOOGLE_SVC$SITEMAP_URI"
$CURL $request_url
CURL=/usr/bin/curl

SITEMAP_URI="http%3A%2F%2Fiiif.io%2Fsitemap.xml"
GOOGLE_SVC="www.google.com/webmasters/tools/ping?sitemap="
LOG_DIR="../log"

request_url="$GOOGLE_SVC$SITEMAP_URI"
$CURL $request_url
```

jpstroop:
Closed by #819

PR #819, opened 2016-04-24
Added Google notification

barnaclebarnes:
For #700. This should ping google after the build has completed. Not sure how to test the Capistrano workflow without being able to deploy. The http call has been tested and Google does pick it up. Namespacing under notify allows future notifications to be slotted in if there are other services to notify.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -25,4 +25,14 @@ namespace :deploy do
   end
 end
 
+namespace :notify do
+  desc 'Notify Google about sitemap update'
+  task :google do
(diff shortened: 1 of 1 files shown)


jpstroop:
I'm :ok_hand: with merging...the broken link that is causing the tests to fail will be fixed by the merge.

zimeon:
Seems sensible to me, :+1:

azaroth42:
That is a very cool addition, :+1: from me too. Thanks @barnaclebarnes!

PR #820, opened 2016-04-27
Search API is now 1.0

tomcrane:
This keeps 0.9 - but maybe it should redirect?

### DIFF
--- a/source/api/annex/notes/editors.md
+++ b/source/api/annex/notes/editors.md
@@ -16,7 +16,7 @@ This document is not subject to [IIIF's versioning semantics][iiif-semver]. Chan
 
   * **[Michael Appleby](https://orcid.org/0000-0002-1266-298X)** [![ORCID iD](/img/orcid_16x16.png)](https://orcid.org/0000-0002-1266-298X), [_Yale University_](http://www.yale.edu/)
   * **[Tom Crane](https://orcid.org/0000-0003-1881-243X)** [![ORCID iD](/img/orcid_16x16.png)](https://orcid.org/0000-0003-1881-243X), [_Digirati_](http://digirati.com/)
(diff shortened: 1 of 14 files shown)


jpstroop:
Not sure what the right thing is regarding redirect. We don't have any or 0.x.y releases published, which makes me tempted to say we should remove 0.9/index.md (maybe doing `git mv 0.9/index.md 1.0/index.md` would be better, so that we keep the history?) and add to the top?

```yaml
redirect_from:
  - /api/search/0.9/index.html
```

azaroth42:
I'd prefer to copy rather than move. There is a 0.9 for metadata/presentation:
  * http://iiif.io/api/metadata/0.9/

And there was 0.9 for Image, I just never found the write up of it to port from where ever it lived into iiif.io.

So :+1: from me as is.

zimeon:
I also say copy rather than mode. :+1: as is.

What do we do about the link checker and <http://www.stanford.edu/~azaroth> having been very efficiently taken down (why 403 though?)

zimeon:
Seems like persistent ids for people would be a nice idea, ones that stay with you when you switch jobs... hmm... ORCID :-)

azaroth42:
Lets drop the link to my dead homepage. I don't even know what the Getty does about personal homepages, let alone having one yet.

tomcrane:
Hi @azaroth42 - shall I replace with link to your ORCID page? Or no homepage for now?

mikeapp:
:+1: copy

zimeon:
We can change again as necessary, but I have switched `http://www.stanford.edu/~azaroth/` -> `https://orcid.org/0000-0003-4441-6852` everywhere.

azaroth42:
+1 for changing to the orcid, thanks all!

PR #822, opened 2016-04-28
keep search 0.9 out of sitemap

jpstroop:
I was going to say over on #820 that I'm :ok: with the 'copy and not redirect' approach, but that we should at least keep 0.9 out of the sitemap. This does that.


tomcrane:
:+1:

mikeapp:
:+1:

zimeon:
:+1:

azaroth42:
:+1: and will raise an issue to discuss what /should/ go into the sitemap.

PR #826, opened 2016-04-28
Emma Stanford's keen-eyed typos

### LINKED ISSUE #825: Typos from Emma Stanford
_Image API_
* 5.2: "This inconsistency will be addressed in the next major version of this specification" should have a period at the end.
* 7.2: I'd like the error code description syntax in the table to be more uniform. I.e., 503 should just be "The server is busy/temporarily unavailable..."

_Presentation API_ 
* 1.1: "The discovery or selection of interesting digitized objects is not directly supported, however hooks to reference further resources are available"--the comma should be a semicolon, as in the next sentence.
* 5.1: "stuctures" should be "structures".
* 5.3: "The canvas's dimensions should be double that of the image"--I would say "double those of the image", unless I'm misunderstanding this sentence.
* 5.7: "All of the properties of the Layer": "layer" shouldn't be capitalized.
* 6.7: "Unlike painting annotations, comments or annotations with other motivations, SHOULD"--I would remove the comma before "SHOULD"

_Search API_
* 3.3.1: "dereferencable" should be "dereferenceable"
* 3.4: "Examples include the text before and after the matched content to allow a result snippet to be presented, the matched text in case stemming or wildcards have been applied, or to link annotations that together fulfill the search query for a phrase, such as when the phrase spans across lines of text." I'm having trouble parsing this sentence, I think because the third example doesn't match the syntax of the first two.
* 3.4.2: "This is most useful when the service has word-level boundaries of the text on the canvas, such as available when OCR has been used to generate the text positions." Should be "such as are available"? Also, the term "OCR" occurs here before it is given in full (i.e., "Optical Character Recognition (OCR)") in 3.4.4.
* 3.4.4: "The first line is “A bird in the hand”, the second line was “is worth two in the bush”": I would change "was" to "is". I would also change "will" to "would" in "If the annotations were instead at word level, then phrases will always require multiple annotations."
* 4.2. Response: This section should surely be numbered 4.3, as there is already a 4.2? Also, "internationaliation" should be "internationalization", and "occured" should be "occurred".
* 5: "anotation" should be "annotation" in several cases.

azaroth42:
Fixes all of them, closes #825


zimeon:
:+1:

mikeapp:
:+1:

tomcrane:
:+1:

Issue #824, opened 2016-04-28
What should go into the sitemap for http://iiif.io/?


azaroth42:
I propose _only_ putting in the current specs to the sitemap. Then googlers will come to the right place, rather than ending up at the metadata 0.9 API and thinking it's somehow current.

Meaning the sitemap look at the site level variables to determine what "current" is, rather than having to re-flag each page? (Not that that's a big deal given rate of spec change)

jpstroop:
> I propose only putting in the current specs to the sitemap. Then googlers will come to the right place, rather than ending up at the metadata 0.9 API and thinking it's somehow current.

Yes. This is how it's currently set up.

>Meaning the sitemap look at the site level variables to determine what "current" is, rather than having to re-flag each page? (Not that that's a big deal given rate of spec change)

Much easier to just use the sitemap plugin, I think.

jpstroop:
Oh, one amendment/clarification: I think beta specs and latest stable should both be in the sitemap (as it is now), e.g.:
``` xml
<url>
  <loc>/api/presentation/2.0/</loc>
</url>
<url>
  <loc>/api/presentation/2.1/</loc>
</url>
```

mikeapp:
Would it be worth blocking crawlers from older spec versions via `robots.txt`?  I assume that as links to older versions accumulate they will still be found and indexed by Google.

jpstroop:
Sure. Wonder if we build  the robots.txt w/ grep & awk, looking for `sitemap: false`?

*accepts challenge*

azaroth42:
Close no-one-cares?

zimeon:
I think close-done, we have implemented control with Jekyll settings and, IMO, the discoverability is fine:

```
> grep -r sitemap: source
source/api/image/1.0/compliance.html:sitemap: false
source/api/image/1.0/index.html:sitemap: false
source/api/image/1.0/rotation_calc.html:sitemap: false
source/api/image/1.1/change-log.html:sitemap: false
source/api/image/1.1/compliance.html:sitemap: false
source/api/image/1.1/index.html:sitemap: false
source/api/image/2.0/change-log.md:sitemap: false
source/api/image/2.0/compliance.md:sitemap: false
source/api/image/2.0/index.md:sitemap: false
source/api/metadata/0.9/index.html:sitemap: false
source/api/metadata/1.0/index.html:sitemap: false
source/api/presentation/2.0/change-log.md:sitemap: false
[...]

PR #827, opened 2016-04-28
Script to generate robots.txt

jpstroop:
As mentioned in #824. With a little finagling @eefahy might be able to hook this into the cap deploy.

### DIFF
--- a/scripts/generate_robots_txt.sh
+++ b/scripts/generate_robots_txt.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+HERE=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+SOURCE=$HERE/../source
+OUT=$SOURCE/robots.txt
+
+cd $SOURCE
+echo "User-agent: *" > $OUT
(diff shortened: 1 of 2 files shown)


azaroth42:
:+1: Haven't tested, but looks good. And at worst harmless.

mikeapp:
:+1: looks good

zimeon:
:+1: looks good

zimeon:
Changing to :-1: on the idea of having `Disallow:` for old versions of the specs. While this might help with our presence in Google, we are also saying to web archiving systems etc. that we don't want the old specs harvested -- when we do. @azaroth42 says that IA uses `robots.txt` both to control harvesting and as a retrospective filter on what is visible, so adding `Disallow` for the old specs would vanish them from IA.

zimeon:
See also #838

mikeapp:
Also :-1: due to archiving concerns.

azaroth42:
Agree that the web archiving issue is problematic. :-1:

jpstroop:
Yup, 👎 based on our discussion. Closing.

PR #830, opened 2016-04-29
net/http and conditional Google ping

### LINKED ISSUE #828: Google notify fails due to Net::HTTP
With the following:

```
(Backtrace restricted to imported tasks)
cap aborted!
NameError: uninitialized constant Net::HTTP

Tasks: TOP => notify:google
(See full trace by running task with --trace)
The deploy has failed with an error: uninitialized constant Net::HTTP
```

This happens after the site is built, so we can still deploy, but would be good to fix sooner rather than later.

@barnaclebarnes any ideas?

jpstroop:
Closes #828

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,6 +1,8 @@
 # config valid only for current version of Capistrano
 lock '3.4.0'
 
+set :log_level, :info
+
 set :application, 'iiif.io'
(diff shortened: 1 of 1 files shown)


jpstroop:
We'll need to merge and deploy to know if this works

PR #835, opened 2016-05-12
Fix schedule to match reality

zimeon:
Move Mark/JHU from today to yesterday when presentation was given. Fix times on Friday.


jpstroop:
👍 going to merge

PR #845, opened 2016-05-15
Minor site updates + announcement

jpstroop:
(I'll keep pushing stuff here as separate commits)


jpstroop:
@IIIF/editors this adds an announcement about the new specs. It's a little sparse...see last commit.

jpstroop:
Need latest redirect-from gem. Will get to it this afternoon.

PR #869, opened 2016-06-24
image api context dc.rights -> dc.license

### LINKED ISSUE #868: Change "license" to map to dct:rights - redux for Image API
While checking the rights/license fields of IIIF, I encountered an oddity: both Image API and Presentation API use "license" in the JSON, as agreed in a previous discussion [1]. But the JSON contexts differ:
- the Image API context [2] maps "license" to dcterms:license
- the Presentation API context [3] maps "license" to dcterms:rights
My understanding of [1] is that the two contexts should have been updated so that "license" is mapped to dcterms:rights in both. If I'm right, the PR [4] was only a partial fix and should be extended to the Image API context [3].
First question is: am I right?

[1] https://github.com/IIIF/iiif.io/issues/636
[2] http://iiif.io/api/image/2/context.json
[3] http://iiif.io/api/presentation/2/context.json
[4] https://github.com/IIIF/iiif.io/pull/657/

jpstroop:
Fixes #868

### DIFF
--- a/source/api/image/2/context.json
+++ b/source/api/image/2/context.json
@@ -136,7 +136,7 @@
     },
     "license": {
       "@type": "@id",
-      "@id": "dcterms:license"
+      "@id": "dcterms:rights"
     },
     "logo": {
       "@type": "@id",


azaroth42:
:+1:

tomcrane:
:+1:

PR #891, opened 2016-08-17
add the iiif slack signup form to community page

mejackreed:


mikeapp:
👍  Travis error is due to an external website being down.

PR #895, opened 2016-08-21
must -> aims to

jpstroop:
These bullets feel like a mix of challenges and requirements. I don't think we need separate lists (though maybe?) because we lay out our goals in the intro, but if this is a list of challenges, we should tighten it up and remove our aspirations.


azaroth42:
:+1: to tightening it up. aims to seems like a good start.

PR #900, opened 2016-08-24
Revise intro

jpstroop:
Trying to tease apart goals and challenges.

### DIFF
--- a/source/api/auth/0.9/index.md
+++ b/source/api/auth/0.9/index.md
@@ -52,15 +52,15 @@ Providing interoperable content through client applications running in a web bro
 
 * A single IIIF Presentation API manifest can reference content resources at multiple institutions and hence from multiple domains.
 * Institutions have different existing access control systems.
-* Most IIIF viewers are client-side JavaScript applications, and may be served from a domain that is different, and thus untrusted, from the image services that it is required to load.
- * Similarly, the domain of the authentication services that protect IIIF-based services may be different from that of a viewer or IIIF resources. Therefore, the authorizing server must not require any prior knowledge of the domain hosting the viewer.
+* Most IIIF viewers are client-side JavaScript applications, and may be served from a domain that is different from, and thus untrusted by, the image services that it is required to load.
+* Similarly, the domain of the authentication services may be different from that of a viewer or the IIIF-based content. Therefore, the authorizing server must not require any prior knowledge of the domain hosting the viewer.
 
 Additionally, the IIIF community has the following goals for this specification:
 
-* A IIIF client should not ask for or accept any credentials itself; the server hosting the content must be responsible for capturing credentials from a user and the IIIF viewer needs no knowledge of or access to this exchange.
-* A browser-based IIIF client must be able to maintain its state during an authentication flow.
-* No registry of trusted domains is required. Anyone should be able to create any kind of viewer and run it from anywhere.
-* Institutions should not have to adopt a new authentication system to participate.
+* A IIIF client should not authenticate the user itself; the server hosting the content must be responsible for capturing credentials from a user and the IIIF viewer needs no knowledge of or access to this exchange.
+* A browser-based IIIF client must be able to maintain its own internal state during an authentication flow.
+* A registry of trusted domains should not be required; anyone should be able to create any kind of viewer and run it from anywhere.
+* Institutions should be able to use their existing authentication systems without modification.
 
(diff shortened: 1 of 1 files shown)


zimeon:
:+1:, significant improvement in clarity

source/api/auth/0.9/index.md
```diff
@@ -46,17 +46,23 @@ This is a work in progress. We are actively seeking implementations and feedback
...
-* Institutions have different existing access control systems and should not have to adopt a new one to participate.
-* A IIIF client can be a JavaScript viewer served from a different domain from the image services, and the authentication services that protect them. This domain is _untrusted_ - the authorizing server must not require any prior knowledge of the domain hosting the viewer. The specification aims to not introduce or require any registry of trusted IIIF viewer domains and must assume that for image delivery, anyone can create any kind of viewer and run it from anywhere.  
+* Institutions have different existing access control systems.
+* Most IIIF viewers are client-side JavaScript applications, and may be served from a domain that is different, and thus untrusted, from the image services that it is required to load.
+ * Similarly, the domain of the authentication services that protect IIIF-based services may be different from that of a viewer or IIIF resources. Therefore, the authorizing server must not require any prior knowledge of the domain hosting the viewer.
```
azaroth42: IIIF-based services? Seems unnecessarily specific/technical for an introduction?

source/api/auth/0.9/index.md
```diff
@@ -46,17 +46,23 @@ This is a work in progress. We are actively seeking implementations and feedback
...
+
+Additionally, the IIIF community has the following goals for this specification:
+
 * A IIIF client should not ask for or accept any credentials itself; the server hosting the content must be responsible for capturing credentials from a user and the IIIF viewer needs no knowledge of or access to this exchange.
```
azaroth42: Is this still true? The client does ask for the Access Token, which we now think of as a credential (right?)

sdellis:
> Institutions should not have to adopt a new authentication system to participate.

Can we expand on this?  It would make everyone's lives easier to know exactly what is required from an authentication system up front, or even better, what minimum versions of common authentication systems would be required?  For example, as far as I know, Princeton would have to get OIT (Central IT) to upgrade our CAS system to be able to support the requirements outlined in the spec.  Are there any implementation examples of the Auth API for any common authentication systems?

source/api/auth/0.9/index.md
```diff
@@ -46,17 +46,23 @@ This is a work in progress. We are actively seeking implementations and feedback
...
 * A browser-based IIIF client must be able to maintain its state during an authentication flow.
+* No registry of trusted domains is required. Anyone should be able to create any kind of viewer and run it from anywhere.
```
azaroth42: A registry of trusted domains should not be required.  (to follow the pattern of the other bullets)

jpstroop: I can make changes later today, or someone else can take over.

-Js

Sent via mobile. Please excuse typos, brevity, etc.

source/api/auth/0.9/index.md
jpstroop: Yes, agree

mikeapp:
@sdellis  The spec does not require central IT to make any changes to the CAS server.  Your IIIF Access Cookie Service (section 2.1) that follows the Login profile would present a page that is protected by CAS or any other mechanism(s) you choose.   There's no requirement for the CAS server itself to be aware that it is authenticating the user for access to a page that is part of the IIIF infrastructure.

mikeapp:
:+1: overall, including Rob's changes

sdellis:
@mikeapp I believe that the [CAS RESTful API](https://wiki.jasig.org/display/CASUM/RESTful+API) must be activated and configured to issue a Ticket Granting Ticket (that we can use as an accessToken).  Are you assuming that is the case?  Unfortunately Princeton does not have that configured at the moment, so I think Central IT would need to make these changes for us to be able to implement the Auth API.

mikeapp:
@sdellis The spec does not assume a TGT mechanism is available from the SSO.  The auth application can easily manage its own cookie/token/user associations.

sdellis:
Ok, but there is an assertion that Description Resources use tokens, not cookies (section 1.3).  Correct me if I'm wrong, but without the RESTful API I believe CAS assumes a browser and the token is wrapped up in a cookie.  I'm not going to beat what seems to be a dead horse, but I am just wary of the assumption that an authentication service should not need to change, especially when that service is often managed by another institutional department that may have different priorities.

azaroth42:
And, as Mike said, that token is provided by a IIIF specific implementation, not by the institutional authentication service.

zimeon:
@sdellis - not requiring a new institutional authentication service is a fundamental requirement so we certainly must explore any concern about that. However, since, as @azaroth42 says, the token is provided by the IIIF auth service, I don't think the requirement is violated.

sdellis:
Thanks for the clarification, all.  I get it now.

PR #905, opened 2016-08-25
Bold introduction of Content and Description Resources

jpstroop:
Looking at the __Terminology__ sections of the other specs, we've never defined terms before, so this is uncharted territory, but I think this makes sense, especially since we bold __access cookie__ and __access token__ in 1.2/1.3.

Related: should we define those __access cookie__ and __access token__ in 1.1? I think it makes sense to do so in context (as we do now), but it's not terribly consistent. For now I've added a sentence to the end of 1.1, but it's in a separate commit that I can step back if we disagree.

### DIFF
--- a/source/api/auth/0.9/index.md
+++ b/source/api/auth/0.9/index.md
@@ -75,19 +75,19 @@ Please send feedback to [iiif-discuss@googlegroups.com][iiif-discuss].
 
 ### 1.1. Terminology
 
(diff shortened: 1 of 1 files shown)


zimeon:
:+1: to both changes. (aside: why is double underscore better than double asterisk for bold?)

jpstroop:
> why is double underscore better than double asterisk for bold

I didn't really think about it, but I think we've generally tended to use `__`.

mikeapp:
:+1:

PR #912, opened 2016-09-07
added link to events on community page

srabun:
See http://link_to_events.iiif.io/community/ so people can actually get to the Events page


azaroth42:
:+1:

jpstroop:
Happy to merge. Since that list is so long now, I wonder if it should be last on the page?

zimeon:
JavaScript to expand/collapse list of institutions (show number when collapsed)?

jpstroop:
Looking for emoji w/ crinkled nose :smile:  Maybe I'm misunderstanding--do you have an example from another site?

PR #916, opened 2016-09-13
ncsu news announcement

srabun:
Announcing NCSU joining consortium plus Nubian Message implementation and other search inside collections.

http://ncsu_news.iiif.io/news/

http://ncsu_news.iiif.io/news/2016/09/12/ncsu-joins-iiifc/

### DIFF
--- a/source/_posts/2016-09-12-ncsu-joins-iiifc.md
+++ b/source/_posts/2016-09-12-ncsu-joins-iiifc.md
@@ -16,9 +16,9 @@ Since the initial push to provide "search inside" functionality for the Nubian M
 
 Welcome and congratulations to NCSU Libraries!
 
-[image-api]: http://iiif.io/api/image/2.1/
-[presentation-api]: http://iiif.io/api/presentation/2.1/
-[search-api]: http://iiif.io/api/search/1.0/
+[image-api]: http://iiif.io/api/image/
+[presentation-api]: http://iiif.io/api/presentation/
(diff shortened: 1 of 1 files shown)


azaroth42:
Very nice! :+1: from me

mikeapp:
One question/comment: the links to the APIs go to specific versions, not sure if it would be better to link to the latest stable or just `/` and let the redirect do its thing.

srabun:
@mikeapp will the current links break when new specs are published? I don't mind changing the links if that's preferred

jpstroop:
I don't feel strongly about the links either way--they won't break with new specs. :+1: overall!

zimeon:
Since the text says just "IIIF APIs", I'd go with unversioned links, otherwise :+1:

srabun:
@zimeon So there is the "IIIF APIs" link but also above that in the first paragraph there are links to the Image, Presentation, and Search APIs specifically...

mikeapp:
@srabun  For example, the url `http://iiif.io/api/search/` will always redirect to the latest version of the API.  If you link to a specific version it will eventually be superseded by a new version and become obsolete.

srabun:
Roger. I will go ahead and update the links so we will have our bases covered

azaroth42:
All comments resolved with +1s, merging...

srabun:
Ok, I made the changes - links to specific APIs are now un-versioned

PR #918, opened 2016-09-16
mirador2.1 release news

srabun:
News announcement for Mirador 2.1 release

See http://mirador_news.iiif.io/news/2016/09/16/mirador-upgrade-released/
and http://mirador_news.iiif.io/news/


zimeon:
:+1: looks good.

Sometime down the road we might have to think about IIIF-C rules for what should or should not be accepted as a news item on the IIIF site. Clearly an announcement like this would be above the line, but I could see questions about: projects from non-members, projects that don't really implement the specs properly/well, projects very commercially driven, attempts to make very frequent posts, etc..

srabun:
Thanks! Agreed @zimeon for now, if I get a questionable request, I'll float it on the CoCo channel first.

jpstroop:
:+1:

PR #919, opened 2016-09-16
linking sched to hague event page

srabun:
Linked sched to The Hague event page, also added "Week of June 5" to The Vatican listing.

See http://hague_schedule.iiif.io/event/

And: http://hague_schedule.iiif.io/event/2016/thehague/


azaroth42:
:+1:

PR #922, opened 2016-09-21
Fix lineheight for RFC highlighted text

### LINKED ISSUE #606: Fix CSS so that RFC words do not create extra space above and below the line
See, for example, <http://iiif.io/api/image/2.0/#error-conditions> where there is a single _RECOMMENDED_. Adjust window width so that this word is on the third line and you'll see extra space before and after that line compared with between first and second (normal text).

zimeon:
Fixes #606. Have checked on current Chrome/Firefox/Safari on Mac.

### DIFF
--- a/source/css/spec-doc.scss
+++ b/source/css/spec-doc.scss
@@ -276,7 +276,8 @@ pre {
   color: #d55;
   font-variant: small-caps;
   font-style: normal;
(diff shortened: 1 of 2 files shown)


jpstroop:
:+1: I pulled it down and it looks fine,

Issue #894, opened 2016-08-18
Fix css for interaction ^H^H^H^H profile table in auth


azaroth42:
first child is backticked

zimeon:
See first table in http://iiif.io/api/auth/0.9/#service-description

zimeon:
There is a bit of a mess in the definitions for `.api-table` in `source/css/main.scss` and `source/css/spec-doc2.scss` -- really it would make sense to rationalize this but I don't want to go messing with too many moving parts so I prefer an additive change that we can be pretty sure won't break up old docs.

azaroth42:
Fixed by #925

PR #925, opened 2016-09-21
Add class first-col-normal for tables where we don't want monospace

zimeon:
Fix for #894: Additive change to CSS to avoid breaking other things. Use is:

```
| Table | In | Here |
| ----- | -- | ----- |
| Who | Knows | What? | 
{: .api-table .first-col-normal }
```

and `Who` will not be monospaced. Applied to Auth spec for [patterns table](http://iiif.io/api/auth/0.9/#service-description) and [change log](http://iiif.io/api/auth/0.9/#d-change-log)


azaroth42:
:+1: Deployed at http://css_first_col_normal.iiif.io/api/auth/0.9/ and looks great

jpstroop:
:+1:

PR #931, opened 2016-09-22
limit ci link checking to internal links only

### LINKED ISSUE #930: Remove link checker from travis tests on commit (or make it check internal links only)
I think it would be valuable to have a periodic link check run on the iiif.io site but in the current form it is either an inordinate waste of time (have to check for bogus failure every build) or, worse, dangerous (one starts to ignore travis failures).

I suspect the best thing would be to have a periodic link check running on the live site (say) weekly that posts status to Slack. But open to other suggestions.

jpstroop:
Closes #930


zimeon:
:+1:

azaroth42:
All the :+1: s from me too. Merging on 3.

Issue #924, opened 2016-09-21
Update Auth API Implementation Notes document


zimeon:
http://iiif.io/api/auth/0.9/implementation is left a bit behind at 0.9.1 and needs update for auth 0.9.3. See also #528

zimeon:
I ripped the following out while stubbing the implementation notes section of the main auth doc (#921). I think these should be covered somewhere in the implementation notes document:

  * Care is required to implement this specification in a way that does not expose credentials thus compromising the security of the resources intended to be protected, or other resources within the same security domain.
  * Implementations must not reuse the access cookie value (or any simple transformation of it) as the access token value. Access token values could be copied across domains if the access token were to be obtained by a malicious client.
  * This specification is modelled after elements of the OAuth2 workflow and the [OAuth2 Security Considerations](https://tools.ietf.org/html/rfc6750#section-5) section provides useful additional guidance regarding threats, mitigations and recommended practices.

zimeon:
I think we should say which version of the spec the Implementation Notes apply to but we should not version the implementation note _per se_, as we may want change it more frequently than the spec itself. I suggest including a change log where we date and comment updates, and point out alignment with minor version releases. Something along the lines of:

| Date | Description |
| --- | --- |
| 2017-01-01 | Comments about new version of Zirconium browser on ZeeOS |
| 2016-10-19 | Release with IIIF Authentication API v1.0 |

and then the **Status of this Document** section at the top should simply say which version of the auth spec it applies to.

azaroth42:
Close?

zimeon:
Agree close, completed with #921, #934 and #943

Issue #923, opened 2016-09-21
Clarify Auth spec might need more for REST of description resources


azaroth42:
Eds are confident that the Auth spec protects content resources (images, audio, video, etc) retrieved via regular HTTP interactions from the browser.  However when REST patterns are established, further issues that we haven't thought of or implemented may come up with authentication of the creation, updating and deletion of description resources.  These issues may result in changes to the Auth specification, however it is anticipated that these changes will be backwards compatible with 1.0.

Something like this should go in the new 1.4.

zimeon:
I think this been dealt with adequately by second para of http://iiif.io/api/auth/0.9/#security -- propose close.

azaroth42:
:+1: to close, fixed.

jpstroop:
Just want to register that I don't feel completely confident that we can guarantee backward compatibility, based on how this has gone so far, but I think the language is loose enough that we're OK. :+1: to closing.

tomcrane:
:+1:

Issue #528, opened 2015-07-28
Mobile (IOS) without popups?!


azaroth42:
Can't open a popup window in IOS browsers, and can't generate a new tab without a specific user action.  Popup blockers are just on or off, not per domain. 

Should write an auth implementation note to capture these issues and any workarounds.

azaroth42:
@tomcrane Did you find a workaround for this, and if so can you document it as a note?

tomcrane:
There is no workaround as such - a client could do something like attempt to ```window.open(strUrl, strWindowName, strWindowFeatures)``` with strWindowFeatures having explicit dimensions (popup) and if that errors or results in a null reference, fall back to a new tab with ```window.open(strUrl, strWindowName)```. We chose not to do this for UX consistency across devices, but other clients could try it.

I think I will enlarge Appendix A substantially and reference it in the main text - or would a separate document be better for impl notes? I prefer a single coherent document. Need to make sure that a client implementer isn't led into trying something by the main text only to be told they might not want to do it that way in the implementation note.

It will be next week before I can spend some time on this. I have the basis of the notes scattered around Google docs and emails. We can ask @edsilv and @aeschylus to review.

azaroth42:
Can close when we add (accept) implementation notes from @tomcrane that explains the workflow for mobile.
  http://auth_notes.iiif.io/api/auth/0.9/implementation/#window-behaviour-across-devices

azaroth42:
I think we can close, now that implementation notes are merged?

zimeon:
Implementation note is live at http://iiif.io/api/auth/1.0/implementation/, with specific mention of new window issue in http://iiif.io/api/auth/1.0/implementation/#event-initiation . Agree close

azaroth42:
Closing per above.

PR #934, opened 2016-09-25
update implementation notes to match spec

tomcrane:
for #528, #923, #924 

I have also added terms to the context for header, confirmLabel etc - but can any of them reuse vocab from elsewhere?


jpstroop:
:+1: to merging and then working on some more discreet issues.

azaroth42:
:+1:

PR #939, opened 2016-09-27
Fix search contexts

### LINKED ISSUE #938: Search API context is missing wrapping `"@context": { .. }`
The Search API context: http://iiif.io/api/search/1/context.json is not a valid context, it needs `{ "@context": { ... } }` around it

zimeon:
Fixes #938

### DIFF
--- a/source/api/search/0/context.json
+++ b/source/api/search/0/context.json
@@ -1,17 +1,16 @@
----
----
 {
-	"search": "http://iiif.io/api/search/0#",
+  "@context": {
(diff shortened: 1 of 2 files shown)


jpstroop:
I don't have a way to test, but seems correct. :+1:

azaroth42:
:+1: from me too

tomcrane:
👍 When live, a chunk of Search JSON-LD will work properly in http://json-ld.org/playground/

mikeapp:
👍

PR #944, opened 2016-09-28
Fix line spacing for alert and note blocks, add space after strong

zimeon:
It seems that the line spacing for the `<strong>` in `.warning` blocks was fixed sometime earlier but has been broken (leading to large space to following line) for `.alert` and `.note` blocks. Change fixes that and adds a little extra space between the end of the `<strong>` block and following text because we don't seem to put any punctuation in there.

### DIFF
--- a/source/css/spec-doc2.scss
+++ b/source/css/spec-doc2.scss
@@ -129,6 +129,7 @@
       strong {
         font-weight: bold;
         color: #E05252;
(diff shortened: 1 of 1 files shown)


mikeapp:
👍

jpstroop:
👍

azaroth42:
LGTM :+1:

PR #948, opened 2016-10-02
Search API link from Presentation

tomcrane:
It's not "a future IIIF Specification" any more


mikeapp:
👍

azaroth42:
:+1: non-normative fix to a spec.  Should we accumulate for 2.1.1 ?

jpstroop:
Yes. 2.1.1

zimeon:
Agree, 2.1.1. Is wording right? Perhaps change s/search ... is described by the IIIF Search API/search ... is supported by the IIIF Search API/ ?

PR #954, opened 2016-10-17
updated staff section

srabun:
Updated the staff section of the Consortium page: http://consortium_staff_update.iiif.io/community/consortium/


zimeon:
:+1:

PR #955, opened 2016-10-17
added EPFL lightning talk

srabun:
added "Time Machine FET Flagship" by Frederic Kaplan lightning talk to the end of the lightning round for The Hague: http://add_lightning_talk.iiif.io/event/2016/thehague/


zimeon:
A bit late :-)

PR #962, opened 2016-10-24
Redirect from Auth API profile URIs to spec

### LINKED ISSUE #961: Redirect Auth API profile URIs to the spec
It would be nicer if the Auth API profile URIs took a user somewhere useful to understand what they mean. Best option seems to be a redirect to the appropriate portion of the spec, e.g.:

http://iiif.io/api/auth/0/login -> http://iiif.io/api/auth/0.9/#login-interaction-pattern

zimeon:
Fixes #961

### DIFF
--- a/source/api/auth/0/clickthrough.md
+++ b/source/api/auth/0/clickthrough.md
@@ -1,0 +1,5 @@
+---
+permalink: /api/auth/0/clickthrough/index.html
+redirect_to: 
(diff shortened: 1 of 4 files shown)


azaroth42:
:+1: Works as expected: 
http://auth_profile_redirects.iiif.io/api/auth/0/login

jpstroop:
:+1:

PR #967, opened 2016-10-26
Use site vars for API version links

zimeon:
Found "Presentation API" in http://iiif.io/api/annex/notes/jsonld/#frames linked to 2.0, changed this and example code to use Jekyll site vars to get current versions.


azaroth42:
:+1:

PR #969, opened 2016-10-28
added Chinese University of Hong Kong to IIIF-C lists

srabun:
Chinese University of Hong Kong has paid and we just now received their MOU - adding them to the official list


jpstroop:
:+1:

PR #973, opened 2016-10-31
Av group page update

### LINKED ISSUE #950: update A/V group page with link to google drive
From Jason: One thing that would be nice to do is to link in or move the historical documents about the BL meeting on A/V into this folder as that's an important document for us. I'd also like the A/V community page to get a link directly into the working group folder once things are set.

srabun:
In an effort to provide consistent information across all IIIF groups (the A/V group was first to comply) - working with Jason, I added additional info to the group page to make it easier to learn more and get involved with the group - see changes http://av_group_page_update.iiif.io/community/groups/av/ 
Closes #950

### DIFF
--- a/source/community/groups/av/index.md
+++ b/source/community/groups/av/index.md
@@ -22,11 +22,11 @@ The IIIF A/V Technical Specification Group aims to extend to A/V the benefits of
 [iiif-conf-2016]: /event/2016/newyork/#wednesday "IIIF Conference 2016"
 [events]: /event "IIIF Events"
 [iiif-discuss]: https://groups.google.com/forum/#!forum/iiif-discuss "IIIF-Discuss Forum"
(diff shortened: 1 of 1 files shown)


zimeon:
I'm a little nervous about the possible churn related to putting call details in, but it does make it easier for folks to find details. :+1:

srabun:
@zimeon churn as in, if call connection or call times change? I'm trying to get everyone to use a recurring Bluejeans meeting, so once the links are solid, the only problem would be if there is a one-off change due to people being at a conference or something - I'm hoping that the calendar will also help to mitigate, so pointing people to the calendar...I am willing to keep things up to date as much as possible. I have gotten a few emails from people wanting to get involved and being really confused as to how to join the groups...if the call info does start to create more problems than it's worth I can take it down. If nothing else, the link to slack channel and google drive folders can point people in the right direction

PR #995, opened 2016-11-14
added first newsletter to news

srabun:
I put the first IIIF Community Newsletter in the News section so as to provide a publishing platform to point people to...I was able to check my changes, links, etc. locally, but now I'm a bit confused as to why the post isn't showing up for me here: http://news_newsletter.iiif.io/news/  - file saved in iiif.io/source/_posts/2016-11-15-newsletter.md

Aiming to send newsletter out via email and point people to news posting on Nov. 15...Thanks!

### DIFF
--- a/source/_includes/head.html
+++ b/source/_includes/head.html
@@ -11,4 +11,7 @@
     <link rel="stylesheet" href="/css/main.css">
     <link type="application/atom+xml" rel="alternate" href="/news/atom.xml"/>
     <script src="/js/vendor/modernizr-2.6.2-respond-1.1.0.min.js"></script>
+    <script src="https://code.jquery.com/jquery-2.2.3.min.js" integrity="sha256-a23g1Nt4dtEYOj7bR+vTu7+T8VP13humZFBJNIYoEJo=" crossorigin="anonymous"></script>
+    <script src="/js/vendor/anchor.min.js"></script>
+    <script src="/js/anchor.js"></script>
   </head>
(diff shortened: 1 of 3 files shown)


jpstroop:
@srabun I made some tweaks to the site (didn't change any content other than the markup on the  headings) that should help this present a little better:

 * Added a shorter excerpt for the news index page. Note that this syntax (`:>`) allows you to split YAML content over multiple lines. You may want to change the text, of course, but I think something like this is better than the `h2` tag that was showing up by default.
 * Moved all of the headings down one level, since the post template makes an `h1` tag with the page title ("IIIF Community Newsletter, Volume 1 Issue 1") in this case.
 * Added a CSS rule for post headings to be bold, so that all of the `**` following `#` could go away in the markdown, and tightened up the padding on lists.
 * Added the Javascript that makes heading linkable

As I said, you may want to change the excerpt, but this should give you this gist. I pushed up those changes and deployed:

 * news page: http://news_newsletter.iiif.io/news/
 * post page: http://news_newsletter.iiif.io/news/2016/11/15/newsletter/

Note that my machine had the css cached and I had to do a hard reload (command + shift + r on a mac, I believe) to get it to refresh.

srabun:
Looks good to me thank you @jpstroop  - good calls.

azaroth42:
:+1:  Perhaps slight tweak might be to link "Slack" like -discuss, rather than having an explicit Join Slack link? Not a big deal though, feel free to merge as is :)

srabun:
Agreed @azaroth42 thank you

jpstroop:
I'd rather not merge since I made changes, but @srabun wanted this yesterday, so pulling the trigger.

PR #1000, opened 2016-11-22
added link to international numbers for av, museums, and devs groups

srabun:
In updating the manuscripts and newspapers group pages I realized it's better to include 1 US number and then link to the international numbers like we do with community calls - so I went ahead and updated Av: http://international_call_info.iiif.io/community/groups/av/ , Museums: http://international_call_info.iiif.io/community/groups/museums/ , and Software devs: http://international_call_info.iiif.io/community/groups/software/

Once newspapers and manuscripts are merged, they will all have the same language in the call connection field.


jpstroop:
PR 1000!

jpstroop:
@srabun You might consider making this an 'include'. See, e.g.,  at the end of [this section](https://github.com/IIIF/iiif.io/blob/master/source/api/image/2.1/index.md#status-of-this-document). The line `{% include copyright.md %}` corresponds to the file `source/_includes/copyright.md`, which gets copied in when the site is built. It keeps you from having to maintain the info in multiple places. You could make something like `call_info.md` and do similar.

zimeon:
I have to get in on PR1000... agree with Jon that an include would likely make maintenance easier for you

azaroth42:
Me too!!111one!
In before the lock.

mikeapp:
👍

zimeon:
I went ahead and merged based on agreement above, seems that the question of using an include need not get in the way of getting this out (doesn't change work required)

PR #1004, opened 2016-11-28
updated homepage news to newsletter

srabun:
updated home page news item to point to the IIIF Community Newsletter - see http://homepage_news.iiif.io/


jpstroop:
Doh, sorry, missed this. Will merge.

PR #1008, opened 2016-12-17
added names to call moderator rotation

srabun:
per community call guidelines, added additional CoCo members and group leaders to call moderator rotation script, with exception of those who opted out.

scripts/call_calendar_generator.rb


jpstroop:
I wonder if it's worth asking CoCo members which category they prefer--I think a few other people could maybe do tech calls.

PR #1009, opened 2016-12-19
added policy for grant support letters

srabun:
please review. I couldn't test on dev, got the following after `cap production deploy`:

....
tar: source/model/shared-canvas/1.0/index.html: Cannot open: No such file or directory
tar: source/model/shared-canvas/1.0/sharedcanvas.css: Cannot open: No such file or directory
tar: source/model/shared-canvas/2.0: Cannot mkdir: No such file or directory
tar: source/model/shared-canvas/2.0/index.md: Cannot open: No such file or directory
tar: source/news: Cannot mkdir: No space left on device
tar: source/news/atom.xml: Cannot open: No such file or directory
tar: source/news/index.html: Cannot open: No such file or directory
tar: source/robots.txt: Cannot write: No space left on device
tar: source/technical-details.html: Cannot write: No space left on device
tar: spec: Cannot mkdir: No space left on device
tar: spec/source: Cannot mkdir: No such file or directory
tar: spec/source/syntax_spec.rb: Cannot open: No such file or directory
tar: spec/spec_helper.rb: Cannot open: No such file or directory
tar: tmp: Cannot mkdir: No space left on device
tar: tmp/.htmlproofer: Cannot mkdir: No such file or directory
tar: tmp/.htmlproofer/cache.log: Cannot open: No such file or directory
tar: Exiting with failure status due to previous errors


azaroth42:
Can we put 1.0 in the URI?  policy/1.0/support/index.html ?

srabun:
how is this: http://grant_policy.iiif.io/community/consortium/policy/1.0/support/

zimeon:
:+1: on the content

However... @azaroth42 -- do we really want 1.0 in the URI? Are we going to have `/community/consortium/policy/support/` and `/community/consortium/policy/1/support/` redirect to the latest version? Are we going to update 1.0 to say "OUT OF DATE" when we push out 1.1 or 2.0. And what would 1.1 or 2.0 (or 1.0.1) mean? Isn't the wayback machine and this git repo enough for recording history of this dated document? I'm inclined to remove the notion of versioning entirely and just have a dated document.

azaroth42:
If they're going to be official documents that are referred to from elsewhere, it seems prudent to me to have them explicitly versioned.  For example, in the letter of support we reference the policy under which the letter is given ... we don't want that to change under our feet, I think?  For example, if we issue a LoS to an organization, which then submits to some multi-month grant making process, during which time we change the policy significantly ... then what can the grant making agency do with the information as to the policy URI? I think the answer is nothing ... at which point we should take it out, and the policy is just for internal community process only.

zimeon:
I would be OK to either:
  * Take the policy link out of the letters and go with simple dated document with no version number (my preference)
  * If we are going to have version numbers then we need to do appropriate maintenance (according to some plan) so that folks don't end up reading the wrong version and getting confused. Since I don't understand what major and minor versions would mean for policy like this, I would go with a single integer sequence (`/1/`, `/2/`...).

azaroth42:
I'm okay with either of those options too.  However for @srabun to send the letters we need a decision promptly and before the next CoCo call.

azaroth42:
Review requested from @IIIF/coco

zimeon:
Discussed on CoCo call 2017-01-17. Agreement that we will put the policy at a single URI and not create versioned URIs for different versions of the policy. Key point is that the policy is for our guidance although we make it public to promote transparency of IIIF. Letters will not link to the policy and thus avoid any possible race conditions/uncertainties with policy changes.

zimeon:
Failing check is just `htmllint` complaint about lack of `lang="en"` on our HTML pages

azaroth42:
Propose we ignore the htmllint check for now.

zimeon:
I agree, can someone update http://grant_policy.iiif.io/ then we can review and merge?

azaroth42:
Per slack, needs a simple index page at /policy/ that just points to support.  Then good to go.

srabun:
index added: http://grant_policy.iiif.io/community/consortium/policy/

http://grant_policy.iiif.io/community/consortium/policy/support/

srabun:
moved to /community/policy: http://grant_policy.iiif.io/community/policy/ 

although my commit is messed up

azaroth42:
Yup, everything looks good. Merging. Thanks @zimeon and @srabun!

PR #1014, opened 2016-12-29
news announcement for kyoto university library network

srabun:
See http://kyoto_univ_announcement.iiif.io/news/2016/12/29/kyoto-univ-joins-iiifc/


zimeon:
:+1: looks good, still not sure what our procedure about such things is. I think we need a way for @srabun to move ahead with stuff like this quickly and easily

azaroth42:
:+1: Will deploy

PR #1027, opened 2017-01-17
Fix lint

azaroth42:
Pages just did not have a start html tag at all, given the doctype.html include. So yeah, no wonder it didn't have a lang=en attribute.


zimeon:
I guess that if the conditions for funky IE are met then there will be two `<html>` elements in the DOM, but that IE will be OK with this? Do we actually use these conditions?

But for everything else, this is a clear win/fix. So :+1:

azaroth42:
I have no idea, and no way to check.  The conditions are a very bad way to do feature detection... if we care about this, we should use modernizr or similar instead.

zimeon:
Technical fix, merged. Clearly in the right direction even if not perfect

PR #1056, opened 2017-01-31
Design considerations wordsmithing

jpstroop:
I made a start at some wordsmithing this AM. Didn't get too far, but thought I put this up since I'm not likely to get back to it until at least Thursday. Do with it what you will :smile:.


zimeon:
Have included some of these suggestions on . I did not include the change "Select Solutions That Are as Simple as Possible and No Simpler" -> "Solutions should be as Simple as Possible and no Simpler" because I think that makes the title just a statement not an imperative like the other section. Closing in favor of #1057

PR #1058, opened 2017-02-06
Correct copyright start date to 2015

zimeon:
FIrst drafts of search and auth APIs were published in 2015 and so that should be start of year range in copyright statement. (Looks silly to say 2012-2017 to me, and misleading)


jpstroop:
:+1:

PR #1079, opened 2017-02-21
registration info added

srabun:
updated Vatican event page with links to Eventbrite registration for Showcase and Conference


zimeon:
Looks good, merged

PR #1082, opened 2017-02-27
Make TM on front page smaller

zimeon:
Seems that both @azaroth42 and I find the size of the "TM" on the front page excessive... here is a fix


azaroth42:
:+1:

PR #1085, opened 2017-02-27
added Penn State to IIIFC

srabun:
I have added Penn State to the IIIF-C list, see: http://iiifc_pennstate.iiif.io/community/consortium/


mikeapp:
👍

Issue #1098, opened 2017-03-14
Add thumbnail declarations to the canvas and image resource examples in presentation API specifications


aeschylus:
It is very beneficial to have thumbnails pre-generated at the canvas level, but it is currently in the "MAY" category. This makes sense, but we might encourage more implementers to provide thumbnails for these nodes if they are featured in the canonical documentation. I'm thinking of the statements here: http://iiif.io/api/presentation/2.1/#thumbnail being given corresponding examples here: http://iiif.io/api/presentation/2.1/#canvas, http://iiif.io/api/presentation/2.1/#image-resources

zimeon:
Examples are certainly a good way to provide a gentle steer. It seems we currently (2.1) have `thumbnail` in just one example: http://iiif.io/api/presentation/2.1/#manifest . Of course the trade-off is size/complexity of examples. In this regard, the [canvas example](http://iiif.io/api/presentation/2.1/#canvas) is currently just a skeleton so it would seem unreasonable to add a whole Image API service object as a thumbnail, would it be helpful or counter-productive to add a simple image URI as thumbnail?

For the [image resources example](http://iiif.io/api/presentation/2.1/#image-resources), wouldn't the `thumbnail` simply be the same Image API service as the image being painted? And thus redundant?

tomcrane:
@zimeon @aeschylus I think it would be helpful to add a halfway house:

```
"thumbnail" : {
   "@id" : "http://example.org/thumbs/mythumb.jpg",
   "@type": "dctypes:Image",
   "width": 150,
   "height: 250
}
```

..from recent work in this area I think thumbnails that don't tell you how big they are aren't much use to real clients. 

On the other hand, my ideal thumbnail looks something like this:

```
"thumbnail": {
  "@id": "https://dlcs.io/thumbs/wellcome/1/8ac70935-b55b-4c6f-89ca-33e6ab9bb161/full/100,62/0/default.jpg",
  "@type": "dctypes:Image",
  "width": 100,
  "height: 62
  "service": {
      "@context": "http://iiif.io/api/image/2/context.json",
      "@id": "https://dlcs.io/thumbs/wellcome/1/8ac70935-b55b-4c6f-89ca-33e6ab9bb161",
      "protocol": "http://iiif.io/api/image",
      "width": 1024,
      "height": 631,
      "sizes": [
         { "width": 100, "height": 62 },
         { "width": 200, "height": 123  },
         { "width": 400, "height": 246  },
[...]

PR #1103, opened 2017-03-16
add thumb example as #1098

tomcrane:
pour encourager les autres...


jpstroop:
👍

azaroth42:
:+1: Non-normative merging on 4.

PR #1112, opened 2017-03-29
added sponsors so far to vatican event page

srabun:
added first three sponsors to the event page - sponsorship benefits begin once payment has been received.

See: http://adding_sponsors.iiif.io/event/2017/vatican/


azaroth42:
LGTM. Merging website only change.

Issue #1104, opened 2017-03-16
Wrong URI for rightsstatements (vocab, not page)


azaroth42:
@anarchivist:
just saw https://github.com/IIIF/iiif.io/commit/5ea386828ef44a7428ef8c4826899704ed0a4ac6 get merged - the rightsstatements.org URI should be `http://rightsstatements.org/vocab/InC-EDU/1.0/`, not `http://rightsstatements.org/page/InC-EDU/1.0/`

aisaac:
Just a warning: the same problem (use of http://rightsstatements.org/page/NoC-NC/1.0/) is also still in the Presentation API http://prezi_211.iiif.io/api/presentation/2.1/#manifest

Issue #1094, opened 2017-03-08
Use rightsstatements.org for license/rights examples in Image API


zimeon:
Clone of #960 to separate Image API case from Presentation API  case. We agreed that we should use examples from rightsstatements.org as these are likely good statements for our community. In Image API 2.1 we have two examples that include `license` (to be renamed `rights` in v3 #644):

  * http://iiif.io/api/image/2.1/#rights-and-licensing-properties -- example currently has `"http://example.org/rights/license1.html"`
  * http://iiif.io/api/image/2.1/#complete-response -- example currently has `[ "http://example.org/rights/license1.html", "https://creativecommons.org/licenses/by/4.0/" ]`

Tasks:
 * [x] fixed on 2.1.1 branch
 * [ ] merged to master

zimeon:
I propose that we change the first example to use just http://rightsstatements.org/page/InC-EDU/1.0/ , and the second to have http://example.org/rights/license1.html & http://rightsstatements.org/page/InC-EDU/1.0/

azaroth42:
:+1: to proposed solution

azaroth42:
Ready, modulo #1104, fixed in #1105.

aisaac:
Hi, it's great what you've done here!
Just wondering: would it help the reader if more context is given, about where such URIs come from? I was thinking of updating the table at the top of 5.4, replacing 
"A link to an external resource that describes the license or rights statement under which content obtained from the Image API service may be used."
by 
"A link to an external resource that describes the license or rights statement under which content obtained from the Image API service may be used. For example, the standardized licenses and rights statements available at https://creativecommons.org/licenses/ and http://rightsstatements.org/en/statements/"
@anarchivist do you think it would be appropriate on our side?

anarchivist:
As noted on the 12 April 2017 call:

* I'm -1 on **`SHOULD`** for using CC or RS.org statements specifically
* I'm +1 on **`SHOULD`** for recommending that the rights statement identified by URI come from some sort of controlled source of vocabulary

aisaac:
Continuing on noting the discussion on 12 April 2017 call: there is reluctance (especially from @zimeon ) on adding the example in the normative part (table at the top of section 5.4). 
I can very much understand this. I suggested an addition to the table as it was the first, natural place that occurred to me. But I will be equally happy if the sentence giving context for the URIs is added in the body of the text between the table and the example.

aisaac:
[...]

PR #1114, opened 2017-03-29
Fix page-->vocab for rs.o links in image

azaroth42:
Fixes rightsstatements.org links to use vocab not page URIs.


azaroth42:
c.f. #1094 and #1104

azaroth42:
Duplicate of 1105. Duh.

PR #1119, opened 2017-03-30
Update to latest versions of all the things

jpstroop:


jpstroop:
OK; that wasn't tedious at all! I think this is ready for review.

.travis.yml
```diff
@@ -13,7 +13,8 @@ script: rake ci
 # environment
 language: ruby
 rvm:
-  - "2.2"
+  - "2.2.3"
```
jpstroop: @eefahy this is important--Jekyll has a dependency on `ActiveSomethingOrOther` that only works with Ruby >2.2.3. I think that's what's on the server, but FYI.

zimeon:
Are we ready to accept breaking every section link that is out there through the new-style prefix of section numbers? Is there really no way to avoid this?

.travis.yml
eefahy: Yep that's on the server:
``` console
▶ ssh iiif          
Last login: Wed Mar 29 20:49:07 2017 from c-67-188-100-140.hsd1.ca.comcast.net
[iiif@ip-172-31-44-83 ~]$ rvm list rubies

rvm rubies

=* ruby-2.2.3 [ x86_64 ]

# => - current
# =* - current && default
#  * - default
```

jpstroop:
Closing. Will send a new PR that pins Jekyll

jpstroop:
See https://github.com/gettalong/kramdown/issues/424

PR #1127, opened 2017-04-05
Update Jekyll

jpstroop:
Replacement for #1119 ; Pins Kramdown to a version that doesn't break anchors

Could others please pull this down, install, and review?


zimeon:
:+1: (or some larger number). Working fine for me, no messed up section anchors (after some bit of hell with ssl certs on rubygems...)

jpstroop:
See https://github.com/gettalong/kramdown/issues/424#issuecomment-291959020

Just need to figure out how to pass those args to Jekyll

zimeon:
I think the key problem is that we want GFM handling but in kramdown 13.x that implies adding section numbers to the anchors, as shown in https://github.com/gettalong/kramdown/issues/424#issuecomment-291959020, it doesn't matter about the other flags:

```
> kramdown --version
1.13.2
> echo "# 5.5 Foo Bar" | kramdown -i GFM
<h1 id="55-foo-bar">5.5 Foo Bar</h1>
> echo "# 5.5 Foo Bar" | kramdown -i GFM --auto-id-stripping
<h1 id="55-foo-bar">5.5 Foo Bar</h1>
> echo "# 5.5 Foo Bar" | kramdown -i GFM --no-auto-id-stripping
<h1 id="55-foo-bar">5.5 Foo Bar</h1>
```

jpstroop:
@IIIF/editors + @eefahy I think this is ready.

zimeon:
:+1:

azaroth42:
+0 ... can't test as don't have the right ruby versions :(

azaroth42:
:+1: from discussion.

mikeapp:
👍

PR #1132, opened 2017-04-10
Fix fixtures, script, context, frame, add tests

azaroth42:
This adds checks for the manifest frame to the extent possible, based on the existing fixtures.
Also fixes the fixtures.

### DIFF
--- a/scripts/build_prezi21_tests.py
+++ b/scripts/build_prezi21_tests.py
@@ -1,0 +1,366 @@
+import os, sys
+from iiif_prezi.factory import ManifestFactory
+try:
+	from collections import OrderedDict	
+except:
+	try:
+		from ordereddict import OrderedDict
+	except:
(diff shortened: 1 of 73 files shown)


azaroth42:
Ping?

zimeon:
The changes look sensible. I note that when I run `build_prezi21_tests.py` manifests 6 and 64 are changed from the github version and don't pass `scripts/check_jsonld.py` -- is that intentional?

azaroth42:
@zimeon Do you have the most recent iiif_prezi installed with the bug fixes that make 6 and 64 possible to build?

zimeon:
@azaroth42 - I confirm that with the latest `iiif-prezi` I also get no changes to the fixture manifests on running `build_prezi21_tests.py` (except for 2 blank lines removed in 64), and that `scripts/check_jsonld.py` is happy.

PR #1136, opened 2017-04-12
Vatican promo

srabun:
News item: http://vatican_promo.iiif.io/news/2017/04/12/vatican-conference/

All sponsors added to event pages

Homepage news link updated


azaroth42:
Travis complaining about the markdown?

srabun:
@azaroth42 looks like it was an issue with the tag on line 5. Should be fixed now..

PR #1141, opened 2017-04-14
added to event materials that the events will be conducted in English

srabun:
Specifying that all events in the Vatican will be conducted in English.

http://events_english.iiif.io/news/2017/04/12/vatican-conference/

http://events_english.iiif.io/event/2017/vatican/


jpstroop:
👍

mikeapp:
👍

PR #1148, opened 2017-05-02
letter to dams

srabun:
letter to DAMS: http://dams_letter.iiif.io/news/2017/05/01/letter-to-dams/

### DIFF
--- a/source/_posts/2017-05-01-letter-to-dams.md
+++ b/source/_posts/2017-05-01-letter-to-dams.md
@@ -30,38 +30,38 @@ Thank you for your consideration.
 
 Sincerely,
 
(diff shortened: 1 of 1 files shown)


azaroth42:
On the museum call today there was discussion about the ordering of the names.  Currently they're simply chronological with us first, and the most recent signer last.  That's not ideal -- could we alphabetize them, and add the location of each museum more consistently after the name?

(@mikeapp)

azaroth42:
:+1:

srabun:
thanks @azaroth42 ! @mikeapp

mikeapp:
👍

Issue #1170, opened 2017-05-30
Typo in Presentation API 2.1 for Image profile URI


azaroth42:
Spotted by @regisrob, in two examples the URI for the Image API profile level is `http://iiif.io/api/image/2/profiles/level2.json` rather than what it should be: `"http://iiif.io/api/image/2/level2.json`

Thanks Regis!

regisrob:
:wink:

zimeon:
Thanks @regisrob !

PR #1171, opened 2017-05-30
fix profile URIs

azaroth42:
NOTE WELL: This change is very late breaking, but completely trivial.


zimeon:
:+1: well spotted @azaroth42 ! 

(One might argue that both or neither example should use `{{ site.image_api.latest.major }}` but I suspect the mix is bigger than just this issue...)

azaroth42:
Closes #1170

mikeapp:
👍

Issue #417, opened 2014-10-15
Visiting the iiif.io site over https causes the site not to work


aeschylus:
I get "this webpage is not available" in Chrome.

azaroth42:
Propose to use a Lets Encrypt cert and have both http and https serve the same content directly (rather than redirecting).
See: https://letsencrypt.org/

jpstroop:
Is this approach to help w/ backward compatibility? Why not redirect?

Should we create an issue for the 3.x.x APIs to change URIs for the context, etc. to https?

azaroth42:
Redirecting out of https to http seems not ideal. But our URIs are all HTTP, for contexts and other identifiers. And lots of redirects will just slow down the site with more requests.

Hopefully http2 will be firmly in place by the time we get to 3.x and the http/https distinction will be irrelevant (everything will be +s)

zimeon:
I think for now we should make both http and https work for the whole website, no redirects.

At some stage in the future (perhaps 3.x) we should make all URIs https. My sense from reading <https://httpwg.github.io/specs/rfc7540.html#discover-http> is that this is good even if http/2 has caught on, but perhaps the solution for dealing with http will be upgrade and not redirect? Anyway for now we kick that down the line a bit

jpstroop:
> At some stage in the future (perhaps 3.x) we should make all URIs https.

#601. 

Since breaking changes for auth and search are still allowed, should we make those HTTPS now?

zimeon:
I think it would be good to make all new URIs https... provided they work

azaroth42:
I think it would be confusing to have some http and some https ... people would get them round the wrong way. Better to be consistent when there's no technical requirement to be divergent, IMO.

zimeon:
This is still the case, https://iiif.io/ does not work. Current best practice is to make all websites natively https, with http open just to redirect.

azaroth42:
⬆️

zimeon:
Moved to https://github.com/IIIF/iiif-root-website/issues/11 . Closing

PR #1178, opened 2017-06-18
use https version of Google Fonts

mejackreed:


azaroth42:
Is there an issue for this? Both in terms of github issue, and technical problem that this solves?

mejackreed:
There is no GitHub issue for this.

In reference to the technical problem that it solves, it seems that as a website using external resources loaded in a user's browser (iiif.io), using the HTTPS protocol version of that resource is a responsible future proof choice.

If this is problematic, feel free to close.

zimeon:
:+1: Will be required for #417 (which hopefully will happen sooner rather than later). I don't see a reason to delay going ahead with this but will wait for another opinion

zimeon:
Moved to https://github.com/IIIF/website/issues/35

Issue #1182, opened 2017-06-25
switch all calls to zoom


srabun:
IIIF Call connection info beginning July 1:
·         IIIF Bi-weekly Community Call - https://stanford.zoom.us/j/356715389
o   Meeting ID: 356 715 389
·         A/V https://stanford.zoom.us/j/538552143
o   Meeting ID: 538-552-143
·         Discovery: https://stanford.zoom.us/j/829945399
o   Meeting ID: 829-945-399
·         Museums https://stanford.zoom.us/j/633505839
o   Meeting ID: 633-505-839
·         Manuscripts: https://stanford.zoom.us/j/674108679
o   Meeting ID: 674-108-679
·         Newspapers: https://stanford.zoom.us/j/948530998
o   Meeting ID: 948-530-998
·         Software devs: https://stanford.zoom.us/j/641742844
o   Meeting ID: 641-742-844
·         Mirador call: https://stanford.zoom.us/j/168558439
o   Meeting ID: 168-558-439
·         Mirador Developers Standup - https://stanford.zoom.us/j/340409741
o   Meeting ID: 340 409 741
·         Text Granularity – TBD – please complete the doodle poll for scheduling: http://doodle.com/poll/e49ru9sk4hwanavy

srabun:
Branch: changing_to_zoom

zimeon:
Fixed by #1184

PR #1184, opened 2017-06-27
updated events page and added page for Toronto

srabun:
see: http://update_events.iiif.io/event/

and: http://update_events.iiif.io/event/2017/toronto/


azaroth42:
Inner heading still says The Hague :)

zimeon:
Do you want to push it out with "Goes here" in the TOC?

jpstroop:
That should get replaced by a macro when the Markdown is converted, unless the Markdown is broken somewhere (on my phone, can't really see).

srabun:
I just didn't change the first heading. It should be fixed now. I do blame jetlag in Australia.

zimeon:
Changes look good and also include bluejeans->zoom changes, fixes #1182

PR #1189, opened 2017-07-10
added text granularity group info

srabun:
added text group info to:

http://text_group.iiif.io/community/groups/
http://text_group.iiif.io/community/groups/text-granularity/
http://text_group.iiif.io/community/groups/text-granularity/charter/


mikeapp:
👍

PR #1207, opened 2017-07-30
Fix some minor typos

kinow:
Hi,

Had to read the presentation and search API's for a presentation and noticed a few typos here and there.

Not sure what is the policy for updating old/current specs. I can update the pull request in case only a subset of the changes are useful - if any at all :-)

Bruno

### DIFF
--- a/source/api/annex/index.html
+++ b/source/api/annex/index.html
@@ -1,7 +1,7 @@
 ---
 title: IIIF API Annex Documents
 id: index
-categores: [pages]
+categories: [pages]
 layout: sub-page
 tags: [annex, rest, image-api]
 ---
--- a/source/api/annex/notes/index.html
+++ b/source/api/annex/notes/index.html
@@ -1,7 +1,7 @@
 ---
 title: IIIF API Implementation Notes
 id: index
-categores: [pages]
+categories: [pages]
 layout: sub-page
 tags: [annex, rest, image-api]
 ---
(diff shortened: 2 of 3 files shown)


zimeon:
Thanks @kinow . It is our usual policy to fix typos only in the unversioned documents and current versions (as patch release or next version), and to leave old versions archived. We also usually update one spec at a time so probably the best way forward will be for us to cherry-pick your commits -- I'll leave this PR open until we do that.

kinow:
Thanks @zimeon . If anyone points which docs must be updated, I'd be happy to update the commits, rebase, and push-force.

azaroth42:
@kinow Thank you so much for the very close reading and noting all the typos (mostly mine!)

The Annex documents and the Shared Canvas model document can be updated at will, but the api/SPEC/x.y/ documents should remain static.  I think the best course of action, is to take the api/ docs out of the PR and then accept it.

Then create an issue, tagged as milestone 2.2, with the typos to fix. 

Thoughts?

kinow:
Sounds like a plan @azaroth42 ! Will update the pull request and, if I understand correctly, create an issue here in GitHub, letting someone else to tag it and link to this PR... or should I just update the PR and someone else will create the issue?

kinow:
PR updated!

azaroth42:
:+1: from me!

azaroth42:
Merging! Thanks @kinow!

PR #1208, opened 2017-08-01
updated code of conduct

srabun:
Publishing new Code of Conduct: http://code_of_conduct.iiif.io/event/conduct/

Guidelines for conduct team members: http://code_of_conduct.iiif.io/event/conduct-guidelines/

Previous version: http://code_of_conduct.iiif.io/event/conduct-09/

Does this work?


zimeon:
I don't think we should invent a new URI version scheme for this document. I think we should keep just the current version live and update a changelog (same strategy as for documents like http://iiif.io/api/annex/notes/editors/#status-of-this-document). We can rely on github for the history if need be. So in this proposal we'd remove `conduct-09.md`, add dates and changelog to `conduct.md` and `conduct-guidelines.md`.

I feel we discussed this at some stage but can't find it, apologies if we did and I'm misremembering.

jpstroop:
I agree.

azaroth42:
This is probably from my comment on the google doc that the actual harassment process (the second part of the document) should be semantically versioned to prevent silent or un-announced changes, or otherwise "I didn't know that there had been a change to the rules!" excuses. Whether it's always in-place (everything is a patch) or there are major/minor versions wasn't discussed, so I commend @srabun in taking the initiative here.

How about the first part is left as is, but then the anti-harassment and reporting gets a version number as per the APIs as a separate document?  I also think this belongs under community not under event, as it explicitly applies to all interactions, online or face to face, not just at specific time-based events.

How about:
  
* iiif.io/community/conduct  (fluffier intro part)
* iiif.io/community/conduct/0.9/anti-harassment
* iiif.io/community/conduct/1.0/anti-harassment

srabun:
The community norms and expectations section is also subject to change, so could we keep it simple and just have a change log for the entire doc rather than splitting off the harassment section? I tend to agree with Simeon's suggestion to just keep a change log with dates for the conduct doc and the guidelines doc.

azaroth42:
That works for me :) Just wanted to float the split as an idea.

PR #1212, opened 2017-08-02
added folger to iiif-c list

srabun:
Folger just joined the IIIF-C - I can't deploy the branch bc "no space left on device"...


azaroth42:
If only @eefahy were allowed to fix this.

azaroth42:
:+1:

PR #1221, opened 2017-08-21
added v&a event

srabun:
Not sure what happened with my branch name. in any case: 

http://v.iiif.io/event/
http://v.iiif.io/event/2017/london/


jpstroop:
Re your branch name, an `&` might have messed it up if you tried to do that.

jpstroop:
:+1:

mikeapp:
👍

PR #1223, opened 2017-08-23
added national gallery of art to consortium list

srabun:
http://adding-nga.iiif.io/community/consortium/

NGA is joining IIIF-C


jpstroop:
:+1:

PR #1281, opened 2017-10-05
Make http://prezi3.iiif.io/ work

glenrobson:
Travis changes to make a branch that has a sub domain of IIIF link to the root rather than the branch name. Once this is merged it should make the css and links work for:

http://prezi3.iiif.io/api/presentation/3.0/

Once this is merged Ill look at putting a forward for:

http://preview.iiif.io/api/prezi3/api/presentation/3.0/

so it forwards to the above as once this is merged the css and links won't work for preview.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 sudo: false
-
+env:
+  global:
+    - ROOT_INSTALL="true"
 before_install:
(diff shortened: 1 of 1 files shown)


glenrobson:
This is what the redirect will look like once its in place:

http://preview.iiif.io/api/prezi3-root/api/presentation/3.0/#objectives-and-scope

PR #1298, opened 2017-11-10
Language & choicehint

### LINKED ISSUE #755: Simplify language assignment?
Although it's not clear from the specification, it turns out that language containers are handled correctly when they're just strings with no language associated:

```
{
  "@context":
  {"occupation": {"@id": "ex:occupation", "@container": "@language"}},
  "occupation": "Nindža"
}
```

Works at the same time as:

```
{
  "@context":
  {"occupation": {"@id": "ex:occupation", "@container": "@language"}},
  "occupation": {"en": "ninja"}
}
```

However multiple are not. This fails:

```
{
  "@context":
  {"occupation": {"@id": "ex:occupation", "@container": "@language"}},
  "occupation": [{"en": "ninja"}, "ninjas"]
}
```

And has to be, instead:

```
{
  "@context":
  {"occupation": {"@id": "ex:occupation", "@container": "@language"}},
  "occupation": {"en": "ninja", "": "ninjas"}
}
```

This seems like an improvement over the lists of @language/@value pairs?

### LINKED ISSUE #1194: Allow choiceHint
From the AV work, we know that we need to hint to the client whether it (the software) should make the choice or whether the user should make the choice.  Something like the description below.

The question is whether the list of values is closed or whether there can be any URI as per `viewingHint`?  My preference would be to allow additional hints via full URIs to enable experimentation without getting in the way of validation.

---

A hint associated with a Choice resource that a client can use to determine the publisher's intent as to which agent _SHOULD_ make the choice between the different options.  In the absence of any `choiceHint` value, the rendering application can use any algorithm or process to make the determination.

* A Choice _MAY_ have exactly one `choiceHint`, and the value _MUST_ be one of the options described below.

| Value | Description |
| ----- | ----------- |
| `client` | The client software is expected to select an appropriate option without user interaction. |
| `user` | The client software is expected to present an interface to allow the user to explicitly select an option. |

``` json-doc
{"choiceHint": "client"}
```

azaroth42:
Closes #755 ; Closes #1194

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -488,24 +488,6 @@ The value _MUST_ be an array of strings, taken from the table below or a full UR
 {"renderingHint": ["auto-advance", "individuals"]}
 ```
 
-##### choiceHint
-
-A hint associated with a Choice resource that a client can use to determine the publisher's intent as to which agent _SHOULD_ make the choice between the different options.  In the absence of any `choiceHint` value, the rendering application can use any algorithm or process to make the determination.  This specification defines the two values specified in the table below. Others may be defined externally, and would be given as a full URI.
-
-The value _MUST_ be a string, taken from the table below or a full URI.
-
-* A Choice _MAY_ have exactly one `choiceHint`.<br/>
-  Clients _SHOULD_ process `choiceHint` on a Choice.
-
-> | Value | Description |
-| ----- | ----------- |
-| `client` | The client software is expected to select an appropriate option without user interaction. |
-| `user` | The client software is expected to present an interface to allow the user to explicitly select an option. |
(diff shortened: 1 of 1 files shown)


azaroth42:
Preview: http://preview.iiif.io/api/language-choicehint/api/presentation/3.0/ 

Note @glenrobson that the CSS isn't working?

glenrobson:
Hi Rob,

If you set 

```- ROOT_INSTALL="false"```

to false rather than true in:

https://github.com/IIIF/api/blob/language-choicehint/.travis.yml

the css will start working again. Set ROOT_INSTALL to true if I've setup a subdomain for it like ```http://prezi3.iiif.io/```

Cheers

Glen

PR #1302, opened 2017-11-13
Add no conneg rule for seeAlso

### LINKED ISSUE #942: Should `format` be used to support conneg on `seeAlso` resources?
Currently [prezi spec for `seeAlso`](http://iiif.io/api/presentation/2.1/#seealso) says `format` and `profile` are just to understand result from following `seeAlso`. NLW/Europeana use of `seeAlso` requires conneg on the URI (i.e. add `Accept` header with value from `format`), and this must alse be repeated on following the 302 redirect. Should any of this be added in spec?

(Issue came up in demo of NLW/Europeana newspapers work on 2016-09-28 call https://docs.google.com/document/d/1Occ45PtAjlTSCFiV7nmJL_lryFTveD5rvKWIbaaqU0k/edit#heading=h.vucbcocsan1l )

azaroth42:
Closes #942 ; add guidance on negotiable `seeAlso` resources to `seeAlso` description.


source/api/presentation/3.0/index.md
```diff
@@ -579,7 +579,7 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id` a
...
+A link to a machine readable document that is related to the resource with the `seeAlso` property, such as an XML or RDF description. Properties of the document should be given to help the client select between multiple descriptions (if provided), and to make appropriate use of the document. If the relationship between the resource and the document needs to be more specific, then the document should include that relationship rather than the IIIF resource. Other IIIF resources, such as a related Manifest, are valid targets for `seeAlso`. The URI of the document _MUST_ identify a single representation with a single `format`. For example, if the same data exists in JSON and XML, with a URL that supports content negotiation for the two formats, then the URIs of each representation can be provided separately.
```
jpstroop: > ... JSON and XML, with a URL that ...

URL or URI?

jpstroop: > then the URIs of each representation can be provided separately. 

Also, not sure this makes sense to me. We're saying that there should be a JSON object per `format`, even if some those formats have the same URI and require conneg to get the desired representation/serialization, is that correct?

jpstroop: @zimeon will suggest a rewrite of last two sentences.

azaroth42: Urgh. Technically they're locators, but we should stick with just URI.

The URI to use is the one for the representation, not the negotiable resource. It's hard to express in generic terms without requiring a deep knowledge of HTTP conneg details :(

zimeon:
Should also fix `@id` -> `id` in the example for `seeAlso`

source/api/presentation/3.0/index.md
zimeon: I suggest replacing the last sentence with:

> For example, if the same data exists in JSON and XML, then separate JSON objects should be added for each representation, with distinct `id` and `format` properties. 

if we want to be explicit about conneg then we could also add:

> Clients are not expected to understand how to content negotiate based on the `format` property of `seeAlso` resources.

but I'm include to go with just the "For example..." sentence.

source/api/presentation/3.0/index.md
```diff
@@ -579,7 +579,7 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id` a
...
+A link to a machine readable document that is related to the resource with the `seeAlso` property, such as an XML or RDF description. Properties of the document should be given to help the client select between multiple descriptions (if provided), and to make appropriate use of the document. If the relationship between the resource and the document needs to be more specific, then the document should include that relationship rather than the IIIF resource. Other IIIF resources, such as a related Manifest, are valid targets for `seeAlso`. The URI of the document _MUST_ identify a single representation with a single `format`. For example, if the same data exists in JSON and XML, then separate JSON objects should be added for each representation, with distinct `id` and `format` properties.
```
jpstroop: One more quibble:
> The URI of the document _MUST_ identify a single representation with a single `format`.

I'm not sure that "format" should be in backticks/monospace here.

azaroth42: Agreed. Given a representation can only ever have a single format by definition, it's redundant anyway.

azaroth42:
:+1: for @jpstroop's tweak

jpstroop:
I think it's up to @tomcrane or @mikeapp to pull the trigger.

mikeapp:
👍

tomcrane:
belatedly :+1:

PR #1303, opened 2017-11-13
no interstitial, recommend auth

### LINKED ISSUE #1155: Rendering cannot refer to an interstitial page
Can the `rendering` property refer to an interstitial page, or must it refer directly to the downloadable resource?

For example, we would like to replicate the experience of the download menu here:
    http://www.getty.edu/art/collection/objects/826/vincent-van-gogh-irises-dutch-1889/

When you select one of the download options, it takes you to a page where you fill in how you're going to use the image, and it collects different information for different uses.  Clearly that menu can be replicated in `rendering`, but the expectation I imagine is that the `id` of the resource will be the thing to download, rather than a second step form to fill.

azaroth42:
Closes #1155 ; disallow interstitial resources such as splash pages, and recommend Auth for gating access.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -547,7 +547,7 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id`, 
 ```
 
 ##### rendering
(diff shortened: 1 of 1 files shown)


jpstroop:
:+1:

tomcrane:
:+1:

Issue #1072, opened 2017-02-15
To Id Map or Not Id Map?


jpstroop:
Given the likely introduction of nested canvases by the A/V work, should we consider [JSON-LD Id Maps](http://json-ld.org/spec/latest/json-ld/#id-maps)? The also has the potential to make it possible to collapse Sequences and Ranges into one thing.

azaroth42:
Propose close wontfix

jpstroop:
In the interest of developer happiness I think we should keep this open, unless there's a clear reason why it wouldn't work.

azaroth42:
The approach wasn't popular for ranges, c.f. @glenrobson's comment about not being able to see which ranges were inside which other ranges. Happy to leave open until we can get some feedback though.

jpstroop:
I get that, but I don't completely buy the use case that manifests need to be human readable.

azaroth42:
@jpstroop If you still care about this issue, could you prepare a summary for Toronto?

jpstroop:
Now that ranges are more readable, is this any more compelling?

zimeon:
Example of use in JSON-LD docs, see `post` in https://json-ld.org/spec/latest/json-ld/#node-identifier-indexing

azaroth42:
Getty AV Discussion, from memory:
Pro: Easy to find references to canvases, if there's more than one reference to it. E.g. from ranges,  from other canvases embedding them, or startCanvas.
Pro: Prevents clients from needing to construct the map by iteration over the Sequence
Con: Significant upheaval to the manifest structure
Con: Harder to read the document
Con: Only very small percentage of manifests would benefit (e.g. with Ranges)
Con: Makes template based construction harder

Diversion: Not all references in ranges are exactly canvas URIs, they could have fragments. Processing for this would still need to process the fragment.

jpstroop:
Don't think I'm not watching :eyes:

azaroth42:
Why aren't you carrying boxes around? Slacker!

jpstroop:
🏠 is empty now!

jpstroop:
> Pro: Easy to find references to canvases, if there's more than one reference to it. E.g. from ranges, from other canvases embedding them, or startCanvas.

Yes. Especially if Sequences go away as well.

> Pro: Prevents clients from needing to construct the map by iteration over the Sequence

Which they _all_ do.

> Con: Significant upheaval to the manifest structure

If not now when, especially since it simplifies things?

[...]

PR #1304, opened 2017-11-13
Add client reqs for default nav structure, improve description

### LINKED ISSUE #1255: Clients SHOULD present first range structure
If the publisher has a multiple top level structure hierarchies (e.g. logical versus physical structures), and has a preference for which the viewer should display by default, then the first hierarchy SHOULD be displayed by the client as a default.

Further evidence towards #1193 plus an additional client requirement on top of that.

azaroth42:
Closes #1255 ;

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -767,12 +767,12 @@ The value _MUST_ be an array of objects.
 
 ##### structures
 
-The structure of an object represented as a Manifest can be described using a hierarchy of Ranges.  The top level Ranges of these hierarchies are given in the `structures` property.
+The navigational structure of an object represented as a Manifest can be described using a hierarchy of Ranges. This can be used to describe the "table of contents" of the object or other navigational structures that the user can interact with beyond a simple linear progression described in the Sequence.  The hierarchy is built by nesting the child Range resources in the `items` array of the higher level Range.  The top level Ranges of these hierarchies are given in the `structures` property. 
 
 The value _MUST_ be an array of objects.
 
 * A Manifest _MAY_ have a list of one or more Ranges in `structures`.<br/>
-  Clients _SHOULD_ process `structures` on a Manifest.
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -767,12 +767,12 @@ The value _MUST_ be an array of objects.
 
 ##### structures
 
-The structure of an object represented as a Manifest can be described using a hierarchy of Ranges.  The top level Ranges of these hierarchies are given in the `structures` property.
+The navigational structure of an object represented as a Manifest can be described using a hierarchy of Ranges. This can be used to describe the "table of contents" of the object or other navigational structures that the user can interact with beyond a simple linear progression described in the Sequence.  The hierarchy is built by nesting the child Range resources in the `items` array of the higher level Range.  The top level Ranges of these hierarchies are given in the `structures` property.
```
jpstroop: I think we should strike the word "navigational", e.g. I have structural metadata that represents the physical structure of a printed book (gatherings, leaves, and sides). I'm not sure that's strictly navigational, or just a structure that can be represented. Also, paging through the sequence is just as much navigation as anything one might do with ranges, so I think the word just fuels confusion.

azaroth42: The main purpose of Ranges is navigation. Agree there are other potential uses of the structure, so happy to move the sentence around a bit, but I think we should keep the notion of navigation as the primary usage

jpstroop: I could turn this into an argument for #1072 but will hold off for now. I think the distinction becomes ever-more blurry w/ A/V

zimeon: Fine to merge after striking the word 'navigational'

azaroth42: :+1: to striking first "navigational". The following sentence describes the expected usage.

mikeapp:
👍

azaroth42:
Merge?

PR #1305, opened 2017-11-14
add URI recommendations section

### LINKED ISSUE #1214: Remove recommended URI patterns
As they're not very valuable and take up time to understand and subsequently ignore.

azaroth42:
Closes #1214 ; add section with recommendations on URI best practices

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1497,7 +1497,7 @@ While any URI is technically acceptable for any of the resources, there are seve
 * Use the type of the resource in the UR, such as `https://data.example.org/iiif/849/manifest/`
 * Don't use `.json` or `.jsonld` on the end of the URI unless you need to
 * Remember to encode special characters
-* Consider future expansion of the data, and related services, such as `https://data.example.org/iiif/849/manifest/search/` for a [IIIF search service][search] over the object's annotations
+* Consider future expansion of the data, and related services, such as `https://data.example.org/iiif/849/manifest/search/` for a [IIIF search service][search-api] over the object's annotations
 * Do not use query parameters or fragments in the URIs
 * Once published, they should be as persistent and unchanging as possible
 
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1452,14 +1452,22 @@ And the corresponding first page of manifests:
 
 This section describes the _RECOMMENDED_ request and response interactions for the API. The REST and simple HATEOAS approach is followed where an interaction will retrieve a description of the resource, and additional calls may be made by following links obtained from within the description. All of the requests use the HTTP GET method; creation and update of resources is not covered by this specification.
 
-###  6.1. Requests
+### 6.1 URI Recommendations
 
-Clients _MUST NOT_ construct resource URIs by themselves, instead they _MUST_ follow links from within retrieved descriptions.
+While any URI is technically acceptable for any of the resources in the API, there are several best practices for designing the URIs for the resources.
 
+* The URI _SHOULD_ use the HTTPS scheme, not HTTP.
+* The URI _SHOULD NOT_ include query parameters or fragments.
+* Once published, they _SHOULD_ be as persistent and unchanging as possible.
+* Special characters _MUST_ be encoded.
 
-###  6.2. Responses
+###  6.2. Requests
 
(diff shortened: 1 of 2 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -1456,32 +1456,26 @@ This section describes the _RECOMMENDED_ request and response interactions for t
...
-If this cannot be generated due to server configuration details, then the content-type _MUST_ instead be `application/json` (regular JSON).
+If this cannot be generated due to server configuration details, then the content-type _MUST_ instead be `application/json` (regular JSON), without a `profile` parameter.
 
 ``` none
 Content-Type: application/json
 ```
 {: .urltemplate}
 
-The HTTP server _MUST_ follow the CORS requirements, including the `Access-Control-Allow-Origin` and the value of the header _SHOULD_ be `*`.
+The HTTP server _MUST_ follow the CORS requirements, including an `Access-Control-Allow-Origin` header with the value of `*`.
```
tomcrane: Is a specific origin allowed, and list of specific origins?

source/api/presentation/3.0/index.md
```diff
@@ -1490,7 +1484,23 @@ Access-Control-Allow-Origin: *
...
+* Consider the hostname for your resources, such that it could redirect to separate systems if needed. For example: `https://data.example.org/`
+* Use a consistent prefix path for IIIF resources, such as `https://data.example.org/iiif/`
+* Use a natural identifier for the resource in its management system in the URI, such as `https://data.example.org/iiif/849/` for object number 849
+* Use the type of the resource in the URI, such as `https://data.example.org/iiif/849/manifest/`
+* Don't use `.json` or `.jsonld` on the end of the URI unless you need to due to implementation requirements
+* Remember to encode special characters
+* Consider future expansion of the data, and related services, such as `https://data.example.org/iiif/849/manifest/search/` for a [IIIF search service][search-api] over the object's annotations
+* Do not use query parameters or fragments as part of the URI
+* When dereferenced, the URI _SHOULD_ return the JSON-LD description of the resource
+* Once published, they _SHOULD_ be as persistent and unchanging as possible
```
zimeon: Discussion in Eds meeting is that the normative ones should probably stay here:

  * The URI _SHOULD_ use the HTTPS scheme, not HTTP.
  * Do not use query parameters or fragments as part of the URI --- make stronger _SHOULD NOT_ ?
  * When dereferenced, the URI _SHOULD_ return the JSON-LD description of the resource -- is this actually a _MUST_ in order for IIIF to work?
  * Once published, they _SHOULD_ be as persistent and unchanging as possible

And then link off to patterns in an implementation note. Within those, not sure that "Use a natural identifier for the resource in its management system in the URI, such as `https://data.example.org/iiif/849/` for object number 849" is good as this may actually get in the way of persistence if system is changed

azaroth42: That seems good. I was struggling to fulfill the requests in the issue and still keep it in the spec. The implementation note would be part of a larger document, I imagine?

zimeon: I imagine we'll find more to add, but I think starting a prezi3 implementation notes document with this would seem good for now

source/api/presentation/3.0/index.md
azaroth42: How about _MUST_ have the header, and value _SHOULD_ be * ?

azaroth42:
Fixes in place now?

source/api/presentation/3.0/index.md
```diff
@@ -1452,36 +1452,38 @@ And the corresponding first page of manifests:
...
+### 6.1 URI Recommendations
 
-Clients _MUST NOT_ construct resource URIs by themselves, instead they _MUST_ follow links from within retrieved descriptions.
-
-Implementation note doesn't belong in spec
-{: .warning}
-In the situation where the JSON documents are maintained in a filesystem with no access to the web server's configuration, then including ".json" on the end of the URI is suggested to ensure that an appropriate `content-type` response header is sent to the client.
+While any URI is technically acceptable for any of the resources in the API, there are several best practices for designing the URIs for the resources.
 
+* The URI _SHOULD_ use the HTTPS scheme, not HTTP.
```
jpstroop: Exception to https://github.com/IIIF/api/pull/1307#discussion_r152038810

PR #1306, opened 2017-11-14
disallow @graph

### LINKED ISSUE #920: Disallow @graph pattern
In the implementation notes for the Presentation API, we say that clients should process the `@graph` JSON-LD keyword if they want to be fully conforming.  Either this should be a requirement in the text of the specification, or it should just be disallowed.

My preference is to disallow its use completely and require servers to output the correct form.  `@graph` is required when the graph to be serialized is unconnected, however that will never occur in IIIF as we define the structure very carefully.  Any custom fields can be added via extensions, which we allow for.

The use case put forwards for `@graph` is the separation of metadata about the representation and the object, however that's unnecessary in IIIF -- we have metadata about the resource being depicted/described in `metadata`, so any use of (for example) dcterms:modified could be unambiguously interpreted as the modification timestamp for the manifest.

The cost of allowing `@graph` is not as high as some other choices, if its value is the full manifest structure.  However if you can generate the right structure, you can (as above) simply serve it from the URI directly.  If the server can't generate the right structure, I don't think we should consider it as an implementation of IIIF, and thus expect the client to use frames (and likely some hand crafting to get things in the right shape after that) to interpret the data.  I don't think we would expect the client to interpret a list of every node (including all the rdf:List blank nodes separately), but that is implicitly legal according to the implementation note.

azaroth42:
Closes #920 ; explicitly disallow the `@graph` pattern from JSON-LD compaction.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -880,6 +880,8 @@ The value of the `@context` property _MUST_ be a list, and the __last__ two valu
 
 Any additional fields beyond those defined in this specification or the Web Annotation Data Model _SHOULD_ be mapped to RDF predicates using further context documents.   If possible, these extensions _SHOULD_ be added to the top level `@context` field, and _MUST_ be added before the above contexts.  The JSON-LD 1.1 functionality of type and predicate specific context definitions _SHOULD_ be used if possible to try to minimize any cross-extension collisions.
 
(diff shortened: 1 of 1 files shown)


azaroth42:
:+1: to pretty engish

zimeon:
Ooops... need to undo attempt to fix with web interface, accidentally deleted old addition. Will fix from command line in a few mins

PR #1308, opened 2017-11-17
1065 language

### LINKED ISSUE #1065: Add language property for external resources
For example to say that an audio file referred to from an annotation is in English.

Should be available for any external content bearing resource (audio, video, ...) and AnnotationLists [as representation of subtitles, etc], and Layers [as cross-canvas grouping mechanism for AnnotationLists].

azaroth42:
Closes #1065 ; Add language property as technical metadata.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -364,11 +364,11 @@ Note that this is different to the `formats` property in the [Image API][image-a
 ```
 
 ##### language
-The language or languages used in the content of an external resource. This property is already available from the Web Annotation model for resources that are annotated onto Canvases, however it _MAY_ also be used for resources referenced with `related`, `rendering`, `rights` and similar properties.
+The language or languages used in the content of an external resource. This property is already available from the Web Annotation model for content resources that are the body or target of an Annotation, however it _MAY_ also be used for resources referenced from `related`, `rendering`, `rights`, and `within`.
 
 The value _MUST_ be an array of strings.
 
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -363,6 +363,20 @@ Note that this is different to the `formats` property in the [Image API][image-a
...
+##### language
+The language or languages used in the content of an external resource. This property is already available from the Web Annotation model for content resources that are the body or target of an Annotation, however it _MAY_ also be used for resources referenced from `related`, `rendering`, `rights`, and `within`.
+
+The value _MUST_ be an array of strings.
+
+ * An external resource _MAY_ have one or more `language` values, and each value _MUST_ be a valid language code, as described under the [languages section][languages-prezi30].<br/>
+   Clients _SHOULD_ process the `language` of external resources to present the most appropriate content to the user.
+ * Other resource types _MUST NOT_ have a `language`.<br/>
+   Clients _SHOULD_ ignore `language` on other resource types.
+
```
jpstroop: Why the `<br/>`s and not just separate bullets?

jpstroop: Why the `<br/>`s and not just separate bullets?

azaroth42: Because it's the client requirements for that property on that resource type. 

It's also consistent all the way through section 3 (so would be another issue if there's a better way to manage it)

jpstroop: OK. Fine with it as-is for now...I wonder if there's a better way to make that distinction.

PR #1309, opened 2017-11-17
1070 ranges

### LINKED ISSUE #1070: Ranges try (and fail) to capture both extent and hierarchy
The canvases / ranges separation allows for ranges to be used to capture navigation hierarchy, separate from the complete extent of the parent range.  2.1 allows the use of members to capture a single order across both canvases and ranges (#716). However, the v3 proposal to reduce down to just members loses the functionality of separation of extent and hierarchy.

For example, a video starts with 20 seconds of applause and audience noise. The start of Act 2 of the Opera then starts at 20 seconds and continues until 620 seconds.  At 60 seconds in, the first song within Act 2 starts for 120 seconds.  Then at 300 seconds in, the second song starts and runs until 360 seconds.  We do not want to create an artificial range for the time spans 0,20 or 180,300, or 360,620.

Extent of: 
 * Act 2 is 20,620
 * Act 2, song 1 is 60,180
 * Act 2, song 2 is 300,360

Goal is to display this hierarchy without additional ranges.

Opera is:
 * Act 1
 * Act 2

azaroth42:
Closes #1070 ; New definition of ranges including extent and hierarchy. Nested pattern, with no-nav and label requirements.


source/api/presentation/3.0/index.md
```diff
@@ -1179,17 +1179,17 @@ Additional features of the [Web Annotation][webanno] data model _MAY_ also be us
 
 ###  5.6. Range
 
-Ranges are used to describe additional structure within an object, such as newspaper articles that span pages or chapters within a book. Ranges can include Canvases, parts of Canvases, or other Ranges, creating a nested tree structure such as a table of contents.
+Ranges are used to describe additional structure within an object, such as newspaper sections or articles, or chapters within a book. Ranges can include Canvases, parts of Canvases, or other Ranges, creating a tree structure like a table of contents.
```
jpstroop: > "...additional structure within an object, ..."

change to "the structure of an object," ?

source/api/presentation/3.0/index.md
```diff
@@ -1179,17 +1179,17 @@ Additional features of the [Web Annotation][webanno] data model _MAY_ also be us
...
 
-All of the Canvases or parts that should be considered as being part of a Range _MUST_ be included within the Range's members, or a descendant Range's members.
+All of the Canvases or parts that should be considered as being part of a Range _MUST_ be included within the Range's `items` list, or a descendant Range's `items`. 
 
-The Canvases and parts of Canvases may or may not be contiguous or in the same order as any Sequence.  Examples include newspaper articles that are continued in different sections, or simply a chapter that starts half way through a page. Parts of Canvases _MUST_ be rectangular and are described using the `xywh=` fragment approach.
+The Canvases and parts of Canvases may or may not be contiguous or in the same order as in any Sequence.  Examples include newspaper articles that are continued in different sections, or simply a chapter that starts half way through a page. Parts of Canvases _MUST_ be rectangular and are described by appending an `xywh=` fragment to the end of the Canvas's URI.
```
jpstroop: > "... may or may not be ..."

"need not be" or "are not necessarily" or start the sentence with "It is not necessary for..."

tomcrane: Examples need to work for time-only canvas:

Examples include newspaper articles that are continued in different sections, a chapter that starts half way through a page, or time segments of a single canvas that represent different sections of a piece of music. Spatial parts of Canvases _MUST_ be rectangular and are described by appending an `xywh=` fragment to the end of the Canvas's URI. Temporal parts of Canvases are described by appending a `t=` fragment to the Canvas's URI. Temporal and spatial fragments can be combined to identify a particular region for a particular time duration.

(with an example)

source/api/presentation/3.0/index.md
```diff
@@ -1179,17 +1179,19 @@ Additional features of the [Web Annotation][webanno] data model _MAY_ also be us
...
 
-All of the Canvases or parts that should be considered as being part of a Range _MUST_ be included within the Range's members, or a descendant Range's members.
+All of the Canvases or parts that should be considered as being part of a Range _MUST_ be included within the Range's `items` list, or a descendant Range's `items`. 
 
-The Canvases and parts of Canvases may or may not be contiguous or in the same order as any Sequence.  Examples include newspaper articles that are continued in different sections, or simply a chapter that starts half way through a page. Parts of Canvases _MUST_ be rectangular and are described using the `xywh=` fragment approach.
+Parts of Canvases are still Canvases and have a `type` of "Canvas". Spatial parts of Canvases _MUST_ be rectangular and are described by appending an `xywh=` fragment to the end of the Canvas's URI. Temporal parts of Canvases _MUST_ be described by appending a `t=` fragment to the end of the Canvas's URI. Spatial and temporal fragments _MAY_ be combined, using an `&` character between them.
```
zimeon: Should we say that spatial and temporal _SHOULD_ appear in a particular order such as  `#xywh=...&t=...` or `#t=...&xywh=...` (and _NOT_ the other) in order to have a canonical URI? In https://www.w3.org/TR/media-frags/#fragment-dimensions it says that the dimensions are independent but the [client guidelines](https://www.w3.org/TR/media-frags/#media-fragment-clients) say that clients are expected to do temporal first, then spatial at the codec level. That would suggest temporal first syntax might be the best option if we specify an order: `#t=...&xywh=...`

mikeapp: If we are further constraining the fragments do we want to prohibit references to a non-existent dimension on the Canvas?  https://www.w3.org/TR/media-frags/#error-media-general says "If the user agent knows the mime type, it is able to detect non-existent dimensions and SHOULD ignore them."

azaroth42: Given #1320, I think the definition of parts of Canvases should actually go under the Canvas definition as it will be used in multiple places.  Then Ranges, Annotations, `start` and anywhere else (that might be it?) can reference the Canvas section.  Thoughts?

source/api/presentation/3.0/index.md
jpstroop: I still think we should strike "additional"

jpstroop: Oops, looking at wrong diff, NM!

azaroth42:
Okay to merge now?

PR #1311, opened 2017-11-17
Clarify single document for LOD round trip

### LINKED ISSUE #1144: Legal to summarize Lists?
In a Collection of Manifests, where the intended representation lets you drill down into parts of the collected Manifests, I'd like to have the important ranges for navigation in the collection document for ease of access, and a hint to the viewer I'd like them to be rendered.

However the AV group has agreed that all parts of the resource must be represented in the Ranges, if there are any.  Does that apply to Manifests as documents (e.g. I can do it), or Manifests as "things" that might be summarized in Collections (e.g. I can't do it)?  This will take effect in 3.0, but today the question still has implications in 2.x because ...

It becomes a bit more subtle than some of the other properties ... we can leave out `sequences` completely, or `rights` or whatever ... but `structures` is a `@type:@list`, meaning that if a system were treating everything as triples in a global pool, there would be two different Lists, one summary from the Collection and one complete from the Manifest. Following from our design principles:

> It should be possible to transform representations back and forth between triples and the JSON-LD serialization without loss, but not necessarily without the use of custom code.

This would _not_ be round-trippable once you take into account multiple documents.

### LINKED ISSUE #1267: Note "within one document" in "Follow Linked Data Principles" design principle
The [Follow Linked Data Principles](http://iiif.io/api/annex/notes/design_patterns/#follow-linked-data-principles) currently states:

> It should be possible to transform representations back and forth between triples and the JSON-LD serialization without loss, but not necessarily without the use of custom code.

I think this should be more clearly restricted to a single document because it is not realistic to think of round-tripping via a big-triplestore-in-the-sky where data from multiple JSON-LD documents is merged. Although it would be possible to pull out a particular Manifest or Collection, it would not be possible to infer which parts of the description were included.

azaroth42:
Closes #1267 ; Closes #1144 ; clarifies that the round trip is for a single document, and mushing everything into a triplestore might result in documents changing shape.

### DIFF
--- a/source/api/annex/notes/design_patterns.md
+++ b/source/api/annex/notes/design_patterns.md
@@ -78,7 +78,7 @@ The intent of adopting this pattern is ensure performance and simplicity of impl
 
 ### 2.7. Follow Linked Data Principles
 
-IIIF specifications conform to [Linked Data][lod], and relevant [web architecture][webarch] standards as defined by the W3C and IETF. They should not require an RDF based development stack to implement, but it must be possible to implement using one.  It should be possible to transform representations back and forth between triples and the [JSON-LD][json-ld] serialization without loss, but not necessarily without the use of custom code.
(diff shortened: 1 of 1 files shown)


mikeapp:
👍

azaroth42:
I don't think it's a principle/pattern that we can use JSON-LD 1.1. To me it was more a temperature taking exercise for the community on the risk that something changes. We don't have a good place for those sorts of decisions, however.

Issue #1297, opened 2017-11-09
Document non-painting representations (e.g. 3d, pdf, office docs, etc)


azaroth42:
There are representations of objects such as 3d models, PDFs, office documents, and so forth that are renderable to a human (sometimes in the browser, sometimes not) but are not paintable to a canvas.  It would be valuable to refer to these representations in a consistent way that would allow systems that do understand them to do something appropriate.

Proposal:  Document that `rendering` with an appropriate `type` and `format` is the appropriate place for these sorts of related representations.

tomcrane:
Yes, an explicitly call out distinctions between linking to a PDF about the object (which might be reasonably linked via annotation) and a PDF version of the object (a rendering of a canvas/range/sequence).

zimeon:
:+1:

mikeapp:
👍

edsilv:
:+1:

zimeon:
Fixed by #1382

PR #1313, opened 2017-11-17
mismatched dimensions are an error

### LINKED ISSUE #1190: Painting content/annotations to canvases without all of the dimensions is an error
This applies only to when annotated-canvas / or resource is annotated onto base-canvas, **only with the `painting` motivation**.

What should a client do in this situation, and the reverse?

Cases:
 * Base canvas has h,w,d ; annotated canvas has h,w,d = Okay
 * Base canvas has h,w,d ; annotated canvas has h,w = Okay
 * Base canvas has h,w,d ; annotated canvas has d = Okay
 * Base canvas has h,w ; annotated canvas has h,w,d = Error
 * Base canvas has h,w ; annotated canvas has h,w = Okay
 * Base canvas has h,w ; annotated canvas has d = Error
 * Base canvas has d ; annotated canvas has h,w,d = Error
 * Base canvas has d ; annotated canvas has h,w = Error
 * Base canvas has d ; annotated canvas has d = Okay

Edit from Getty AV discussion:  Any attempt to annotate a higher dimension resource onto a lower dimension resource is an error.  In order to accomplish it, a higher dimension canvas should be created and have them both annotated on to that. Requires #1191. Commentary annotations on a canvas are fine with higher dimension resources (eg video _about_ a page, rather than video that represents the page).

To expand the first `Error?` case, this would be when there's a Canvas that has only height and width, because it was intended to just capture an image plus transcription, and then a new Canvas with only duration (as an audio track, with transcription) is annotated on to it ... there's no way to align the audio canvas with the spatial canvas.  Or the second error, switch the two Canvases such that the image canvas is annotated onto the audio track canvas.

The options I think are:

* Make them error states, and require that the base canvas have all of the dimensions that the annotated canvas does
* Clients should infer the additional dimensions to the base canvas from the annotated canvas for the purposes of rendering ; Rob doesn't like this as when you have more than one conflicting annotated canvas (e.g. two different h,w canvases annotated to a duration canvas) then you can't tell which height and width should be inferred
* Other?

azaroth42:
Closes #1190 ; And towards #1297.  Specify that mismatched dimensions are an error state, with example


source/api/presentation/3.0/index.md
```diff
@@ -1074,11 +1074,11 @@ The Canvas represents an individual page or view and acts as a central point for
...
+A Canvas _MUST_ have a rectangular aspect ratio (described with the `height` and `width` properties) and/or a `duration` to provide an extent in time. These dimensions allow resources to be associated with specific regions of the Canvas, within the space and/or time extents provided. Content _MUST NOT_ be associated with space or time outside of the Canvas's dimensions, such as at coordinates below 0,0, greater than the height or width, before 0 seconds, or after the duration. Content resources that have dimensions which are not specified in the Canvas _MUST NOT_ be associated with the Canvas. For example it is an error to associate a Video resource (with height, width and duration) with a Canvas that does not have all all three dimensions. The resource can be referenced with the `rendering` property, or by Annotations with a `motivation` other than "painting", in the `annotations` property.
```
tomcrane: I think this would benefit from a second example:

For example, it is an error to use a `painting` annotation to associate a Video resource (which has height, width and duration) with a Canvas that does not have all all three dimensions, but it is not an error to use a `painting` annotation to associate an image (which has width and height only) with a Canvas that has all of width, height and duration, either for the entire duration of the Canvas or for part of that duration. 

Where the resource has dimensions that are not available in the target Canvas, the resource can be referenced with the `rendering` property, or by Annotations with a `motivation` other than "painting", in the `annotations` property.

(the wording can be tightened up further)

azaroth42:
Okay to merge?

source/api/presentation/3.0/index.md
```diff
@@ -1088,11 +1088,11 @@ The Canvas represents an individual page or view and acts as a central point for
...
+A Canvas _MUST_ have a rectangular aspect ratio (described with the `height` and `width` properties) and/or a `duration` to provide an extent in time. These dimensions allow resources to be associated with specific regions of the Canvas, within the space and/or time extents provided. Content _MUST NOT_ be associated with space or time outside of the Canvas's dimensions, such as at coordinates below 0,0, greater than the height or width, before 0 seconds, or after the duration. Content resources that have dimensions which are not defined for the Canvas _MUST NOT_ be associated with that Canvas. For example, it is an error to associate a Video resource (which has height, width and duration) with a Canvas that does not have all all three dimensions, but it is _not_ an error to use a "painting" Annotation to associate an Image (which has only height and width) with a Canvas that has all three. Such a resource _SHOULD_ instead be referenced with the `rendering` property, or by Annotations with a `motivation` other than "painting", in the `annotations` property.
```
zimeon: "all all three dimensions." -> "all three dimensions."
"all three." -> "all three dimensions." (seems better to explicitly include dimensions)

source/api/presentation/3.0/index.md
```diff
@@ -1088,11 +1088,11 @@ The Canvas represents an individual page or view and acts as a central point for
...
+Content resources are associated with the Canvas via Web Annotations.  Content that is to be rendered as part of the Canvas _MUST_ be associated by an Annotation with the "painting" `motivation`. These Annotations are recorded in the `items` of one or more AnnotationPages, refered to in the `items` array of the Canvas. If the Annotation should be rendered quickly, in the view of the publisher, then it _SHOULD_ be embedded within the Manifest directly.  Other AnnotationPages can be referenced with just their `id`, `type` and optionally a `label`, and clients _SHOULD_ dereference these pages to discover further content.  Content in this case includes media assets such as images, video and audio, textual transcriptions or editions of the Canvas, as well as commentary about the object represented by the Canvas.  These different uses _MAY_ be split up across different AnnotationPages.
```
zimeon: Q: I wonder about the use of "publisher" as assumed "publisher of the thing we are talking about" -- there are already a couple of instances in prezi3. Is this a pattern we want to use generally? Does it need any introduction?

source/api/presentation/3.0/index.md
```diff
@@ -1088,11 +1088,11 @@ The Canvas represents an individual page or view and acts as a central point for
...
 
-Content resources are associated with the Canvas via Web Annotations. The Annotations are recorded in the `items` of one or more AnnotationPages, refered to in the `content` array of the Canvas. If the Annotation should be rendered quickly, in the view of the publisher, then it _SHOULD_ be embedded within the Manifest directly.  Other AnnotationPages can be referenced with just their `id`, `type` and optionally a `label`, and clients _SHOULD_ dereference these pages to discover further content.  Content in this case includes media assets such as images, video and audio, textual transcriptions or editions of the Canvas, as well as commentary about the object represented by the Canvas.  These different uses _MAY_ be split up across different AnnotationPages.
+Renderers _MUST_ scale content into the space represented by the Canvas, and _SHOULD_ follow any `timeMode` adjustment provided for time-based media.  If the Canvas represents a view of a physical object, the spatial dimensions of the Canvas _SHOULD_ be the same scale as that physical object, and images _SHOULD_ depict only the object.
```
zimeon: Is `timeMode` an adjustment? Sounds a bit odd to me but I'm struggling to come up with something better. "definition", "specification", "instruction"???

source/api/presentation/3.0/index.md
tomcrane: :+1:

source/api/presentation/3.0/index.md
tomcrane: It might be worth formally defining "publisher". Is there a better word? What about dynamic manifests?

azaroth42: "the organization providing the manifest"?

source/api/presentation/3.0/index.md
azaroth42: In the docs, timeMode is a "mode" :|  
We shouldn't call it a behavior for confusion with thing-formerly-known-as-viewingHint.

How about just  "follow any `timeMode` value" ?

source/api/presentation/3.0/index.md
```diff
@@ -1088,11 +1088,11 @@ The Canvas represents an individual page or view and acts as a central point for
...
+A Canvas _MUST_ have a rectangular aspect ratio (described with the `height` and `width` properties) and/or a `duration` to provide an extent in time. These dimensions allow resources to be associated with specific regions of the Canvas, within the space and/or time extents provided. Content _MUST NOT_ be associated with space or time outside of the Canvas's dimensions, such as at coordinates below 0,0, greater than the height or width, before 0 seconds, or after the duration. Content resources that have dimensions which are not defined for the Canvas _MUST NOT_ be associated with that Canvas. For example, it is an error to associate a Video resource (which has height, width and duration) with a Canvas that does not have all three dimensions, but it is _not_ an error to use a "painting" Annotation to associate an Image (which has only height and width) with a Canvas that has all three dimensions. Such a resource _SHOULD_ instead be referenced with the `rendering` property, or by Annotations with a `motivation` other than "painting", in the `annotations` property.
```
tomcrane: I don't know why this didn't stick out before, but the last sentence here ("Such a resource _SHOULD_...") refers to the resource mentioned in the first clause of the preceding sentence (the error case) rather than the resource mentioned in the second, which now looks confusing to me. We could swap the resources round so that the last sentence refers to the resource just described:

> For example, it is valid to use a "painting" Annotation to associate an Image (which has only height and width) with a Canvas that has height, width and duration, but it is an error to associate a Video resource (which has height, width and duration) with a Canvas that only has height and width. Such a resource _SHOULD_ instead be referenced with the `rendering` property, or by Annotations with a `motivation` other than "painting", in the `annotations` property.

mikeapp: 👍

source/api/presentation/3.0/index.md
```diff
@@ -1088,11 +1088,11 @@ The Canvas represents an individual page or view and acts as a central point for
...
+Content resources are associated with the Canvas via Web Annotations.  Content that is to be rendered as part of the Canvas _MUST_ be associated by an Annotation with the "painting" `motivation`. These Annotations are recorded in the `items` of one or more AnnotationPages, refered to in the `items` array of the Canvas. If, according to the organization providing the Manifest, clients _SHOULD_ render the Annotation quickly then it _SHOULD_ be embedded within the Manifest directly.  Other AnnotationPages can be referenced with just their `id`, `type` and optionally a `label`, and clients _SHOULD_ dereference these pages to discover further content.  Content in this case includes media assets such as images, video and audio, textual transcriptions or editions of the Canvas, as well as commentary about the object represented by the Canvas.  These different uses _MAY_ be split up across different AnnotationPages.
```
mikeapp: refered -> referred

azaroth42: :+1: fixed

PR #1316, opened 2017-11-17
behavior and viewingDirection

### LINKED ISSUE #1073: rename viewingHint to renderingHint
From Editors discussion following Feb. A/V meeting in London: Now that we're dealing with A/V, (A in particular for this question), should we rename viewingHint ro renderingHint?

azaroth42:
Closes #1073 ; rename viewingHint/renderingHint to behavior, and viewingDirection/renderingDirection back to viewingDirection.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -434,21 +434,21 @@ The value _MUST_ be a non-negative floating point number.
 {"duration": 125.0}
 ```
 
-##### renderingDirection
+##### viewingDirection
 The direction that a set of Canvases _SHOULD_ be displayed to the user. This specification defines four direction values in the table below. Others may be defined externally and given as a full URI.
 
 The value _MUST_ be a string, taken from the table below or a full URI.
 
- * A Collection _MAY_ have exactly one `renderingDirection`, and if so, it applies to the order in which its members are rendered.<br/>
-   Clients _SHOULD_ process `renderingDirection` on a Collection.
- * A Manifest _MAY_ have exactly one `renderingDirection`, and if so, it applies to all of its sequences unless the sequence specifies its own value.<br/>
-   Clients _SHOULD_ process `renderingDirection` on a Manifest.
- * A Sequence _MAY_ have exactly one `renderingDirection`.<br/>
-   Clients _SHOULD_ process `renderingDirection` on a Sequence.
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -459,33 +459,33 @@ The value _MUST_ be a string, taken from the table below or a full URI.
...
 
 ``` json-doc
-{"renderingDirection": "left-to-right"}
+{"viewingDirection": "left-to-right"}
 ```
 
-##### renderingHint
-A hint to the client as to the most appropriate method of displaying the resource. This specification defines the values specified in the table below. Others may be defined externally, and would be given as a full URI.
+##### behavior
+A set of flags that describe requested methods for the client to use when displaying the resource to the user. This specification defines the values specified in the table below. Others may be defined externally, and would be given as a URI.
```
tomcrane: I don't like "methods"...

A set of flags that provide the client with details of the required user experience for the resource.

zimeon: :+1 on @tomcrane's suggestion of focusing on the UX expected, otherwise PR looks good

jpstroop: Agree with this change.

mikeapp: Not all behaviors are MUST, is ‘required’ too strong?   Otherwise +1

jpstroop: Reading again, I'm not sure that I like "flags" either.

tomcrane: "properties"?

tomcrane: "hints" as an echo of their former name? They _are_ hints; the particular user experience a client viewer might construct to convey the pagedness of a digitised book (for example) is up to the imagination of viewer developers.

zimeon: Using "hints" in place of "flags" would be fine by me

azaroth42: Definitely not "properties", as that sounds like json object properties. Without hint in the name, it seems strange to call them hints. Given the rename to behavior, they seem a little more ... insistent? ... than just a "hint". A request to the viewer to do something that shouldn't be ignored, rather than a hint that the experience might be better. "behaviors" ?

PR #1317, opened 2017-11-17
remove datatype assertions from contexts

### LINKED ISSUE #1272: xsd:integer data type assignment on JSON schema native type prevents compaction
ref #1252 
datatype assignment in `@context` should be removed since the JSON schema explicitly defines an unquoted value as an integer.

azaroth42:
Closes #1272 ; removes data types and class assertions from contexts, which prevent compaction (and don't do what we thought they did anyway)


mikeapp:
👍

PR #1318, opened 2017-11-17
URL -> URI

### LINKED ISSUE #1314: 3.0 Spec issue: refine Object definitions for "string" and "URI"
there are many references to "string" and "URI" in the specification.  As a programmer, my interpretation of these references as Objects is that they are ambiguous and not consistent with the JSON-LD spec. 

Specifically, I have a problem with "The value MUST be a string." where the "value" is actually a "[term](https://json-ld.org/spec/latest/json-ld/#dfn-terms)".  A JSON-LD value has a unique definition in that it is never an IRI.  The confusion between value and IRI terminology in the spec is significant enough to merit further review.

Recommend qualifying "string" as a "JSON string", "JSON-LD value", "IRI string" or "String".  Also, for clarity, a distinction between "IRI", "URI" and "URL" could be established similar to [this](https://www.w3.org/TR/activitystreams-core/#urls) with IRI being the default used in the specification

azaroth42:
Closes #1314
Fix URL to URI in Authentication section (being the only place it occurs)


mikeapp:
👍

PR #1319, opened 2017-11-17
add id attrs to tables

### LINKED ISSUE #1239: add id to each table in Presentation spec
It would be nice if there were ids in the HTML output so that it is possible to link directly to various tables in the specification. I'm thinking particularly of the "Summary of Metadata Requirements" section of the Presentation spec.

There's a link to the [heading for the section](http://iiif.io/api/presentation/2.1/#b-summary-of-metadata-requirements), but I'm not seeing a way to deep link to a table like the "Paging Properties".

azaroth42:
Closes #1239 ; adds id via # kramdown syntax to each table in prezi3 spec


mikeapp:
👍 Running locally the links work as expected.

PR #1326, opened 2017-11-27
description -> summary (#1242)

### LINKED ISSUE #1242: Description should be short abstract of resource (split1)
Original: We don't really need `description` as there's no limit on the number of characters that can go into a `metadata` value. It would also allow for setting the `label` to something else, and for it to be appropriately internationalized. At the moment it's impossible to override the client provided `label`. The use case for keeping `description` separate is if it should be presented in a different discursive block of text, compared to the more fielded values in `metadata`.  This could be a flag on the `metadata` "pair" instead?

2017-09-29: See https://github.com/IIIF/api/issues/1242#issuecomment-333181378 for updated issue

azaroth42:
Closes #1242 ; replaces description with summary throughout.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -200,22 +200,22 @@ Clients _SHOULD_ display the pairs in the order provided. Clients _SHOULD NOT_ u
 {"metadata": [ {"label": {"en": ["Creator"]}, "value": {"en": ["Anne Artist (1776-1824)"]}} ]}
 ```
 
-##### description
-A short textual description or summary of the resource that the property is attached to, intended to be conveyed to the user when the `metadata` fields are not being displayed.
+##### summary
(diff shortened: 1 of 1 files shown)


azaroth42:
Failure is an unrelated link checker 4xx/5xx. ( #1329 )

zimeon:
Maybe I was hasty with review --- the example shows ONE `summary` not in an array. Do we want to allow more than one, I don't see why that is helpful but if we do allow multiple then perhaps we need client guidelines as to what to do (concatenate?)

azaroth42:
Hasty with comment? It's a language map, so multiple values would be per language key in the map :)

zimeon:
Then the text is unclear -- I read _"one or more `description`s"_ to mean _"one or more `description` properties, each of which is a language map"_

zimeon:
This ties with a second comment I wondered about -- should we avoid pluralizing property names by instead saying _"`summary` properties"_ which also makes it clear which thing has a given cardinality.

azaroth42:
We'll have the same issue with all languageMap fields ... so should create a new issue to resolve consistently?

PR #1327, opened 2017-11-27
Add prezi3 to /api in draft list

azaroth42:
Add prezi3 link to the draft specs.
Remove request for implementation and feedback for Auth draft.
Clarify that we don't guarantee availability of non current major versions of resources (re v1.0 contexts)


azaroth42:
Failure is unrelated link checker error preventing deployment. ( #1329 )

PR #1341, opened 2017-12-06
1322 this resource

### LINKED ISSUE #1322: Consistency:  "this resource"
We have several ways of describing "the object that has the property currently being defined". 

For consistency, how about variations on "this resource" for the class of resource(s) involved:

 * this resource - any type of resource
 * this IIIF resource - any class we define in the spec
 * this content resource - content bearing, binary resources (image, video etc)
 * this external resource - other, referenced resources

With a preference for the simplest "this resource" unless the extra word is needed for clarity.

So for `label`: 
> A human readable label, name or title for this resource.

azaroth42:
Closes #1322 ; make section 3 consistent (and update `annotations` description)

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -181,7 +181,7 @@ The value of the property _MUST_ be a JSON object, as described in the [language
 ```
 
 ##### metadata
-A list of descriptive entries, given as pairs of human readable `label` and `value` entries to be displayed to the user. There are no semantics conveyed by this information, only strings to present to the user when the user is interacting with this resource.  A pair might be used to convey information about the creation of the object, a physical description, ownership information, and many other use cases. 
+A list of descriptive entries to be displayed to the user when they interact with this resource, given as pairs of human readable `label` and `value` entries. There are no semantics conveyed by this information, only strings to present to the user when the user is interacting with this resource.  A pair might be used to convey information about the creation of the object, a physical description, ownership information, and many other use cases. 
 
 The value of the `metadata` property _MUST_ be an array of objects, where each object has both `label` and `value` properties. The values of both `label` and `value` _MUST_ be JSON objects, as described in the [languages][languages-prezi30] section.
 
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -181,7 +181,7 @@ The value of the property _MUST_ be a JSON object, as described in the [language
 ```
 
 ##### metadata
-A list of descriptive entries, given as pairs of human readable `label` and `value` to be displayed to the user. There are no semantics conveyed by this information, only strings to present to the user.  A pair might be used to convey to the user information such as information about the creation of the object, a physical description, or ownership information, amongst other use cases.
+A list of descriptive entries to be displayed to the user when they interact with this resource, given as pairs of human readable `label` and `value` entries. There are no semantics conveyed by this information, only strings to present to the user when the user is interacting with this resource.  A pair might be used to convey information about the creation of the object, a physical description, ownership information, and many other use cases.
```
zimeon: simplify: "...present to the user when the user is interacting..." -> "...present to the user when they are interacting..." or even shorter "...present to the user when interacting..."

source/api/presentation/3.0/index.md
```diff
@@ -155,7 +155,7 @@ Other properties are allowed, either via custom extensions or endorsed by IIIF.
 ###  3.1. Descriptive Properties
 
 ##### label
-A human readable label, name or title for the resource. This property is intended to be displayed as a short, textual surrogate for the resource if a human needs to make a distinction between it and similar resources, for example between pages or between a choice of images to display.
```
zimeon: To avoid multiple "thus" I suggest starting second sentence with "The `label` property"

zimeon: Seems that I tied comment to the deleted line when it is intended to apply to the replacement

source/api/presentation/3.0/index.md
tomcrane: +1 to "when interacting"

source/api/presentation/3.0/index.md
```diff
@@ -543,7 +540,7 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id`,
...
+A link to an external resource that is an alternative, non-IIIF representation of this resource. The external resource _MUST_ be able to be displayed directly to a human user, and _MUST NOT_ have a splash page or other interstitial resource that gates access to it. If access control is required, then the [IIIF Authentication API][auth] is _RECOMMENDED_. Examples might include the preferred viewing environment for the IIIF resource, such as a viewer page on the publisher's web site. Other uses include a rendering of a manifest as a PDF or EPUB with the images and text of the book, or a slide deck with images of the museum object.
```
tomcrane: "controls" instead of "gates"?

azaroth42: It doesn't necessarily control access either, it might just act as a gateway (hence "gates", which I agree could be better). How about "mediates" ?

Issue #1258, opened 2017-09-25
Add `transcribing` motivation in addition to `painting`


tomcrane:
From AV Working Group:

Although a client might render captions (e.g., on a video) in the same way as other painting motivation annotations, a specific `captioning` motivation would allow the client to present UI to the user to choose captions, e.g., for compliance with accessibility requirements, and for a better user experience. This is a common feature of video players that a user would expect to see.

tomcrane:
This could also be used by a viewer that is capable of rendering WebVTT to tell that some of the content annotations are captions, and it should not render the WebVTT annotations (see https://raw.githubusercontent.com/IIIF/iiif-av/master/source/api/av/examples/11a.json).

jeffreycwitt:
This seems to be a good example of a larger issue where clients should be expected to display different "kinds" of texts in different way. A generic "painting" motivation doesn't seem to be enough to alert a client to these difference. 

To the "captioning" motivations, others like "normalized-transcription", "diplomatic-transcription", and "translation" could be added. I suspect there are others. A client might want to be able to create different views for these different "kinds" of texts. 

In the past I've been handling this with Layers, but it would be best to handle this in a way that is consistent with what the A/V groups is doing. 

In short these seems like an issue bigger than A/V. Is it possible to add a more general github label, so that this issue can be discussed as part of a larger discussion of IIIF supported motivations?

zimeon:
Note relation to #511

zimeon:
Sense in Toronto discussion is that we want to allow refining motivations without adding an ever increasing number of motivations to the Presentation API specification. Thus, we might add a `captioning` motivation in another namespace (community controlled list, say `iiifc:` prefix for illustration, the prefix would be defined in presentation context) but this would be in addition to the `painting` motivation (that is in the spec), and would thus also support a local addition `ext:special-ing` in a set of motivations:

```
"motivation": ["painting", "iiifc:captioning", "ext:special-ing"]
```

Would need to understand process for consideration of how to add other motivations to the community namespace.

tomcrane:
The more I think about this the less sure I am that `captioning` - the suggestion example that leads to these issues - is a subclass of `painting`. It doesn't have to be, in the list suggestion above. Many clients will render such annotations on top of the image or video, and often the annotation will target fragments of the time dimension only. But the intent of `captioning` can be met by other means that would still satisfy the publisher of those captions.

Here's what the spec says about `painting`:

[...]

PR #1346, opened 2017-12-14
update canvas annopage rendering requirement

### LINKED ISSUE #1201: Clients SHOULD process inline annotationPages first
In discussion of #1068, we have agreed to combine presentation 2.0 `images` and `otherContent` into a single `content` in presentation 3.0 which includes inline and referenced `AnnotationPage` entries. In 2.0, the distinction served to indicate what should be painted immediately (`images`) or loaded later (`otherContent`). What are the rules with `content` (inline vs referenced?) and are any explicit hints necessary?

azaroth42:
Closes #1201 ; adds more text to requirements about prioritization of annotation pages


azaroth42:
I think that the split is intended to be both ways, right? So the answer should be no you can't but painting under annotations.  Clarifying this in #1351 for #1258.

tomcrane:
The spilt is both ways, but it wasn't 100% obvious. It may seem natural to many people to put huge numbers of word-level transcription annos under `annotations` rather than `items`, because they feel a lot less content-y.

For a Canvas, annotations under `items` MUST have either `painting` or `transcribing` motivations, because this is where you put the content of the canvas; annotations under `annotations` MUST NOT have those motivations, because that's where you put everything else. The latter is addressed in http://prezi3.iiif.io/api/presentation/3.0/#annotations but the general idea is still spread out a little, the rewrite of 1,2 + 3 can bring this together.

PR #1261, opened 2017-09-26
add links include

jpstroop:
Proposal. Now that the site has been pared down, centrally manage links in an include? I have constantly having to scroll to the bottom.

PR #1347, opened 2017-12-14
1121 services

### LINKED ISSUE #1121: Create/Assign @type to info.json resources
The proposal is to add a *mandatory* `@type` to info.json resources, such as `iiif:ImageService` in 3.0
The rationale is a bit longer to explain...

In JSON-LD 1.1 there will be the facility to do per-predicate and per-class (`@type`) contexts, nested within the main context document.  This is described here: http://json-ld.org/spec/latest/json-ld/#scoped-contexts

We can avoid the current known issue of having to use `@context` keys within the structure by having sufficient hooks to refer to them by. When expanding the JSON-LD to RDF, the contexts disappear as they're not themselves triples. Thus when expanding (e.g. as part of framing) and then later compacting back to the recommended JSON form, the contexts are lost.  With scoped contexts, the main context keeps the references rather than the data.  Non-core extensions would still need to be re-injected ... but we have yet to see [m]any of these.

As our extension point is just `service`, we can't rely on predicate based scoped contexts, so need to use class based, and our core services currently don't have them.

In (e.g.) the Presentation context we would have:

```
  "IIIFImageService": {"@id": "iiif:ImageService", 
      "@type": "@id", 
      "@context": "http://iiif.io/api/image/3/context.json"},
```

And then in the data:

```
  "service": {
      "id": "http://example.org/iiif/identifier",
      "type": "IIIFImageService",
      "profile": "http://iiif.io/api/image/2/level1.json"
  }
```

And then the same for all other services that we intend to embed, notably auth and search.

azaroth42:
Closes #1121 ; add ImageService1,2 and SearchService1 to Prezi3

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -556,7 +556,7 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id`, 
 ```
 
 ##### service
-A link to an external service that the client might interact with directly and gain additional information or functionality for using this resource, such as from an image to the base URI of an associated [IIIF Image API][image-api] service. The service resource _SHOULD_ have additional information associated with it in order to allow the client to determine how to make appropriate use of it. Please see the [Service Registry][registry-service] document for the details of currently known service types.
+A link to an external service that the client might interact with directly and gain additional information or functionality for using this resource, such as from an image to the base URI of an associated [IIIF Image API][image-api] service. The service resource _SHOULD_ have additional information associated with it in order to allow the client to determine how to make appropriate use of it. Please see the [Service Registry][registry-services] document for the details of currently known service types.
 
 The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id` and `type` properties, and _SHOULD_ have the `label` and `profile` properties.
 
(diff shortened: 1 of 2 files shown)


azaroth42:
Changed "a few" to "the following" ... sorry for making reviews stale!

source/api/presentation/3.0/index.md
```diff
@@ -571,6 +571,29 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id` a
   }]}
 ```
 
+For cross-version consistency, this specification defines the following types for backwards compatibility with other IIIF APIs. Future versions of these APIs will define their own types.
+
+| Value          | Specification |
+| -------------- | ------------- |
+| ImageService1  | [Image API version 1][]  |
+| ImageService2  | [Image API version 2][]  |
+| SearchService1 | [Search API version 1][] |
```
jpstroop: Why the empty links?

azaroth42: Ahh... assistance please in getting the new link tricks to work?

jpstroop: What's the trick? Something related to https://github.com/IIIF/api/pull/1261?

azaroth42:
I think we need to resolve #1352 before we can merge this.

azaroth42:
Up to date with the links. Also fixed lots and lots of other links to be more regular and uniquely named. Added services and links for the distinction between token/cooke/logout, and search vs autocomplete.

azaroth42:
The normative / text differences are limited to http://preview.iiif.io/api/1121-services/api/presentation/3.0/#service

The rest of the changes are the link un-breaking.

source/api/presentation/3.0/index.md
```diff
@@ -1513,7 +1540,7 @@ Recipes for enabling CORS, conditional Content-Type headers and other technical
 
 ## 7. Authentication
 
-It may be necessary to restrict access to the descriptions made available via the Presentation API.  As the primary means of interaction with the descriptions is by web browsers using XmlHttpRequests across domains, there are some considerations regarding the most appropriate methods for authenticating users and authorizing their access.  The approach taken is described in the [Authentication][auth] specification, and requires requesting a token to add to the requests to identify the user.  This token might also be used for other requests defined by other APIs.
+It may be necessary to restrict access to the descriptions made available via the Presentation API.  As the primary means of interaction with the descriptions is by web browsers using XmlHttpRequests across domains, there are some considerations regarding the most appropriate methods for authenticating users and authorizing their access.  The approach taken is described in the [Authentication][auth-api] specification, and requires requesting a token to add to the requests to identify the user.  This token might also be used for other requests defined by other APIs.
```
jpstroop: Style question: should XmlHttpRequests be `XmlHttpRequest`s?

azaroth42: Maybe? We don't tend to `backtick` our own class names, so I'd be at least hesitant to do it here ... but I don't think we have any consistency.

jpstroop: Make a separate ticket and move on? This is easily solved in a pub.

azaroth42: Yup, 👍  Will make the issue now.

azaroth42: --> #1371

PR #1351, opened 2017-12-14
1258 1310 transcribing + registries

### LINKED ISSUE #1258: Add `transcribing` motivation in addition to `painting`

### LINKED ISSUE #1310: Create registry-style "annex" document
To ensure (or at least promote) consistency, it seems like we need a registry for:

* `type` : e.g. that a PDF is a `Text` not a `Dataset`, an Image API is a `ImageService3` etc
* `profile` : e.g. the profiles for various metadata schemas and services
* `service` : e.g. the set of known services
* property extensions : e.g. the set of known, registered extensions
* value extensions : e.g. set of known, registered additions to enumerated values like `timeMode`, `behavior` etc.

This is blocking at least #1121, #1269

azaroth42:
Closes #1258 ; overreaches the issue massively by creating a skeleton set of registries, but need them to refer to from Prezi3 API docs...

Closes #1310 

Thoughts on text, and on registries?


azaroth42:
> Would the motivation registry repeat the W3C Anno spec's motivations? They are special because they are already included in Prezi3 via the context and mentioned in the new text, but it would seem odd if they weren't mentioned on the motivations registry page.

Yes, I think so too.

> I think each registry document should link to the document listing the other registries

Done.

This issue spans all of the APIs. I think the cookbook is a Prezi specific issue ... but agree we can at least create the splash page for it now... but where?

`annex/cookbook/` on the grounds that there's unlikely to be an auth or image cookbook?

tomcrane:
There might be auth recipes in a cookbook - here's how you do "everyone can deep zoom, only authed users can download derivatives > 1000px" - but they don't need to be in a _separate_ cookbook.

azaroth42:
We could take the cookbook discussion to an issue?

azaroth42:
Cookbook meta issue:  #1367

azaroth42:
Closes #1310

azaroth42:
Reviews please? I think this is ready to go

source/_includes/links.md
```diff
@@ -30,10 +30,13 @@
 [openannotypes]: http://www.openannotation.org/spec/core/core.html#BodyTargetType
 [paging-prezi30]: {{ site.url }}{{ site.baseurl }}/api/presentation/3.0/#paging-properties
 [prev-version]: {{ site.url }}{{ site.baseurl }}/api/presentation/2.0/
+[prezi-api]: {{site.url}}{{ site.baseurl }}/api/presentation/3.0/
```
jpstroop: maybe `prezi-api30` or something along those lines?

azaroth42: We should also take out prev-version then, I guess?

azaroth42:
I'm :-1: on plural ... everything else is singular ... note `/api/`  `/event/`  `/api/annex/` etc despite there being multiple APIs, Events and Annexes.

We could move them, but I think that would mean we need to be clearer about what goes into the annex space and what goes into the top level /api/ space.  Currently, other than annex, only APIs live there.  If registries go there, then definitions should go there to, such as the Image API Selector and IIIF defined services. And then it starts to get confusing, IMO, about what's an API and what's just additional normative content.

jpstroop:
>I'm 👎 on plural ... everything else is singular ... note /api/ /event/ /api/annex/ etc despite there being multiple APIs, Events and Annexes.

They really should be plural, but whatever.

azaroth42:
No disagreement! But we should be consistent, and we're stuck with at least /api/.

PR #1356, opened 2017-12-17
1263 posterCanvas

tomcrane:
I have this in Descriptive properties, right after thumbnail. But does it belong there? Could be structural. I won't do the table at the bottom until confirming where it goes.


azaroth42:
The text refers to duration only canvases, but the cardinality block allows association with most resource types.  *We* know this is because the client should use the posterCanvas at the highest level in the structure it has, and hence if the Manifest has a posterCanvas then it gets presented across all Canvases ... but I think that could be spelt out in the text

source/api/presentation/3.0/index.md
```diff
@@ -240,6 +240,19 @@ The value _MUST_ be a JSON array, with each item in the array being a JSON objec
 {"thumbnail": [{"id": "https://example.org/img/thumb.jpg", "type": "Image"}]}
 ```
 
+##### posterCanvas
+A single Canvas providing content that represents this resource, such as image to show while a duration-only Canvas is playing audio, or images, text and sound shown for video content before the user initiates playback. Whereas a client might use `thumbnail` for navigation or summary views of multiple resources, a client can choose to show content from `posterCanvas` as part of the presentation of this resource without requiring the content of `posterCanvas` to be part of this resource. The poster Canvas can have different dimensions to the Canvas(es) of this resource.
```
azaroth42: "represents" seems too restrictive -- an ad for a product that's displayed during the audio of a podcast manifest ... does the ad represent the manifest? 
represent -> is associated with ?

azaroth42: Is there a use case for canvases without duration to have a posterCanvas?  If a posterCanvas is associated with a Manifest that contains Canvases without duration (and some that do) ... what does the client do? Flick between the h/w canvas content and the poster content?

zimeon: ... such as **an** image to show ...

source/api/presentation/3.0/index.md
```diff
@@ -240,6 +240,19 @@ The value _MUST_ be a JSON array, with each item in the array being a JSON objec
...
+
+  * A Collection _MAY_ have exactly one `posterCanvas`.
```
azaroth42: Clients _MAY_ ... for all? _SHOULD_ for Manifest and Canvas?

source/api/presentation/3.0/index.md
```diff
@@ -240,6 +240,19 @@ The value _MUST_ be a JSON array, with each item in the array being a JSON objec
...
+  * A Manifest _MAY_ have exactly one `posterCanvas`.
+  * A Sequence _MAY_ have exactly one `posterCanvas`.
+  * A Canvas _MAY_ have exactly one `posterCanvas`.
+  * A Range _MAY_ have exactly one `posterCanvas`.
+  * Other resource types _MUST NOT_ have a poster canvas.
+
+Publishers should not assume that a client will display the content of the poster Canvas to the user.
```
zimeon: Could add here a comment along the lines of @azaroth42's https://github.com/IIIF/api/pull/1356#issuecomment-356680303

source/api/presentation/3.0/index.md
tomcrane: done

source/api/presentation/3.0/index.md
tomcrane: posterCanvas for Range should be shown in preference to posterCanvas for Manifest?

source/api/presentation/3.0/index.md
tomcrane: Avoided notion of representation - instead, _associated with but not part of_

How much should the spec say about sensible use of posterCanvas? I think the spec should be silent on complexities like "Flick between the h/w canvas content and the poster content" with just the general indication that the client _MAY_ show the most appropriate content for the user's context. In practice I would hope that use of posterCanvas is limited to Manifest and perhaps Range (segments or episodes maybe).

tomcrane:
I have put `posterCanvas` in the Descriptive properties table, next to `thumbnail` as it is in the text. Cardinality addressed for future-proofing.

source/api/presentation/3.0/index.md
```diff
@@ -240,6 +240,19 @@ The value _MUST_ be a JSON array, with each item in the array being a JSON objec
...
+
+Clients _MAY_ display the content of a linked poster Canvas when presenting the resource, and if more than one is available _MAY_ choose the one most suited to the client user interface. When more than one Canvas is available, for example if `posterCanvas` is provided for the currently selected Range and the current Manifest, and the client is able to show a poster Canvas, the client _SHOULD_ pick the Canvas most specific to the content. Publishers _SHOULD NOT_ assume that `posterCanvas` content will be seen in all clients. Clients _SHOULD_ take care to avoid conflicts between time-based media in the `posterCanvas` and the content of the resource it is associated with.
+
+  * A Collection _MAY_ have one or more `posterCanvas` properties.
```
azaroth42: For consistency, each bullet should also have the client requirement of MAY

tomcrane: Client reqts added. 

Style question - for plural use of `tick-quoted` properties that end in _s_

I couldn't bring myself to write any variant of `posterCanvas`s, `posterCanvas`es... so I introduced the variation "`posterCanvas` properties".

source/api/presentation/3.0/index.md
```diff
@@ -240,6 +240,25 @@ The value _MUST_ be a JSON array, with each item in the array being a JSON objec
 {"thumbnail": [{"id": "https://example.org/img/thumb.jpg", "type": "Image"}]}
 ```
 
+##### posterCanvas
+One or more Canvases providing content associated with the object represented by this resource but not part of that object. Examples include an image to show while a duration-only Canvas is playing audio; images, text and sound standing in for video content before the user initiates playback; or a film poster to attract user attention. The content provided by `posterCanvas` differs from a thumbnail: a client might use `thumbnail` to summarise and navigate multiple resources, then show content from `posterCanvas` as part of the presentation of a single resource. A poster Canvas can have different dimensions to the Canvas(es) of this resource.
```
jpstroop: > A poster Canvas can have different dimensions to the Canvas(es) of this resource.

A poster Canvas can have different dimensions to those of the Canvas(es) of this resource.

source/api/presentation/3.0/index.md
```diff
@@ -240,6 +240,25 @@ The value _MUST_ be a JSON array, with each item in the array being a JSON objec
...
+  * A Manifest _MAY_ have one or more `posterCanvas` properties.<br/>
+   Clients _MAY_ render `posterCanvas` on a Manifest.
+  * A Sequence _MAY_ have one or more `posterCanvas` properties.<br/>
+   Clients _MAY_ render `posterCanvas` on a Sequence.
+  * A Canvas _MAY_ have one or more `posterCanvas` properties.<br/>
+   Clients _MAY_ render `posterCanvas` on a Canvas.
+  * A Range _MAY_ have one or more `posterCanvas` properties.<br/>
+   Clients _MAY_ render `posterCanvas` on a Range.
+  * Other resource types _MUST NOT_ have a `posterCanvas`.<br/>
+   Clients _SHOULD_ ignore `posterCanvas` on other resource types.
```
jpstroop: > Clients _SHOULD_ ignore `posterCanvas` on other resource types.

Why not _MUST_?

tomcrane: Or leave out the last sentence altogether?

zimeon: I think we had this issue on something else in the last few days. I prefer not to have a _MUST_ for a case that isn't allowed anyway. I'd either keep as _SHOULD_ or else follow @tomcrane's suggestion and simple remove that sentence -- slight preference to remove

azaroth42: The pattern at the moment is if the resource type MUST NOT have an `x`, then clients SHOULD ignore `x` on those types.

So I'm :+1: on the current text, or propose making a new issue if we want to change it across the board in section 3.

zimeon: Agreed on call that we'll leave as _SHOULD_

tomcrane:
👍

PR #1364, opened 2018-01-23
description of thumbnail-nav behavior

tomcrane:


tomcrane:
Added a non A/V example, keeping the text length concise. The Bayeux Tapestry would be a great _specific_ example but we don't use specific examples.

PR #1366, opened 2018-01-25
Updating nokogiri and other dependencies

glenrobson:
Also updating travis build version before it goes unsupported.

Using ruby version 2.4.2


jpstroop:
Should be fine as long as kramdown stays pinned, but it might not hurt to check spot check the markdown

azaroth42:
Spot checked. Merging.

PR #1377, opened 2018-02-07
Use https for example.org URIs

### LINKED ISSUE #1334: Use https://example.org/ for content URIs in examples
We are promoting the use of HTTPS for IIIF resources so it would be best to show this in the spec examples. I think this means a search and replace of `http://example.org/` with `https://example.org/` in the prezi 3 spec.

zimeon:
Closes #1334 

Changes `http://example.org` to `https:/example.org` everywhere in new spec/annex, also removes a few `www.` prefixes


jpstroop:
👍

PR #1380, opened 2018-02-08
rename, redefine related to homepage

### LINKED ISSUE #1286: Clarify `related` to be only the "home page" of this resource
The information carried in `related` is redundant with the information that can be carried in `metadata`, without the ability to have internationalized labels.  The client never needs to have the URI to the related resource, as the only thing that it will use it for is to generate HTML to display a link to the resource, unlike `seeAlso` or `service` for example where the client will interact with the resource directly. The set of related resources is a set, not a list, and hence the order of the links to be displayed is not deterministic.

Proposal: Remove the `related` property from Presentation 3.0 as redundant, and instead just put links into `metadata`.

azaroth42:
Closes #1286


source/api/presentation/3.0/index.md
```diff
@@ -524,19 +524,19 @@ The value _MUST_ be a string, taken from the table below or a full URI.
...
+##### homepage
+A link to an external web page that primarily describes the real world object that this resource is a presentation of. The web page is usually published by the organization responsible for the real world object, and might be generated by a collection management system or hand crafted. The external resource _MUST_ be able to be displayed directly to the user.  External resources that are related, but not home pages, _MUST_ instead be added into the `metadata` property, with an appropriate `label` or `value` to describe the relationship.
 
 The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id`, `type` and `label` properties, and _SHOULD_ have a `format` property.
 
- * Any resource type _MAY_ have one or more external resources related to it.<br/>
-   Clients _SHOULD_ render `related` on a Collection, Manifest or Canvas, and _MAY_ render `related` on other resource types.
+ * Any resource type _MAY_ have exactly one homepage.<br/>
```
zimeon: Perhaps remove *exactly* ... as I read it, it seems at odds with MAY (if I were to use exactly I'd have some like "MUST have exactly zero or one")

azaroth42: That's the pattern we have at the moment.  Compare the definition of `timeMode`

zimeon: OK, will :+1: for now but might want to remove all the `exactly` words later...

azaroth42: --> #1397

source/api/presentation/3.0/index.md
```diff
@@ -524,19 +524,19 @@ The value _MUST_ be a string, taken from the table below or a full URI.
 
 #### 3.4.1 External Links
 
-##### related
-A link to an external resource that describes or is about this resource. The external resource _MUST_ be able to be displayed directly to the user. Examples might include a video or academic paper about the resource, a website, an HTML description, and so forth.
+##### homepage
+A link to an external web page that primarily describes the real world object that this resource is a presentation of. The web page is usually published by the organization responsible for the real world object, and might be generated by a collection management system or hand crafted. The external resource _MUST_ be able to be displayed directly to the user.  External resources that are related, but not home pages, _MUST_ instead be added into the `metadata` property, with an appropriate `label` or `value` to describe the relationship.
```
jpstroop: > is a presentation of

represents

>  collection management system or hand crafted.

collection management system or library catalog. ? Hand-crafted seems odd to me.

azaroth42:
Fixed conflicts.

PR #1382, opened 2018-02-08
clarify incompatible content resources and renderings

### LINKED ISSUE #1297: Document non-painting representations (e.g. 3d, pdf, office docs, etc)

azaroth42:
closes #1297


source/api/presentation/3.0/index.md
```diff
@@ -541,7 +541,7 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id`,
...
+A link to an external resource that is an alternative, non-IIIF representation of this resource. Such representations typically cannot be painted onto a single Canvas, as they either include too many views (such as a PDF with multiple pages), have incompatible dimensions (such as a 3d model with a 2d Canvas), or are composite resources requiring additional rendering functionality. The external resource _MUST_ be able to be displayed directly to a human user, and _MUST NOT_ have a splash page or other interstitial resource that mediates access to it. If access control is required, then the [IIIF Authentication API][auth-api] is _RECOMMENDED_. Examples might include the preferred viewing environment for the IIIF resource, such as a viewer page on the publisher's web site. Other uses include a rendering of a book as a PDF or EPUB, a slide deck with images of a building, or a 3d model of a statue.  External resources that are related to this resource _MUST_ instead be referenced in the `metadata` field, or linked to a particular Canvas segment with an Annotation.
```
tomcrane: > Examples might include the preferred viewing environment for the IIIF resource, such as a viewer page on the publisher's web site.

The trouble with that example is that it will often be the same page as pointed at by `homepage`. That doesn't mean it isn't a valid `rendering`, but how to avoid confusion? Should I point `homepage` and `rendering` at the same URI?

azaroth42: True. No, it should just go in homepage. Will take out the example.

azaroth42:
Fixed overlap with `homepage`.

source/api/presentation/3.0/index.md
```diff
@@ -541,7 +541,7 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id`,
...
+A link to an external resource that is an alternative, non-IIIF representation of this resource. Such representations typically cannot be painted onto a single Canvas, as they either include too many views (such as a PDF with multiple pages), have incompatible dimensions (such as a 3d model with a 2d Canvas), or are composite resources requiring additional rendering functionality. The external resource _MUST_ be able to be displayed directly to a human user, and _MUST NOT_ have a splash page or other interstitial resource that mediates access to it. If access control is required, then the [IIIF Authentication API][auth-api] is _RECOMMENDED_. Examples include a rendering of a book as a PDF or EPUB, a slide deck with images of a building, or a 3d model of a statue.  External resources that are related to this resource _MUST_ instead be referenced in the `metadata` field, or linked to a particular Canvas segment with an Annotation. Any home page or preferred web view of the resource _MUST_ instead be referenced in the `homepage` field.
```
jpstroop: > as they either include too many views (such as a PDF with multiple pages), have incompatible dimensions (such as a 3d model with a 2d Canvas), or are composite resources requiring additional rendering functionality. 

and then a couple of sentences later:

> Examples include a rendering of a book as a PDF or EPUB, a slide deck with images of a building, or a 3d model of a statue.

Can this be tightened up to get all of the examples in one place?

Also, are we sure about all of the RFC keywords here?

azaroth42: Took out the parentheticals, as they were all in the examples below.   I think were sure about the keywords...

You MUST be able to show the resource to the user -- otherwise it's not a rendering.
There MUST NOT be a splash page -- this was my clickthrough page issue.
We RECOMMEND our own Auth pattern.
And the final two are that you must follow the spec, rather than dumping things in here.

PR #1383, opened 2018-02-08
rename startCanvas->start, clarify fragments

### LINKED ISSUE #1320: Rename startCanvas to start, allow ref to part of a Canvas
Now that Canvases can have a duration, the start of the startCanvas is not necessarily the beginning of the duration. It could be that the interesting part of the content begins on the first canvas, but not at the beginning of its duration.  This would let the player initialize to a temporal offset within the Canvas, rather than playing the 10 seconds of silence or 3 minutes of audience noise at the beginning.
For x,y canvases, this would allow scrolling to a region of the Canvas, such as the most interesting part of the 150k px Princeton scroll.

Proposal:  `startCanvas` can also refer to a Canvas with a fragment, in the same way that Ranges can.

azaroth42:
Closes #1320


azaroth42:
And once again conflicts killed the approvals :(

PR #1388, opened 2018-02-08
Clarify collection descriptions

### LINKED ISSUE #1357: Tighten definition of Collections
From #1343, and per @jpstroop on the 12/20 community call, there is some weirdness around the definition of Collections that contributes to the difficulty of paging.

2.1 definition:
> An ordered list of manifests, and/or further collections. Collections allow easy advertising and browsing of the manifests in a hierarchical structure, potentially with its own descriptive information. They can also provide clients with a means to locate all of the manifests known to the publishing institution.

And the first sentence of the Collection section:
> Collections are used to list the manifests available for viewing, and to describe the structures, hierarchies or curated collections that the physical objects are part of.

Which is shortened in 3.0 currently to get rid of the discovery use case:
> An ordered list of Manifests, and/or further Collections. Collections allow easy navigation among the Manifests in a hierarchical structure, potentially each with its own descriptive information.

Neither of which tie the iiif:Collection notion explicitly to the Curated Collection of Objects cultural heritage notion ... but that is of course how they are mostly used.

If the Collections are intended to be presentational, e.g. displayed to a human in a viewing application, then they should probably have relatively few objects, and certainly not many thousands.  However there are curated collections with many many thousands of resources in an undifferentiated pool that would need to be arbitrarily subdivided.

azaroth42:
Closes #1357

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -121,7 +121,7 @@ Content resources such as images, audio, video or text that are associated with 
 ##### Collection
 {: #overview-collection}
 
-An ordered list of Manifests, and/or further Collections.  Collections allow easy navigation among the Manifests in a hierarchical structure, potentially each with its own descriptive information.
+An ordered list of Manifests, and/or further Collections.  Collections allow easy navigation among the Manifests in a hierarchical structure, potentially each with its own descriptive information. Collections might be used to model dynamic result sets from a search, fixed sets of related resources, or other groupings of Manifests for presentation.
 
 ##### AnnotationPage
 {: #overview-annotationpage}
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -1343,22 +1343,22 @@ AnnotationCollections _MUST_ have a URI, and it _SHOULD_ be an HTTP URI.  They _
...
+Collections are used to list the manifests available for viewing, and to describe the structures, hierarchies or sets that the resources are part of.  Collections _MAY_ include both other Collections and Manifests, in order to form a tree-structured hierarchy.  Collections might be used to model dynamic result sets from a search, fixed sets of related resources, or other groupings of Manifests for presentation to the user, typically for navigation amongst the member items.
 
-Collection objects _MAY_ be embedded inline within other collection objects, such as when the collection is used primarily to subdivide a larger one into more manageable pieces, however manifests _MUST NOT_ be embedded within collections. An embedded collection _SHOULD_ also have its own URI from which the description is available.
+Collection objects _MAY_ be embedded inline within other Collection objects, such as when the Collection is used primarily to subdivide a larger one into more manageable pieces, however Manifests _MUST NOT_ be embedded within Collections. An embedded Collection _SHOULD_ also have its own URI from which the JSON description is available.
```
tomcrane: How do we make it clear that if a resource is "embedded inline", we mean that the full JSON of the resource has been included, not some reduced version? Variations on "embedded" appear quite a few times in the spec, do you think we should define this formally? In _Terminology_?

source/api/presentation/3.0/index.md
```diff
@@ -1343,22 +1343,22 @@ AnnotationCollections _MUST_ have a URI, and it _SHOULD_ be an HTTP URI.  They _
 
 ###  5.8. Collection
 
-Collections are used to list the manifests available for viewing, and to describe the structures, hierarchies or curated collections that the objects are part of.  Collections _MAY_ include both other Collections and Manifests, in order to form a tree-structured hierarchy.  
+Collections are used to list the manifests available for viewing, and to describe the structures, hierarchies or sets that the resources are part of.  Collections _MAY_ include both other Collections and Manifests, in order to form a tree-structured hierarchy.  Collections might be used to model dynamic result sets from a search, fixed sets of related resources, or other groupings of Manifests for presentation to the user, typically for navigation amongst the member items.
```
mikeapp: manifests -> Manifests

source/api/presentation/3.0/index.md
azaroth42: --> #1396

jpstroop:
Needs @mikeapp and @tomcrane

Issue #1052, opened 2017-01-27
add Japanese Auth translation


srabun:
From Shuji Kamitsuna from Japan:
"IIIF Authentication 1.0" in Japanese is available now:
http://www.asahi-net.or.jp/~ax2s-kmtn/ref/iiif/apiauthentication1.0.html

Add to: http://iiif.io/api/#community-translations

azaroth42:
Closed by #1390.

PR #1390, opened 2018-02-09
Add Japanes Auth translation

glenrobson:
Adding authentication translation as requested in #1052.


glenrobson:
http://preview.iiif.io/api/add_auth_translation/api/

Issue #1330, opened 2017-11-27
Consistently talk about properties, their cardinality and language map values


zimeon:
Issue came up in #1326 that text like _"one or more `summary`s"_ might be taken to mean either of:

  * one `summary` property, which may have one or more `summary` strings by virtue of being a language map
  * one or more `summary` properties, each of which may have multiple `summary` strings by virtue of being a language map

I like the idea of calling them "`name` properties" so we can talk about how many properties separately from how many values in language maps.

azaroth42:
I agree that the wording can be tightened up here.

To use as examples:
> The value of the property must be a JSON object, as described in the languages section.
> A Collection must have at least one `label`.

> The value must be a JSON array, with each item in the array being a JSON object that must have an id property and should have at least one of type and format
> A Collection should have exactly one `thumbnail`, and may have more than one.

The intent is, respectively, that there MUST be a `label` property and it MUST have at least one value present in its language map; and that there SHOULD be a `thumbnail` property and if there is, then there SHOULD be exactly one value (which is a JSON object), there MAY be more than one value, meaning that the type of the property's value is an array (of JSON objects).

The issue with calling them "`foo` properties" is the conflation of JSON object property and data model predicate ... there's only ever one `label` or `thumbnail` key on the JSON object, even if they have multiple values in the abstract data model.

For the language maps, how about:  

> A _ClassName_ MUST have at least one value in the `propertyName` property.

?

azaroth42:
Adding another slightly different case of behaviors or `behavior`s or `behaviors` ...

> Valid on AnnotationCollection, AnnotationPage, Annotation, SpecificResource and Choice. If this `behavior` is provided, then the client should not render the resource by default, but allow the user to turn it on and off.

Where there is both the property and a IIIF defined value for the property.

OTOH, I do NOT think we should move the values from `behavior`, `viewingDirection` and `timeMode` down in to the new values section, separate from their property definition ... raising the spectre [or is it a meltdown?] of why we have it for motivations only ...

azaroth42:
How about ...

Currently:
> A Collection MUST have at least one label.

Means:
> A Collection MUST have the label property with at least one value.

Or:
> A Collection SHOULD have one thumbnail, and MAY have more than one.

Means:
> A Collection SHOULD have the thumbnail property with at least one value, and there MAY be more than one value.

zimeon:
Fixed by #1411

PR #1392, opened 2018-02-09
add repeat, alphasort entries

### LINKED ISSUE #1328: "repeat" behavior
Dave Beaudet, in #1264:
> consider use case where you'd have an in-exhibition viewer that cycles through via auto-advance and upon reaching the end, starts over from the beginning.

I think that this is a separate behavior from "auto-advance" and needs its own identity.  I could imagine both cases where you want to repeat from the first canvas (as above), and situations where you don't (watching a multi-canvas video).

Propose:  "repeat" value for `behavior`. When this value and "auto-advance" are present, the client _SHOULD_ continue from the last canvas back to the first once the last canvas has been completed.

azaroth42:
Closes #1328


source/api/presentation/3.0/index.md
```diff
@@ -483,16 +483,17 @@ The value _MUST_ be an array of strings, taken from the table below or a URI.
...
-| `paged` | Valid on Manifest, Sequence and Range, which include Canvases that have at least `height` and `width` dimensions. Canvases included in resources with this `behavior` represent pages in a bound volume, and _SHOULD_ be presented in a page-turning interface if one is available.  The first canvas is a single view (the first recto) and thus the second canvas likely represents the back of the object in the first canvas. If this is not the case, see the "non-paged" `behavior`. |
+| `auto-advance` | Valid on Collection, Manifest, Sequence and Canvas, that include or are Canvases with at least the `duration` dimension. When the client reaches the end of a Canvas with a duration dimension that has (or is within a resource that has) this `behavior`, it _SHOULD_ immediately proceed to the next Canvas and render it. If there is no subsequent Canvas in the current context, then this `behavior` should be ignored. When applied to a Collection, the client should treat the first Canvas of the next Manifest as following the last Canvas of the previous Manifest, respecting any `startCanvas` specified. |
```
jpstroop: > Valid on Collection, Manifest, Sequence and Canvas, that include or are Canvases

Typo? Shouldn't "Collection, Manifest, Sequence and Canvas" each be pluralized? We use this construction elsewhere as well and it reads awkwardly (e.g. `repeat`, `paged`)

source/api/presentation/3.0/index.md
```diff
@@ -483,16 +483,17 @@ The value _MUST_ be an array of strings, taken from the table below or a URI.
...
 | `continuous` | Valid on Manifest, Sequence and Range, which include Canvases that have at least `height` and `width` dimensions.  Canvases included in resources with this `behavior` are partial views and an appropriate rendering might display all of the Canvases virtually stitched together, such as a long scroll split into sections. This `behavior` has no implication for audio resources. The `viewingDirection` of the Sequence or Manifest will determine the appropriate arrangement of the Canvases. |
+| `facing-pages` | Valid only for Canvas, where the Canvas has at least `height` and `width` dimensions. Canvases with this `behavior`, in a Sequence or Manifest with the "paged" `behavior`, _MUST_ be displayed by themselves, as they depict both parts of the opening.  If all of the Canvases are like this, then page turning is not possible, so simply use "individuals" instead. |
```
jpstroop: I don't think "behavior" should be in backtics here -  we're talking about _a_ behavior, not the _keyword_ `behavior`, right?

jpstroop: Applies below as welll

source/api/presentation/3.0/index.md
```diff
@@ -483,16 +483,17 @@ The value _MUST_ be an array of strings, taken from the table below or a URI.
...
+| `individuals` | Valid on Collection, Manifest, Sequence and Range. For Collections with this `behavior`, each of the included Manifests are distinct objects. For Manifest, Sequence and Range, the included Canvases are distinct views, and _SHOULD NOT_ be presented in a page-turning interface. This is the default `behavior` if none are specified. |
 | `multi-part` | Valid only on Collection. Collections with this `behavior` consist of multiple Manifests that each form part of a logical whole, such as multi-volume books or a set of journal issues. Clients might render the Collection as a table of contents, rather than with thumbnails. |
+| `no-nav` | Valid only for Range. Ranges with this `behavior` _MUST NOT_ be displayed to the user in a navigation hierarchy. This allows for Ranges to be present that capture unnamed regions with no interesting content. |
```
jpstroop: > with no interesting

Maybe explain a little more by adding a use case?

source/api/presentation/3.0/index.md
azaroth42: This is #1330, I think.  Or at least should be part of that!

source/api/presentation/3.0/index.md
azaroth42: :+1: I thought this at the time, but it wasn't part of the issue so I didn't change it.  Changing :)

source/api/presentation/3.0/index.md
azaroth42: (Leaving it for that discussion, have made a note of this there -- so wontfix yet)

source/api/presentation/3.0/index.md
azaroth42: It's the extent vs nav structure case.  This allows boring content to be part of a sub-range without polluting a nice clean ToC with it.  Will try to explain succinctly.

PR #1393, opened 2018-02-09
h/w to MAY to not prejudice against pure audio

### LINKED ISSUE #1361: Consistency of normative statements for image and non-image resources
The _SHOULDs_ and _MAYs_ on `height`, `width` and `duration` for Canvas (http://prezi3.iiif.io/api/presentation/3.0/#height) favour spatial coordinates over temporal (_SHOULD_ have a `width` but _MAY_ have a `duration`).

The normative language here could get very complex... _MUST_ have a height if it doesn't have a duration; _MAY_ have a duration if it has width and height; _MUST_ have duration if no width/height... and so on.

Also. as a sometime front end developer I suggest that dimensions of all content resources are _SHOULDs_ rather than _MAYs_, subject to the same combination clarifications.

azaroth42:
Closes #1361  (despite what the branch name implies)

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -401,10 +401,10 @@ The height of this Canvas or external content resource. For content resources, t
 
 The value _MUST_ be a non-negative integer or floating point number.
 
- * A Canvas _SHOULD_ have exactly one `height`, and _MUST NOT_ have more than one. If it has a `height`, it _MUST_ also have a `width`.<br/>
-   Clients _MUST_ process, and _MAY_ render, `height` on a Canvas.
+ * A Canvas _MAY_ have exactly one `height`, and _MUST NOT_ have more than one. If it has a `height`, it _MUST_ also have a `width`.<br/>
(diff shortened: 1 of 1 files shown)


zimeon:
I don't think I agree with the change from *render* to *explicitly display* for a couple of reasons:

  1. what does *explicitly* add to *display*? (Displaying something seems pretty explicit)
  2. why is *display* better than *render*? (To my ear, "render an audio track" makes more sense that "display an audio track")

Am I missing something?

azaroth42:
@zimeon I can back that out, but the thought was that you need to process the dimension (e.g. create a canvas of the correct width) but you don't need to explicitly render the value of the dimension to the user. Just remove the render/display bits completely?

tomcrane:
If we do keep the render/display bit, I think the text should be 

> ... _MAY_ display the value of ...

However, I don't think the spec needs to mention this at all; if a client feels the need to do this (perhaps it's a content creation tool, not a viewer) then it will just do it.

azaroth42:
Fixed. Reviewable again now

PR #1404, opened 2018-02-14
define embedded,referenced, link from first instances ; closes #1396

### LINKED ISSUE #1396: Consistency: be clear about what "embedded" means
@tomcrane:
>How do we make it clear that if a resource is "embedded inline", we mean that the full JSON of the resource has been included, not some reduced version? Variations on "embedded" appear quite a few times in the spec, do you think we should define this formally? In Terminology?

Define "embedded" and "referenced" in Terminology?

azaroth42:
Closes #1396

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -84,6 +84,9 @@ Collectively the use cases require a model in which one can characterize the obj
 
 The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _SHOULD NOT_, _RECOMMENDED_, _MAY_, and _OPTIONAL_ in this document are to be interpreted as described in [RFC 2119][rfc-2119].
 
+* __embedded__ When a resource is embedded within another resource, the complete representation of the embedded resource is present within the embedding resource, and dereferencing the URI of the embedded resource will not result in additional important information. Example: A Sequence is embedded in a Manifest.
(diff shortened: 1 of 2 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -84,6 +84,9 @@ Collectively the use cases require a model in which one can characterize the obj
 
 The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _SHOULD NOT_, _RECOMMENDED_, _MAY_, and _OPTIONAL_ in this document are to be interpreted as described in [RFC 2119][rfc-2119].
 
+* __embedded__ When a resource is embedded within another resource, the complete representation of the embedded resource is present within the embedding resource, and dereferencing the URI of the embedded resource will not result in additional important information. Example: A Sequence is embedded in a Manifest.
```
zimeon: Two issues: the bullets seems odd without introduction, and if used there needs to be punctuation between the word and the sentence. I wonder whether it might be cleaner to use a paragraph, say:

The words __embedded__ and __referenced__ are to be interpreted as follows. When a resource is __embedded__ within another resource, the complete representation of the embedded resource is present within the embedding resource, and dereferencing the URI of the embedded resource will not result in additional important information. Example: A Sequence is embedded in a Manifest. When a resource is __referenced__ from another resource, an incomplete representation of the referenced resource is present within the referencing resource, and dereferencing the URI of the referenced resource will result in additional information.  Typically, the `id`, `type`, and `label` properties will be included in the reference. Example:  A Manifest is referenced in a Collection.

azaroth42: I wanted to do a definition list, but <dl> is only available as HTML not in markdown.  I think a paragraph of text will just get ignored, and I'd rather call out these definitions as obviously as possible, given they're not in a place most people read.

How about...

> This specification also uses the following terms:
> * __embedded__: ...
> * __referenced__: ...

?

PR #1406, opened 2018-02-17
Remove paging, closes #1343

### LINKED ISSUE #1343: Deprecate the paging model
In the Presentation API v2.1, there are two resources that allow paging: AnnotationLists and Collections.  In v3.0, AnnotationLists are pushed to the equivalent construction in the Web Annotation Model, an Annotation Collection Page that follows exactly the same model (Activity Streams 2.0 Ordered Paged Collection). This pattern was not available in the Open Annotation model, and its inclusion in the W3C Technical Recommendation drew from the use in IIIF amongst other sources. As such, we no longer need to specify the paging model for that construction separately.

Collections already allow, in the same way as Ranges, a recursive inclusion to arbitrary depth.  The semantics of a Collection are simply:
> Collections allow easy advertising and browsing of the manifests in a hierarchical structure, potentially with its own descriptive information.

Which allows already for paging as a mechanism for browsing the set of manifests -- by including page-like collections within a master collection, with an appropriate label. Collections and Manifests in 2.1 (and later) may also be interleaved, unlike with the strict paging model where the page is a transmission artifact rather than a persistent entity.

Given that the use cases that necessitated the introduction of the paging section appear to be met, and the Collection Pages are inconsistent with the ActivityStreams pages (in that they are Collections, not CollectionPages), I propose to deprecate the paging model in Presentation API 3.0

azaroth42:
Closes #1343

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -641,9 +641,7 @@ A link to another resource that contains this resource, such as a Manifest that 
 
 The value _MUST_ be an array of JSON objects.  Each object _MUST_ have the `id` and `type` properties, and _SHOULD_ have the `label` property.
 
- * Collections or Annotation Pages that serve as [pages][prezi-api-3-paging] _MUST_ be within exactly one paged resource.<br/>
-   Clients _SHOULD_ render `within` on a Collection or Annotation Page.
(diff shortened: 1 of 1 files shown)


jpstroop:
Do we need a note somewhere about how we will do paging?

PR #1408, opened 2018-02-18
service uses target doc version, collections collect current doc version

azaroth42:
This will, I think, close out all of the cross version issues.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -577,7 +577,7 @@ The value _MUST_ be an array of JSON objects. Each object _MUST_ have the `id`, 
 ##### service
 A link to an external service that the client might interact with directly and gain additional information or functionality for using this resource, such as from an image to the base URI of an associated [IIIF Image API][image-api] service. The service resource _SHOULD_ have additional information associated with it in order to allow the client to determine how to make appropriate use of it. Please see the [Service Registry][registry-services] document for the details of currently known service types.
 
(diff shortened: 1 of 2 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -950,11 +955,11 @@ The value of the `@context` property _MUST_ be a list, and the __last__ two valu
 }
 ```
 
-Any additional fields beyond those defined in this specification or the Web Annotation Data Model _SHOULD_ be mapped to RDF predicates using further context documents.   If possible, these extensions _SHOULD_ be added to the top level `@context` field, and _MUST_ be added before the above contexts.  The JSON-LD 1.1 functionality of type and predicate specific context definitions _SHOULD_ be used if possible to try to minimize any cross-extension collisions.
+Any additional fields beyond those defined in this specification or the Web Annotation Data Model _SHOULD_ be mapped to RDF predicates using further context documents.   These extensions _SHOULD_ be added to the top level `@context` field, and _MUST_ be added before the above contexts.  The JSON-LD 1.1 functionality of [type and predicate specific context definitions][json-ld-scoped-contexts] _MUST_ be used to minimize cross-extension collisions.
```
jpstroop: > The JSON-LD 1.1 functionality of [type and predicate specific context definitions][json-ld-scoped-contexts] _MUST_ be used to minimize cross-extension collisions.

Maybe refer to the term "scoped contexts" since this is the name of the section to which you are linking, and that terminology is used in the next para?

PR #1419, opened 2018-02-20
make square req at 1

### LINKED ISSUE #501: Require support for square size at level 1 in Image3.0
Can't require it during 2.x due to conflict with minor version differences in compliance levels and only having a major version differentiator in the URI.  So delay requirement until next major version.

See Sem Ver: #470 and Square: #425

azaroth42:
closes #501


jpstroop:
Rebase against `prezi3` branch after https://github.com/IIIF/api/pull/1422 is merged to fix.

PR #1420, opened 2018-02-20
id,type

### LINKED ISSUE #590: Use id/type instead of @id/@type
Schema.org, Social Web / ActivityStreams, Annotation WG will all adopt this.
Manu, danbri, tantek, etc all agree.

There will be a best practices document published from W3C that states this

azaroth42:
Closes #590


jpstroop:
Rebase against `prezi3` branch after https://github.com/IIIF/api/pull/1422 is merged to fix.

Issue #1274, opened 2017-09-30
Image API servers should be required to support OPTIONS method


christopher-johnson:
many Image API server implementations do not provide OPTIONS.  
The [CORS specification](https://www.w3.org/TR/cors/) indicates that OPTIONS is required for requests with preflight. 

This should be a MUST for all image API servers. This is not indicated in the 2.1.1 spec 
http://iiif.io/api/image/2.1/#image-information-request

azaroth42:
We already link to the CORS specification in that section.  
Are you trying to use an html5 canvas to render the images? Or what situation is causing a failure due to the lack of support in implementations for OPTIONS?

christopher-johnson:
There is no way for a client to know what access control methods a server supports without a preflight. 
It is a client bug, imo, if it simply allows the browser to throw errors for xhrs on servers that do not support what it wants.  This is demonstrated at http://projectmirador.org/demo/.

An xhr for content-type `application/json` or `application/ld+json` is **not** a "simple request", and technically MUST always require a browser preflight, (even though `jquery` allows it).   Angular.js xhr for example will trigger an OPTIONS preflight for content-type `application/json`.  

It is correct by the specifications to require a IIIF server serving a `application/ld+json` to support OPTIONS.  

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS
[2] https://fetch.spec.whatwg.org/#cors-safelisted-request-header

azaroth42:
Duh, for info.json which is what you linked to. Sorry, yes, implementations need to support OPTIONS on info.json for it to work.

:+1: to requiring it.

tomcrane:
Also it should be called out in the Auth API and impl notes. It's kind of implicit in "support CORS" - you can't possibly make auth work without supporting OPTIONS, and there are several examples of it in the wild*, but the spec expects you to join the dots yourself about permitted HTTP verbs.

* e.g., https://github.com/digirati-co-uk/iiif-auth-server/blob/cbc803e55721f6ef71374bc2d4166bbfb39b1ad5/iiifauth/iiif_auth_server.py#L112

azaroth42:
Closed by #1440

christopher-johnson:
@azaroth42 I have observed that if one tries to fetch an image from the getty museum e.g.
```bash
curl -v -X OPTIONS https://data.getty.edu/museum/api/iiif/103470/full/224,300/0/default.jpg
```
[...]

PR #1428, opened 2018-02-20
1090 media type

### LINKED ISSUE #1090: Define JSON-LD profile URI for Image info.json
cf #1066 and azaroth42/museum-lod#7. 

Propose the context document URI.

azaroth42:
Closes #1090


jpstroop:
Should we try to do https://github.com/IIIF/api/issues/1274 in the same PR?

PR #1437, opened 2018-02-20
Update canonical size to w,h

### LINKED ISSUE #1434: Verify canonical use of `w,h` compatible with `sizes` and remove consistency warning
Image API 2.1 has:

> Warning There is an inconsistency between the specification of the sizes list and the canonical URI syntax. Clients should use the Canonical URI Syntax when making image requests based on entries in sizes. For maximum compatibility, servers should support both the w, and w,h forms of the size parameter for values in sizes that maintain the aspect ratio. This inconsistency will be addressed in the next major version of this specification.

This should be resolved and removed in 3.0. See #544 & #607

zimeon:
Fixes #1434

### DIFF
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -424,8 +424,8 @@ In order to support the above requirements, clients _SHOULD_ construct the image
 
 | Parameter | Canonical value |
 | --------- | --------------- |
-| region    | "full" if the whole image is requested, (including a "square" region of a square image)<br/>otherwise the `x,y,w,h` syntax. |
-| size      | "max" if the default size is requested,<br/>the `w,` syntax for images that should be scaled maintaining the aspect ratio,<br/>and the `w,h` syntax for explicit sizes that change the aspect ratio. |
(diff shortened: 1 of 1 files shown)


source/api/image/3.0/index.md
```diff
@@ -424,8 +424,8 @@ In order to support the above requirements, clients _SHOULD_ construct the image
 
 | Parameter | Canonical value |
 | --------- | --------------- |
-| region    | "full" if the whole image is requested, (including a "square" region of a square image)<br/>otherwise the `x,y,w,h` syntax. |
-| size      | "max" if the default size is requested,<br/>the `w,` syntax for images that should be scaled maintaining the aspect ratio,<br/>and the `w,h` syntax for explicit sizes that change the aspect ratio. |
+| region    | "full" if the whole image is requested<br/>otherwise the `x,y,w,h` syntax. |
+| size      | "max" if the default size is requested,<br/>otherwise the `w,h` syntax. |
```
tomcrane: Is it accurate to say that `max` is the default size?

zimeon: agree, should say "if the maximum size is requested"

PR #1442, opened 2018-02-20
Fix acknowledgements to members of the IIIF community

### LINKED ISSUE #1441: Acknowledgements: s/members of the IIIF/members of the IIIF community/
In both image and presentation 2.1 we have the acknowledgement:

> Many thanks to the members of the IIIF for their continuous engagement, innovative ideas and feedback.

which seems a bit off because IIIF is a framework and we don't mean just IIIF-C members. I propose instead:

> Many thanks to the members of the IIIF community for their continuous engagement, innovative ideas and feedback.

zimeon:
Fixes #1441 , also add dummy date line for image


jpstroop:
Non-normative. Merging.

Issue #1373, opened 2018-01-31
info.json does not need to list features/values that are implied by the level


jpstroop:
Follow on from #1275 

In, e.g.:

```
  "qualities": ["default", "color", "gray", "bitonal"],
  "formats": ["jpg", "png", "tiff", "webp"],
  "features": ["sizeAboveFull", "arbitraryRotation"]
```

`default` quality and `jpg` format _never_ need to be listed because they are required at level 0. Level 2 doesn't need to list `png` format, etc. 

Careful w/ `qualities` though, as those may vary by image (beyond `default`) at any level of compliance

zimeon:
I agree that we should not need to specify anything already in profile, but I also think it should not be an error to repeat something in the profile.

azaroth42:
:+1: It would also not be an error to have the same string listed twice in a given property... just meaningless repetition.

eg:  
```
  "extraFormats": ["tif", "tif", "tif"]
```

azaroth42:
Agree close, fixed.

tomcrane:
:+1:

PR #1446, opened 2018-02-21
Simplify profile, reshuffle info.json docs

### LINKED ISSUE #1275: Simplify and align profile property in Image 3
From #520 and #1013, the `profile` property in the info.json document for the Image API has some complications:
  * It currently requires an order, but isn't modeled as an rdf:List at the RDF level.
  * If we make it a list, then it would be a different predicate compared to other uses of `profile`
  * It requires a mixed datatype value of first a string, and then an object.  This is contrary to the simple as possible principle, and the agreement for Presentation to always use arrays (#1131)
  * The merge of profile URI and additional functionality description in 2.x was for a good semantic reason (the extra profile extends the base profile) and allows for community specific profiles to be given URIs and asserted in the document ... but no one has even flinched in this direction to our knowledge.  It's currently, thus, over-engineered, even if it is implemented.

Design constraints for how to deal with profile and additional functionality in Image 3.0:
  * Consistent with other APIs' usage of `profile` (single URI?)
  * Be round-trippable through RDF [re #1013]
  * Retain the capabilities of 2.1.1

Proposal:
  * `profile` is a single URI, and SHOULD be from the list of iiif URIs.
  * It SHOULD dereference to a description of the functionality
  * Add a new property that holds the extended information, as per 2.x  (rather than going back to 1.x where it was all at the top level)

azaroth42:
Closes #1275


source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
 
 ### 5.2. Technical Properties
 
-| Technical Property   | Required? | Description |
+The JSON response has several technical properties described in the table below that describe the available functionality for the image content.
```
zimeon: would remove "described in the table below " because that will be obvious from layout, and "described...that describe" is ponderous to read

jpstroop: 👍

source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
-| `tiles` | Optional | A set of descriptions of the parameters to use to request regions of the image (tiles) that are efficient for the server to deliver. Each description gives a width, optionally a height for non-square tiles, and a set of scale factors at which tiles of those dimensions are available. |
+| `maxWidth`  | Optional  | The maximum width in pixels supported for this image. Requests for images sizes with width greater than this may not be supported. _MUST_ be specified if `maxHeight` is specified. |
+| `maxHeight` | Optional  | The maximum height in pixels supported for this image. Requests for images sizes with height greater than this may not be supported. If `maxWidth` is specified and `maxHeight` is not, then clients should infer that `maxHeight = maxWidth`.  |
+| `maxArea`   | Optional  | The maximum area in pixels supported for this image. Requests for images sizes with width\*height greater than this may not be supported. |
```
zimeon: is this reference to `width` and `height` values... so backtick?

source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
+{
+  "@context" : "http://iiif.io/api/image/{{ page.major }}/context.json",
+  "id": "http://www.example.org/image-service/abcd1234/1E34750D-38DB-4825-A38A-B60A345E591C",
+  "type": "ImageService3",
+  "protocol": "http://iiif.io/api/image",
+  "width": 6000,
+  "height": 4000,
+  "maxHeight": 2000,
+  "maxWidth": 3000,
+  "maxArea": 6000000
```
zimeon: Should make `maxArea` less than `maxWidth * maxHeight` else it will never come into effect -- use 4000000, say.

source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
 | `scaleFactors` | Required | The set of resolution scaling factors for the image's predefined tiles, expressed as positive integers by which to divide the full size of the image. For example, a scale factor of 4 indicates that the service can efficiently deliver images at 1/4 or 25% of the height and width of the full image. A particular scale factor value _SHOULD_ appear only once in the `tiles` list. |
 | `width` | Required | The width in pixels of the predefined tiles to be requested, given as an integer. |
 | `height` | Optional | The height in pixels of the predefined tiles to be requested, given as an integer. If it is not specified in the JSON, then it defaults to the same as `width`, resulting in square tiles. |
 {: .api-table}
 
-Servers _SHOULD_ support requests for images with parameters specified by the `sizes` and `tiles` fields for all combinations of qualities and formats supported.
-
-The following shows a valid image information response, including the optional `sizes` and `tiles` properties.
+The `width` of a tile, or the combination of `width` and `height` if `height` is specified, _MUST_ be unique among the members of the `tiles` list.
```
zimeon: I suggest: "Objects in the `tiles` list _MUST_ each have a unique combination of `width` and `height`, where `height` = `width` if not explicitly specified."

source/api/image/3.0/index.md
jpstroop: Do you mean in, e.g.:
> The maximum width in pixels...
?

I don't think that should be in backticks

source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
-| `maxWidth`  | Optional  | The maximum width in pixels supported for this image. Requests for images sizes with width greater than this may not be supported. _MUST_ be specified if `maxHeight` is specified. |
-| `qualities` | Optional  | The set of image quality parameter values available for the image.  If not specified then clients should assume only qualities declared in the compliance level document.|
-| `supports`  | Optional  | The set of features supported for the image.  If not specified then clients should assume only features declared in the compliance level document. |
+| Property   | Required? | Description |
+| ---------- | --------- | ----------- |
+| `profile` | Required | A string indicating the highest [compliance level][compliance-levels] which is fully supported by the service. The value _MUST_ be one of "level0", "level1", "level2". |
+| `extraQualities` | Optional | A list of strings that can be used in the quality parameter slot, in addition to the ones specified in the referenced profile. |
```
zimeon: should not introduce "slot" -- suggest s/used in the quality parameter slot/used as the quality parameter/ (and in following 2 lines)

source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
+| `extraFeatures` | Optional | A list of strings identifying features supported by the service, in addition to the ones specified in the referenced profile. These strings are defined either in the [table][features-table] below or by [registering an extension][image-extensions]. |
 {: .api-table}
 
-The `maxWidth`, `maxHeight` and `maxArea` parameters provide a way for image servers to express limits on the sizes supported for the image. If `maxWidth` alone, or `maxWidth` and `maxHeight` are specified then clients should expect requests with larger linear dimensions to be rejected. If `maxArea` is specified then clients should expect requests with larger pixel areas to be rejected. The `maxWidth / maxHeight`  and `maxArea` parameters are independent, servers may implement either or both limits. Servers _MUST_ ensure that sizes specified by any `sizes` or `tiles` properties are within any size limits expressed. Clients _SHOULD NOT_ make requests that exceed size limits expressed.
-
-The set of features that may be specified in the `supports` property of an Image profile are:
+The set of features that may be specified in the `extraFeatures` property are:
```
zimeon: --> The set of features **defined by this specification that may be included** in the `extraFeatures` property are:

source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
+The `maxWidth`, `maxHeight` and `maxArea` parameters provide a way for image servers to express limits on the sizes supported for the image. If `maxWidth` alone, or `maxWidth` and `maxHeight` are specified then clients should expect requests with larger linear dimensions to be rejected. If `maxArea` is specified then clients should expect requests with larger pixel areas to be rejected. The `maxWidth / maxHeight`  and `maxArea` parameters are independent, servers may implement either or both limits. Servers _MUST_ ensure that sizes specified by any `sizes` or `tiles` properties are within any size limits expressed. Clients _SHOULD NOT_ make requests that exceed size limits expressed. The `width` and `height` properties are still required in order to construct tile requests and know the aspect ratio of the image.
+
+``` json-doc
```
zimeon: Needs `profile` to be valid -- include it or ellipsis comment, I suggest the `profile`

source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
 
-| Size Object Property | Required? | Description |
+| Property   | Required? | Description |
 | ---------- | -------- | ----------- |
 | `type`     | Optional | The type of the object. If present, the value _MUST_ be the string `Size`. |
 | `width`    | Required | The width in pixels of the image to be requested, given as an integer. |
 | `height`   | Required | The height in pixels of the image to be requested, given as an integer. |
 {: .api-table}
 
+``` json-doc
```
zimeon: Needs `profile` to be valid -- include it or ellipsis comment, I suggest the `profile`

source/api/image/3.0/index.md
```diff
@@ -477,92 +477,126 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
 | `width` | Required | The width in pixels of the predefined tiles to be requested, given as an integer. |
 | `height` | Optional | The height in pixels of the predefined tiles to be requested, given as an integer. If it is not specified in the JSON, then it defaults to the same as `width`, resulting in square tiles. |
 {: .api-table}
 
-Servers _SHOULD_ support requests for images with parameters specified by the `sizes` and `tiles` fields for all combinations of qualities and formats supported.
-
-The following shows a valid image information response, including the optional `sizes` and `tiles` properties.
+The `width` of a tile, or the combination of `width` and `height` if `height` is specified, _MUST_ be unique among the members of the `tiles` list.
 
 ``` json-doc
```
zimeon: Needs `profile` to be valid -- include it or ellipsis comment, I suggest the `profile`

zimeon: Needs `profile` to be valid -- include it or ellipsis comment, I suggest the `profile`

source/api/image/3.0/index.md
zimeon: I meant in maxArea def'n, but I think it doesn't apply. Pls ignore

source/api/image/3.0/index.md
azaroth42: Applies to all examples, right?  Should profile be defined in the first section rather than needing to read forwards to find out what it is?

source/api/image/3.0/index.md
```diff
@@ -477,143 +477,116 @@ Servers should also support [CORS][cors-response] on Image Information responses
...
-| `tiles` | Optional | A set of descriptions of the parameters to use to request regions of the image (tiles) that are efficient for the server to deliver. Each description gives a width, optionally a height for non-square tiles, and a set of scale factors at which tiles of those dimensions are available. |
+| `maxWidth`  | Optional  | The maximum width in pixels supported for this image. Requests for images sizes with width greater than this may not be supported. _MUST_ be specified if `maxHeight` is specified. |
+| `maxHeight` | Optional  | The maximum height in pixels supported for this image. Requests for images sizes with height greater than this may not be supported. If `maxWidth` is specified and `maxHeight` is not, then clients should infer that `maxHeight = maxWidth`.  |
+| `maxArea`   | Optional  | The maximum area in pixels supported for this image. Requests for images sizes with width\*height greater than this may not be supported. |
```
jpstroop: > images sizes with width\*height greater than this may not be supported?

width\*height ?

jpstroop: Oh, escaping?

azaroth42: Yep, escaped to avoid it trying to parse as a bold with no end

source/api/image/3.0/index.md
```diff
@@ -652,40 +622,75 @@ The following shows a simple use of each of these properties:
...
+
+The JSON response also contains properties that describe additional functionality available via the Image API service.
 
 | Property   | Required? | Description |
 | ---------- | --------- | ----------- |
-| `service`  | Optional  | The `service` property provides hooks for additional information to be included in the image description, for example a link to an authentication service. The value _MUST_ be a list of objects. |
+| `extraQualities` | Optional | A list of strings that can be used as the quality parameter, in addition to the ones specified in the referenced profile. |
+| `extraFormats` | Optional | A list of strings that can be used as the format parameter, in addition to the ones specified in the referenced profile. |
+| `extraFeatures` | Optional | A list of strings identifying features supported by the service, in addition to the ones specified in the referenced profile. These strings are defined either in the [table][features-table] below or by [registering an extension][image-extensions]. |
 {: .api-table}
```
jpstroop: Probably OK, but we may want to revisit these when we look at https://github.com/IIIF/api/issues/1373

source/api/image/3.0/index.md
jpstroop: Oh, and:

> may not be supported

do we really mean that they _might_ be? Why can't we say _MUST_ not be supported?

azaroth42:
Sorry, forgot that changing section titles would mess up #fragment refs.  Fixed (I think!)

source/api/image/3.0/index.md
azaroth42: True. I think they're not supported. Will fix.

azaroth42:
I agree with jpstroop's comment that max should be explicit limits. Otherwise they're not valuable, as clients will just try to make requests to see if they work, rather than knowing that they won't

zimeon:
The are you arguing for "_MUST_ not be supported"?

jpstroop:
> Then are you arguing for "MUST not be supported"?

I am.

azaroth42:
It's advertising an implementation, or even image, specific limit. Which in effect is "MUST NOT" be supported, but is a meaningless distinction to me as "You MUST NOT support something that you don't support" is redundant.

zimeon:
I feel uncomfortable constraining implementations in this way, it seems to be stepping over the line of "define success, not failure". To me, the essence of our max* parameters is that clients should not expect to be able to make requests for sizes larger than this. If some implementation wanted to support an extension called `specialSizesOverMax` then that would seem OK to me and point to a looser definition of  expectation to reject a request that exceeds the max* sizes.

azaroth42:
How about:

> Clients MUST NOT expect requests with a (width/height/area) above this value to be supported.

PR #1450, opened 2018-02-22
stanford-->getty

azaroth42:


jpstroop:
non-normative and I'm sure @tomcrane doesn't disagree.

PR #1453, opened 2018-02-22
allow auto-advance on range

### LINKED ISSUE #1445: Allow auto-advance on Range
In order to advance within a discontiguous range (e.g. an oral history recording that is split up across tapes, with intervening content), auto-advance would be applied to a Range that includes both sections.  

Clients should then be aware of whether the user is playing the Sequence or the Range and which the "next" segment is.

azaroth42:
Closes #1445


jpstroop:
👎

Issue #1387, opened 2018-02-08
Addressing points rather than extents


tomcrane:
IIIF (via OA/W3C annotations) uses the [W3C Media Fragments specification](https://www.w3.org/TR/media-frags/) to target spatial and temporal extents of the canvas:

`xywh=500,500,100,100`

`t=10,20`

A common annotation use case is map markers or other point annotations. A common solution to this today is a target that looks like `500,500,1,1`. For time-based media, points might be used to indicate a cut in a film, or where to start playing a clip from.

The media fragments spec has no support for a point, e.g., `xy=500,500`. The syntax does allow for `t=20`, but says that it should be interpreted as `t=20,end`. We don't want the spatial equivalent of that, where our `xy` target is interpreted as "extend to the bottom right corner".

What should IIIF do about point targeting? Carry on using `500,500,1,1`? If so, this pattern should be documented. If not, something more elegant, to support `xy=500,500`, should go in the spec. We would then need to define what `t=20` means.

zimeon:
If we do decide we really need a point then I think the clearest choices are either:

  1. use illegal zero-size fragments `t=10,10`, ’xywh=10,20,0,0` that otherwise follow the media fragments spec; or
  2. invent new point types along the lines of `tp=10`, `xy=10,20` that follow the start/top-left semantics from the media fragments spec

I do not think that redefining the meaning of `t=10` is a viable option because that would be a conflict with valid semantics in media fragments

workergnome:
I don't see a use for regions of time with zero duration, nor regions of space with zero volume.  Even a map marker occurs at a single pixel, not the top-left corner of a pixel, and a frame of video has duration, even if it's just the duration of a single frame.  

Points (and instants) are really nasty things that are also imaginary.

tomcrane:
Being imaginary doesn't stop us using latitude and longitude coordinates to put markers on a map, or talking about music using both instants and extents:

"the trombones enter at 4:14" vs "listen for the cowbells from 7:34-11:01"  

The distinction between targeting a point and targeting a region when annotating spatial or temporal content seems important to me. It's not about precision, it's about intent, and if the intent is clear from unambiguous syntax, a client can produce a different appropriate user experience. Pins rather than polygons, perhaps. Or ▼ on a timeline rather than ｜.

For spatial, `1,1` is currently a de facto magic syntax to indicate a point rather than a rectangle below the resolution of the canvas coordinates. I agree that people could live with that (although sometimes they use 0,0).

[...]

PR #1474, opened 2018-02-25
Airplane Work

azaroth42:
Lots of high level readability changes

Preview:  http://preview.iiif.io/api/airplane/api/presentation/3.0/


source/api/presentation/3.0/index.md
```diff
@@ -110,49 +117,51 @@ The overall description of the structure and properties of the digital represent
...
+##### Annotation Collection
+{: #overview-annotationcollection}
+
+An ordered list of Annotation Pages. Annotation Collections allow higher level groupings of Annotations to be recorded. For example, all of the English translation Annotations of a medieval French document could be kept separate from the transcription or an edition in modern French, or the director's commentary on a film can be separated from the script.
 
 ##### Annotation Page
 {: #overview-annotationpage}
 
-An ordered list of Annotations in a single response, typically associated with a single Canvas, and can be part of an Annotation Collection.
+An ordered list of Annotations, typically associated with a Canvas but may be referenced from other resource types as well.  Annotation Pages are often part of an Annotation Collection.
```
zimeon: Introduces double space...

source/api/presentation/3.0/index.md
```diff
@@ -110,49 +117,51 @@ The overall description of the structure and properties of the digital represent
...
 
 The requirements for which classes have which properties are summarized in [Appendix A][prezi-api-3-appendixa].
 
 Other properties are allowed, either via local extensions or those endorsed by the IIIF community. If a client discovers properties that it does not understand, then it _MUST_ ignore them. See the [Linked Data Context and Extensions][ldce] section for more information about extensions.
 
+This section also defines processing requirements for clients for each of the combinations of class and property.  These requirements are for general purpose client implementations that are intended to be used to render the entire resource to the user, and not necessarily for consuming applications with specialized use or individual component implementations that might be used to construct a client. The inclusion of these requirements gives publishers a baseline expectation for how they can expect implmentations advertised as compliant with this specification to behave when processing their content. 
+
 ###  3.1. Descriptive Properties
 
+These properties describe the resource, and are intended for display directly to the user.
```
zimeon: "These properties are used to describe a resources, ..." perhaps ? "the" leads me to think "which?"

source/api/presentation/3.0/index.md
```diff
@@ -316,50 +319,30 @@ The value _MUST_ be an [`xsd:dateTime` literal][xsd-datetime]. The value _MUST_
...
-}
-```
 
-###  3.3. Technical Properties
+###  3.2. Technical Properties
 
 ##### id
 
-The URI that identifies this resource. It is _RECOMMENDED_ that an HTTPS URI be used for all resources. If this resource is only available embedded (see the [terminology section][prezi-api-3-terminology] for an explanation of "embedded") within another resource, such as a Canvas within a Manifest, then the URI _MAY_ be the URI of the encapsulating resource with a unique fragment on the end. This is not true for Canvases, which _MUST_ have their own URI without a fragment.
+The URI that identifies this resource. It is _RECOMMENDED_ that an HTTPS URI be used for all resources defined by this specification. If this resource is only available embedded (see the [terminology section][prezi-api-3-terminology] for an explanation of "embedded") within another resource, such as a Range within a Manifest, then the URI _MAY_ be the URI of the encapsulating resource with a unique fragment on the end. This is not true for Canvases, which _MUST_ have their own URI without a fragment.
```
zimeon: HTTPS -> HTTP(S)

source/api/presentation/3.0/index.md
```diff
@@ -518,12 +507,13 @@ The value _MUST_ be a string, taken from the table below or a full URI.
 ```
 
 ##### behavior
-A set of user experience features for the client to use when presenting this resource to the user. This specification defines the values specified in the table below. Others may be defined externally, and would be given as a URI.
 
-The value _MUST_ be an array of strings, taken from the table below or a URI.
+A set of user experience features that the publisher of the content would prefer the client to use when presenting this resource. This specification defines the values specified in the table below. Others may be defined externally, and _SHOULD_ be [registered as an extension][registry-home].
```
zimeon: Should we make the distinction between local (not to be registered) and community endorsed (to be registered) extensions here?

source/api/presentation/3.0/index.md
```diff
@@ -592,8 +583,44 @@ The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id`, `t
 }
 ```
 
+##### rights
+
+A link to an external resource that describes the license or rights statement under which this resource may be used. The rationale for the value being a URI and not a human readable text is that typically there is one statement or license for many resources, and the text is too long to be displayed to the user at the same time as the object. If displaying rights information directly to the user is a requirement, or a publisher defined label is needed, then it is _RECOMMENDED_ to include the information using the `requiredStatement` property.
+
+The value _MUST_ be an array of JSON objects, each of which _MUST_ have an `id` and _SHOULD_ have at least one of `type` and `format`.
```
zimeon: This version of `rights`, as opposed to what we had in 2.x as `license`, seems a possible awkward match with the sort of machine readable statement we had imagined (e.g. CC) which likely does support conneg and this isn't well described with e media type etc.

source/api/presentation/3.0/index.md
```diff
@@ -701,9 +730,10 @@ The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id` and
...
 
-The value _MUST_ be a JSON object, which _MUST_ have the `id` and `type` properties.
+A link from this Manifest or Range, to a Canvas, or part of a Canvas, that is contained within it. The reference to part of a Canvas is handled in the same way that Ranges reference parts of Canvases. When processing this relationship, a client _SHOULD_ advance to the specified Canvas, or specified segment of the Canvas, when beginning navigation through the Range. This allows the client to begin with the first Canvas that contains interesting content rather than requiring the user to manually navigate to find it.
+
+The value _MUST_ be a JSON object, which _MUST_ have the `id` and `type` properties.  The object _MUST_ be either a Canvas, or a Specific Resource with a Selector and a `source` property where the value is a Canvas.
```
zimeon: This is quite a mouthful now, do we need an example?

source/api/presentation/3.0/index.md
```diff
@@ -908,11 +935,11 @@ Clients _SHOULD_ allow only `a`, `b`, `br`, `i`, `img`, `p`, `small`, `span`, `s
...
+The top level resource in the response _MUST_ have the `@context` property, and it _SHOULD_ appear as the very first key/value pair of the JSON representation. This tells Linked Data processors how to interpret the document. The IIIF Presentation API context, below, _MUST_ occur once per response and _MUST_ be omitted from any embedded resources. For example, when embedding a Canvas, the Canvas _MUST NOT_ have the `@context` property.
 
-The value of the `@context` property _MUST_ be a list, and the __last__ two values _MUST_ be the Web Annotation context and the Presentation API context, in that order. And further contexts _MUST_ be added at the beginning of the list.
+The value of the `@context` property _MUST_ be an array, and the __last__ two values _MUST_ be the Web Annotation context and the Presentation API context, in that order. Further contexts, such as [registered by extensions][registry-home], _MUST_ be added at the beginning of the array.
```
zimeon: "such as registered by extensions" -> "such as those for registered extensions" ?

source/api/presentation/3.0/index.md
```diff
@@ -908,11 +935,11 @@ Clients _SHOULD_ allow only `a`, `b`, `br`, `i`, `img`, `p`, `small`, `span`, `s
 { "summary": { "en-latn": [ "<p>Short <b>summary</b> of the resource.</p>" ] } }
 ```
 
-### 4.6. Linked Data Context and Extensions
+### 4.5. Linked Data Context and Extensions
 
-The top level resource in the response _MUST_ have the `@context` property, and it _SHOULD_ appear as the very first key/value pair of the JSON representation. This tells Linked Data processors how to interpret the information. The IIIF Presentation API context, below, _MUST_ occur once per response, and be omitted from any embedded resources. For example, when embedding a Canvas, the Canvas _MUST NOT_ have the `@context` field.
+The top level resource in the response _MUST_ have the `@context` property, and it _SHOULD_ appear as the very first key/value pair of the JSON representation. This tells Linked Data processors how to interpret the document. The IIIF Presentation API context, below, _MUST_ occur once per response and _MUST_ be omitted from any embedded resources. For example, when embedding a Canvas, the Canvas _MUST NOT_ have the `@context` property.
```
zimeon: Seems odd to have an example (last sentence) with a normative _MUST NOT_

azaroth42: :+1: fixed.

source/api/presentation/3.0/index.md
azaroth42: :+1: fixed.

source/api/presentation/3.0/index.md
azaroth42: Yes... but how? A second json-doc, I guess, as each resource can only have one `start`?

source/api/presentation/3.0/index.md
azaroth42: I agree ... see the issue on rights. I didn't change any of the text from current prezi3 branch, just moved it.

source/api/presentation/3.0/index.md
```diff
@@ -549,9 +539,9 @@ The value _MUST_ be an array of strings, taken from the table below or a URI.
 
 ##### timeMode
 
-A mode associated with an Annotation that is to be applied to the rendering of any time-based media, or otherwise could be considered to have a duration, used as a body resource of that Annotation. Note that the association of `timeMode` with the Annotation means that different resources in the body cannot have different values. This specification defines the values specified in the table below. Others may be defined externally, and would be given as a full URI.
+A mode associated with an Annotation that is to be applied to the rendering of any time-based media, or otherwise could be considered to have a duration, used as a body resource of that Annotation. Note that the association of `timeMode` with the Annotation means that different resources in the body cannot have different values. This specification defines the values specified in the table below. Others may be defined externally, and _SHOULD_ be [registered as an extension][registry-home].
```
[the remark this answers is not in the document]
azaroth42: How about doing this once in the extensions section?

source/api/presentation/3.0/index.md
azaroth42: :+1: fixed. Though not introduced by this PR.

source/api/presentation/3.0/index.md
azaroth42: :+1: Fixed.

source/api/presentation/3.0/index.md
azaroth42: :P Fixed

source/api/presentation/3.0/index.md
```diff
@@ -110,49 +117,51 @@ The overall description of the structure and properties of the digital represent
...
+##### Annotation Collection
+{: #overview-annotationcollection}
+
+An ordered list of Annotation Pages. Annotation Collections allow higher level groupings of Annotations to be recorded. For example, all of the English translation Annotations of a medieval French document could be kept separate from the transcription or an edition in modern French, or the director's commentary on a film can be separated from the script.
 
 ##### Annotation Page
 {: #overview-annotationpage}
 
-An ordered list of Annotations in a single response, typically associated with a single Canvas, and can be part of an Annotation Collection.
+An ordered list of Annotations, typically associated with a Canvas but may be referenced from other resource types as well. Annotation Pages are often part of an Annotation Collection.
```
mikeapp: The lack of a verb before 'but' bothers me.  "An ordered list of Annotations, it is typically associated with a Canvas but may ..."

source/api/presentation/3.0/index.md
```diff
@@ -110,49 +117,51 @@ The overall description of the structure and properties of the digital represent
...
 
 ##### Annotation
 {: #overview-annotation}
 
-All content is linked to a Canvas via Annotations. The same mechanism is used for the visible or audible resources as is used for transcriptions, commentary, tags and other content. This provides a single, coherent method for aligning information, and provides a standards based framework for distinguishing parts of resources and parts of Canvases. As Annotations can be added later, it promotes a distributed system in which publishers can align their content with the descriptions created by others.
+Content resources are associated with a Canvas via Annotations. The same mechanism is used for the visible and/or audible resources as is used for transcriptions, commentary, tags and other content. This provides a single, coherent method for aligning information, and provides a standards based framework for distinguishing parts of resources and parts of Canvases. As Annotations can be added later, it promotes a distributed system in which publishers can align their content with the descriptions created by others.
```
mikeapp: "standards-based"

source/api/presentation/3.0/index.md
```diff
@@ -728,24 +761,25 @@ The value _MUST_ be a JSON object, which _MUST_ have the `id` and `type` propert
...
-The value _MUST_ be an array of JSON objects. The items will be resources of different types, as described below.
+Much of the functionality of the IIIF Presentation API is simply recording the order in which child resources occur within a parent resource, such as Collections or Manifests within a parent Collection, or Canvases within a Manifest. All of these situations are covered with a single property, `items`. 
+
+The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id` and `type` properties. The items will be resources of different types, as described below.
 
 * A Collection _MUST_ have the `items` property. Each item _MUST_ be either a Collection or a Manifest.<br/>
   Clients _MUST_ process `items` on a Collection.
-* A Manifest _MUST_ the `items` property. Each item _MUST_ be a Canvas.<br/>
+* A Manifest _MUST_ the `items` property with at least one item. Each item _MUST_ be a Canvas.<br/>
```
mikeapp: "have the" [have is missing]

source/api/presentation/3.0/index.md
zimeon: Yes, I think so, Perhaps change current example to be simple canvas:

```
{ "start": { "id": "https://example.org/iiif/1/canvas3", "type": "Canvas" } }
```

and add a second example with `SpecificResource` following #1387:

```
{
  "start": {
    "type": "SpecificResource",
    "source": {
      "id": "https://example.org/iiif/id1/canvas1",
      "type": "Canvas"
    },
    "selector": {
      "type": "PointSelector",
      "conformsTo": "http://iiif.io/api/annex/annotation/pointSelector/",
      "t": 14.5
    }
  }
}
```

source/api/presentation/3.0/index.md
```diff
@@ -923,29 +952,69 @@ The value of the `@context` property _MUST_ be a list, and the __last__ two valu
...
+
+The identifier in `id` _MUST_ be able to be dereferenced to retrieve the JSON description of the Manifest, and thus _MUST_ use the HTTP(S) URI scheme.
+
+The Manifest _MUST_ have an `items` property, which is an array of JSON-LD objects. Each object is a Canvas, described in the next section. It _MAY_ also have a `structures` property listing one or more [Ranges][prezi-api-3-range] which describe additional structure of the content, such as might be rendered as a table of contents.  The Manifest _MAY_ have an `annotations` property, which includes Annotation Page resources where the Annotations have the Manifest as their `target`. These will typically be comment style annotations, and _MUST NOT_ have `painting` as their `motivation`. 
+
+The example below includes only the Manifest-level information, however actual implementations _MUST_ embed at least the Canvases in `items`.
```
zimeon: Delete this sentence perhaps? The example does include `items`

source/api/presentation/3.0/index.md
```diff
@@ -1106,26 +1175,26 @@ The example below includes only the Manifest-level information, however actual i
...
 
-Every Canvas _SHOULD_ have a `label` to display. If one is not provided, the client _SHOULD_ automatically generate one for use based on the Canvas's position within the `items` list.
+Every Canvas _SHOULD_ have a `label` to display. If one is not provided, the client _SHOULD_ automatically generate one for use based on the Canvas's position within the `items` property.
 
 Content resources are associated with the Canvas via Web Annotations. Content that is to be rendered as part of the Canvas _MUST_ be associated by an Annotation with the "painting" `motivation`. These Annotations are recorded in the `items` of one or more Annotation Pages, referred to in the `items` array of the Canvas. Annotations that do not have the "painting" `motivation` MUST NOT be in pages referenced in `items`, but instead in the `annotations` property.
```
zimeon: Not introduced by this PR but could fix MUST NOT to use rfc formatting

source/api/presentation/3.0/index.md
```diff
@@ -1106,26 +1175,26 @@ The example below includes only the Manifest-level information, however actual i
...
+Content that is derived from the Canvas, such as a transcription of text in an image or the words spoken in an audio representation, _MUST_ be associated by an Annotation with the "transcribing" `motivation`. Annotations _MAY_ have any other `motivation` as well. Thus content of any type may be associated with the Canvas via a "painting" annotation meaning the content is __part of__ the Canvas, a "transcribing" annotation meaning the content is __from__ the Canvas but not necessarily part of it, or an Annotation with another `motivation` meaning that it is somehow __about__ the Canvas.
```
zimeon: Inconsistent formatting of __part of__, __from__ and __about__ -- we don't have a practice of bolding, I think this would be fine without the emphasis

source/api/presentation/3.0/index.md
```diff
@@ -1298,81 +1292,117 @@ Ranges _MAY_ link to an Annotation Collection that has the content of the Range
...
 }
 ```
 
-### 5.7. Collection
+### 5.6. Annotations
+
+Annotations follow the [Web Annotation][webanno] data model. The description provided here is a summary plus any IIIF specific requirements. It must be noted that the W3C standard is the official documentation.
 
-Collections are used to list the Manifests available for viewing, and to describe the structures, hierarchies or sets that the resources are part of. Collections _MAY_ include both other Collections and Manifests, in order to form a tree-structured hierarchy. Collections might be used to model dynamic result sets from a search, fixed sets of related resources, or other groupings of Manifests for presentation to the user, typically for navigation amongst the member items.
+Note that the Web Annotation data model defines different patterns for the `value` property, when used within an Annotation. The value of a Textual Body or a Fragment Selector, for example, are strings rather than JSON objects with languages and values. Care _MUST_ be taken to use the correct string form in these cases.
```
zimeon: I don't think it make sense to use RFC _MUST_ when talking about care, suggest changing to plain must

zimeon:
Modulo resolution of comments, I'm :+1: on this PR

source/api/presentation/3.0/index.md
azaroth42: Propose to do this in whatever PR fixes #1387, not this one.

source/api/presentation/3.0/index.md
azaroth42: :+1: Fixed.

source/api/presentation/3.0/index.md
azaroth42: :+1: Fixed.

source/api/presentation/3.0/index.md
azaroth42: :+1: fixed.

source/api/presentation/3.0/index.md
azaroth42: :+1: Yup, fixed.

source/api/presentation/3.0/index.md
azaroth42: Sure. I wanted to draw attention to the differences, but the repeated linguistic patterns make them pretty obvious. Fixed.

source/api/presentation/3.0/index.md
azaroth42: :+1: Fixed.

source/api/presentation/3.0/index.md
zimeon: :+1:

source/api/presentation/3.0/index.md
```diff
@@ -1298,81 +1290,117 @@ Ranges _MAY_ link to an Annotation Collection that has the content of the Range
...
-    }
-  ]
+
+### 5.7. Annotation Collection
+
+Annotation Collections represent groupings of Annotation Pages that should be managed as a single whole, regardless of which Canvas or resource they target. This allows, for example, all of the Annotations that make up a particular translation of the text of a book to be collected together. A client might then present a user interface that allows all of the Annotations in an Annotation Collection to be displayed or hidden according to the user's preference.
+
+Note that the Web Annotation Data Model uses the JSON-LD 1.0 specification and defines the use of `label` on an Annotation Collection to be one or more strings, rather than a JSON object with languages as keys. Clients _SHOULD_ accept both forms.
+
+Annotation Collections _MUST_ have a URI, and it _SHOULD_ be an HTTP URI. They _MUST_ have a `label` and _MAY_ have any of the other descriptive, linking or rights properties.
```
mikeapp: HTTP(S)

mikeapp:
4.1 and 4.2 now have the same example

mikeapp:
Two comments immediately above, otherwise I'm 👍

source/api/presentation/3.0/index.md
```diff
@@ -728,24 +761,25 @@ The value _MUST_ be a JSON object, which _MUST_ have the `id` and `type` propert
...
+
+The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id` and `type` properties. The items will be resources of different types, as described below.
 
 * A Collection _MUST_ have the `items` property. Each item _MUST_ be either a Collection or a Manifest.<br/>
   Clients _MUST_ process `items` on a Collection.
-* A Manifest _MUST_ the `items` property. Each item _MUST_ be a Canvas.<br/>
+* A Manifest _MUST_ have the `items` property with at least one item. Each item _MUST_ be a Canvas.<br/>
   Clients _MUST_ process `items` on a Manifest.
-* A Canvas _SHOULD_ have the `items` property. Each item _MUST_ be an Annotation Page<br/>
+* A Canvas _SHOULD_ have the `items` property with at least one item. Each item _MUST_ be an Annotation Page<br/>
```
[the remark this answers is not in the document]
tomcrane: Another thought here - probably a non-starter as it may cause more problems than it solves.

We're likely (as with sequence) to have a 99.99..% case where the content is always in `canvas.items[0].items`, especially as only "painting" annos are allowed ("transcribing" go in `annotations`). So,...

Could a Canvas _be_ an Annotation Page, so its `items` property contains annotations directly (like P2 `canvas.images`)?

Pros - another layer removed from the model; more like P2 (or items as canvas.content rather than canvas.pagesofcontent)
Cons - can't partition `canvas.items` into separately labelled pages, with one or more external to the manifest; possible other problems if a Canvas is also an AnnotationPage

What scenarios would result in a very large number of painting annos, that couldn't be embedded?

This is another issue if there are any legs in it.

source/api/presentation/3.0/index.md
```diff
@@ -812,9 +846,9 @@ The value _MUST_ be an array of JSON objects. Each item _MUST_ have at least the
...
 
-Additional motivations may be added to the Annotation to further clarify the intent, drawn from extensions. Known extensions are listed in the [Motivation registry][registry-motivations]. Clients _MUST_ ignore extension motivation values that they do not understand. Other motivation values given in the Web Annotation specification _SHOULD_ be used where appropriate, and examples are given in the [Presentation API Cookbook][cookbook].
+Additional motivations may be added to the Annotation to further clarify the intent, drawn from [extensions][ldce]. Clients _MUST_ ignore motivation values that they do not understand. Other motivation values given in the Web Annotation specification _SHOULD_ be used where appropriate, and examples are given in the [Presentation API Cookbook][cookbook].
```
tomcrane: See my comment on https://github.com/IIIF/api/pull/1487

source/api/presentation/3.0/index.md
```diff
@@ -826,40 +860,40 @@ Additional motivations may be added to the Annotation to further clarify the int
...
+    { "id": "https://example.org/images/thumb1.jpg", "type": "Image" }
   ]
 }
 ```
 
-### 4.4. Language of Property Values
+### 4.3. Language of Property Values
 
-Language _MAY_ be associated with strings that are intended to be displayed to the user for the `label` and `summary` fields, plus the `label` and `value` fields of the `metadata` and `requiredStatement` objects.
+Language _MAY_ be associated with strings that are intended to be displayed to the user for the `label` and `summary` properties, plus the `label` and `value` properties of the `metadata` and `requiredStatement` objects.
```
tomcrane: As we now have a spec where it's hard not to declare a language for a string, should this _MAY_ be a _SHOULD_?

source/api/presentation/3.0/index.md
```diff
@@ -164,11 +173,11 @@ The value of the property _MUST_ be a JSON object, as described in the [language
  * A Canvas _SHOULD_ have the `label` property with at least one entry.<br/>
    Clients _MUST_ render `label` on a Canvas, and _MUST_ generate a `label` for Canvases that do not have them.
```
tomcrane: This is at odds with line 1185 (5.3 Canvas), which says

> Every Canvas _SHOULD_ have a `label` to display. If one is not provided, the client _SHOULD_ automatically generate one for use based on the Canvas's position within the `items` property.

source/api/presentation/3.0/index.md
```diff
@@ -1106,26 +1178,26 @@ The example below includes only the Manifest-level information, however actual i
...
 
-Every Canvas _SHOULD_ have a `label` to display. If one is not provided, the client _SHOULD_ automatically generate one for use based on the Canvas's position within the `items` list.
-
-Content resources are associated with the Canvas via Web Annotations. Content that is to be rendered as part of the Canvas _MUST_ be associated by an Annotation with the "painting" `motivation`. These Annotations are recorded in the `items` of one or more Annotation Pages, referred to in the `items` array of the Canvas. Annotations that do not have the "painting" `motivation` MUST NOT be in pages referenced in `items`, but instead in the `annotations` property.
+Every Canvas _SHOULD_ have a `label` to display. If one is not provided, the client _SHOULD_ automatically generate one for use based on the Canvas's position within the `items` property.
```
tomcrane: Disagrees with client reqts for `label` above. I think it is a SHOULD.

source/api/presentation/3.0/index.md
```diff
@@ -82,23 +82,30 @@ Collectively, the use cases require a model in which one can characterize the ob
 
 ### 1.3. Terminology
 
-The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _SHOULD NOT_, _RECOMMENDED_, _MAY_, and _OPTIONAL_ in this document are to be interpreted as described in [RFC 2119][rfc-2119].
-
-This specification also uses the following terms:
+This specification uses the following terms:
```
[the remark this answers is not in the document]
zimeon: Good point, I think it might be helpful to add another introductory statement saying that we use the JSON-LD terminology which would give that context. Could do something like:

> This specification uses the terminology of the [JSON-LD specification](https://www.w3.org/TR/json-ld/) to describe the serialization of resources. This specification also uses the following terms:

source/api/presentation/3.0/index.md
zimeon: I think _SHOULD_ for generate, it seems too ill-defined for _MUST_

source/api/presentation/3.0/index.md
zimeon: If you didn't know the language (because of some source data where language is not specified, for example) then you probably shouldn't try to associate a language. I could see "_SHOULD_ if known" but I wonder whether _SHOULD_ on its own is too strong?

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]
azaroth42: We don't really use the terminology of JSON-LD, just of JSON (RFC7159).  E.g. in https://json-ld.org/spec/latest/json-ld/#terminology we use the first section, but not really the second LD specific ones other than "context" Happy to add in the 7159 link though.

source/api/presentation/3.0/index.md
```diff
@@ -110,49 +117,51 @@ The overall description of the structure and properties of the digital represent
 
 A virtual container that represents a particular view of the object and has content resources associated with it or with parts of it. The Canvas provides a frame of reference for the layout of the content, both spatially and temporally. The concept of a Canvas is borrowed from standards like PDF and HTML, or applications like Photoshop and PowerPoint, where an initially blank display surface has images, video, text and other content "painted" on to it.
 
-##### Content
-{: #overview-content}
+##### Range
+{: #overview-range}
 
-Content resources such as images, audio, video or text that are associated with a Canvas.
+An ordered list of Canvases, and/or further Ranges. Ranges allow Canvases, or parts thereof, to be grouped together in some way. This could be for content-based reasons, such as might be described in a table of contents or the set of scenes in a play. Equally, physical features might be important such as quires or gatherings, or when recorded music is split across different physical carriers such as two CDs.
```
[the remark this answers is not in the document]
azaroth42: Sure. This wasn't a change introduced by this PR just a move, but will fix.

source/api/presentation/3.0/index.md
```diff
@@ -110,49 +117,51 @@ The overall description of the structure and properties of the digital represent
...
 
 ### 2.2. Additional Types
 
-##### Collection
-{: #overview-collection}
+This specification makes use of the following types, defined elsewhere:
```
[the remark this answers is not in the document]
azaroth42: They used to be that order, but now that Canvases contain Annotation Pages that contain Annotations that reference Content, it seemed to make more sense to keep going down the structural hierarchy.  Then Ranges and Collections are slightly off to one side (as per the diagram)

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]
azaroth42: If it were a page, then it couldn't have other pages... so everything would be one big array of annotations. :-1:

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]
tomcrane: Doesn't need to be JSON-LD specific, just that "embedded" means embedded `{...}` rather than some other notion of embedded.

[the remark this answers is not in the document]
zimeon: Yes, I guess we do mostly use JSON and not JSON-LD so that would be the better link

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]
azaroth42: Worked on this section. I'd prefer this ordering but can change back if the current version isn't acceptable.

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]
azaroth42: Fixed.

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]
azaroth42: Fixed.

source/api/presentation/3.0/index.md
```diff
@@ -316,50 +319,32 @@ The value _MUST_ be an [`xsd:dateTime` literal][xsd-datetime]. The value _MUST_
...
+The URI that identifies this resource. It is _RECOMMENDED_ that an HTTP(S) URI be used for all resources defined by this specification. If this resource is only available embedded (see the [terminology section][prezi-api-3-terminology] for an explanation of "embedded") within another resource, such as a Range within a Manifest, then the URI _MAY_ be the URI of the encapsulating resource with a unique fragment on the end. This is not true for Canvases, which _MUST_ have their own URI without a fragment.
 
-The value _MUST_ be a string.
+The value _MUST_ be a string, and the value _MUST_ be an HTTP(S) URI for resources defined in this specification. If this resource is retrievable via HTTP(S), then the URI _MUST_ be the URI at which it is published.
```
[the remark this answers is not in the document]
azaroth42: Yep. Good catch. Fixed.

source/api/presentation/3.0/index.md
```diff
@@ -316,50 +319,32 @@ The value _MUST_ be an [`xsd:dateTime` literal][xsd-datetime]. The value _MUST_
...
    Clients _MAY_ render `id` on a Range.
- * An Annotation Collection _MUST_ have the `id` property, and the value _MUST_ be an HTTP(S) URI.<br/>
+ * An Annotation Collection _MUST_ have the `id` property.<br/>
    Clients _MAY_ render `id` on an Annotation Collection.
- * An Annotation Page _MUST_ have the `id` property, and the value _MUST_ be the HTTP(S) URI at which it is published.<br/>
+ * An Annotation Page _SHOULD_ have the `id` property.<br/>
    Clients _MAY_ render `id` on an Annotation Page.
- * An Annotation _MUST_ have the `id` property, and the value _MUST_ be an HTTP(S) URI. The Annotation's representation _SHOULD_ be published at that URI.<br/>
+ * An Annotation _MUST_ have the `id` property.<br/>
    Clients _MAY_ render `id` on an Annotation.
```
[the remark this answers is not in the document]
azaroth42: The use case for rendering `id` for Collection, Manifest and Canvas is citation and re-use.  If you can't get the Manifest URI to drop into UV out of the current sub-standard viewer's UI, then it's very hard to find it.  Prefer to leave these as SHOULD.

source/api/presentation/3.0/index.md
```diff
@@ -434,7 +421,7 @@ A schema or named set of functionality available from this resource. The profile
 
 The value _MUST_ be a string, either taken from the table below or a URI.
```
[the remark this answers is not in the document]
azaroth42: Ha! indeed not.  Should point to the service registry.  Fixed.

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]
tomcrane: OK, agreed. Like this: https://github.com/UniversalViewer/universalviewer/issues/439

source/api/presentation/3.0/index.md
```diff
@@ -592,8 +587,43 @@ The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id`, `t
...
+
+A link to an external resource that is an alternative, non-IIIF representation of this resource. Such representations typically cannot be painted onto a single Canvas, as they either include too many views, have incompatible dimensions, or are composite resources requiring additional rendering functionality. The external resource _MUST_ be able to be displayed directly to a human user, although the presentation is likely to be outside of the IIIF client. The external resource _MUST NOT_ have a splash page or other interstitial resource that mediates access to it. If access control is required, then the [IIIF Authentication API][auth-api] is _RECOMMENDED_. Examples include a rendering of a book as a PDF or EPUB, a slide deck with images of a building, or a 3D model of a statue. External resources that are related to this resource _MUST_ instead be referenced in the `metadata` property, or linked to a particular Canvas segment with an Annotation. Any home page or preferred web view of the resource _MUST_ instead be referenced in the `homepage` property.
```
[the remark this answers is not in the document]
azaroth42: :+1: Fixed.

source/api/presentation/3.0/index.md
```diff
@@ -715,9 +747,10 @@ The value _MUST_ be a JSON object, which _MUST_ have the `id` and `type` propert
 ```
 
 ##### includes
-A link from this Range to an Annotation Collection that includes the Annotations of content resources for the Range. Clients might use this to present content to the user from a different Canvas when interacting with the Range, or to jump to the next part of the Range within the same Canvas.  
 
-The value _MUST_ be a JSON object, which _MUST_ have the `id` and `type` properties, and the `type` _MUST_ be `Annotation Collection`.
+A link from this Range to an Annotation Collection that includes the "transcribing" Annotations of content resources for the Range. Clients might use this to present content to the user from a different Canvas when interacting with the Range, or to jump to the next part of the Range within the same Canvas.
```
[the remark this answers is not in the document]
azaroth42: :+1: Fixed.

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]
azaroth42: But fixed missing `.`

source/api/presentation/3.0/index.md
azaroth42: Will fix there.

source/api/presentation/3.0/index.md
azaroth42: I'm happy with _SHOULD_ -- it is the International IIF after all. 
I'm not happy with an unenforcable conditional.  That's just MAY, as you wouldn't want to throw a warning if it was missing, because you can never tell if the "if known" part is in effect or not.  Propose a new issue.

source/api/presentation/3.0/index.md
azaroth42: :+1: to SHOULD generate.

source/api/presentation/3.0/index.md
azaroth42: :+1: Fixed at l173.

source/api/presentation/3.0/index.md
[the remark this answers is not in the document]

source/api/presentation/3.0/index.md
azaroth42: --> https://github.com/IIIF/api/issues/1479

source/api/presentation/3.0/index.md
azaroth42: --> https://github.com/IIIF/api/issues/1492

source/api/presentation/3.0/index.md
```diff
@@ -204,9 +217,10 @@ Clients _SHOULD_ display the pairs in the order provided. Clients _SHOULD NOT_ u
 ```
 
 ##### requiredStatement
-Text that must be displayed when this resource is displayed or used. For example, the `requiredStatement` property could be used to present copyright or ownership statements, an acknowledgement of the owning and/or publishing institution, or any other text that the organization deems critical to display to the user.
 
-Given the wide variation of potential client user interfaces, it will not always be possible to display this statement to the user in the client's initial state. If initially hidden, the method of revealing them _MUST_ be obvious, such as a button or scroll bar.
+Text that _MUST_ be displayed when this resource is displayed or used. For example, the `requiredStatement` property could be used to present copyright or ownership statements, an acknowledgement of the owning and/or publishing institution, or any other text that the organization deems critical to display to the user.
+
+Given the wide variation of potential client user interfaces, it will not always be possible to display this statement to the user in the client's initial state. If initially hidden, the method of revealing it _MUST_ be obvious.
```
jpstroop: > _MUST_ be obvious.

Hmmm...I  don't think we can RFC "obvious". "Clients should make every effort..."

source/api/presentation/3.0/index.md
jpstroop: 👍

source/api/presentation/3.0/index.md
```diff
@@ -47,7 +47,7 @@ This is a work in progress and may change without any notices. Implementers shou
 
 ##  1. Introduction
 
-Access to digital representations of structured resources is a fundamental requirement for many research activities, the transmission of cultural knowledge and for the daily pursuits of every web citizen. Digital content is the primary mode of transmission for access to cultural heritage, science and entertainment. Collections of both digitized physical objects and born-digital content benefit from a standardized description of their structure, layout and presentation mode.
+Access to digital representations of structured resources is a fundamental requirement for many research activities, the transmission of cultural knowledge and for the daily pursuits of every web citizen. Digital content is the primary mode of transmission for access to cultural heritage, science and entertainment. Collections of both digitized physical objects and much born-digital content benefit from a standardized description of their structure, layout and presentation mode.
```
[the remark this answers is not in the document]
azaroth42: Fixed.

source/api/presentation/3.0/index.md
```diff
@@ -82,23 +82,32 @@ Collectively, the use cases require a model in which one can characterize the ob
 
 ### 1.3. Terminology
 
-The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _SHOULD NOT_, _RECOMMENDED_, _MAY_, and _OPTIONAL_ in this document are to be interpreted as described in [RFC 2119][rfc-2119].
+This specification uses the following terms:
+
+* __embedded__: When a resource (A) is embedded within another resource (B), the complete JSON representation of resource A is present within the JSON representation of resource B, and dereferencing the URI of resource A will not result in additional important information. Example: Canvas A is embedded in Manifest B.
```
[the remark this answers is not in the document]
azaroth42: :+1: fixed.

source/api/presentation/3.0/index.md
azaroth42: > If initially hidden, clients _MUST_ make the method of revealing it as obvious as possible.
?

PR #1495, opened 2018-03-06
remake within->partOf changes

azaroth42:


source/api/presentation/3.0/index.md
```diff
@@ -722,16 +722,17 @@ The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id` and
...
+##### partOf
+
+A link to another resource that contains this resource, such as a Manifest that is part of a Collection. When encountering the `partOf` property and the referenced resource is not included in the current representation, clients might retrieve the referenced resource to contribute to the processing of this resource. For example, if a Canvas is encountered as a stand alone resource, it might be part of a Manifest that includes other contextual information, or a Manifest might be part of a Collection, which would aid in navigation.
 
 The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id` and `type` properties, and _SHOULD_ have the `label` property.
 
- * Any resource types _MAY_ have the `within` property with at least one item<br/>
-   Clients _MAY_ render `within` on other resource types.
+ * Any resource types _MAY_ have the `partOf` property with at least one item<br/>
+   Clients _MAY_ render `partOf` on other resource types.
```
zimeon: I don't think "other resources types" makes sense when "Any resource type" can have `partOf`. 

Should it be "Clients _MAY_ render `partOf` links."?

source/api/presentation/3.0/index.md
```diff
@@ -722,16 +722,17 @@ The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id` and
 
 #### 3.4.2. Internal Links
 
-##### within
-A link to another resource that contains this resource, such as a Manifest that is part of a Collection. When encountering the `within` property and the referenced resource is not included in the current representation, clients might retrieve the referenced resource to contribute to the processing of this resource. For example, if a Canvas is encountered as a stand alone resource, it might be `within` a Manifest that includes other contextual information, or a Manifest might be `within` a Collection, which would aid in navigation.
+##### partOf
+
+A link to another resource that contains this resource, such as a Manifest that is part of a Collection. When encountering the `partOf` property and the referenced resource is not included in the current representation, clients might retrieve the referenced resource to contribute to the processing of this resource. For example, if a Canvas is encountered as a stand alone resource, it might be part of a Manifest that includes other contextual information, or a Manifest might be part of a Collection, which would aid in navigation.
```
zimeon: I think the first two sentences are awkward, I suggest starting: "A link to another resource that contains this resource, for example a link from a Manifest to a Collection that it is part of. When a client encounters the `partOf` property, it might retrieve the referenced resource if it is not included in the current representation, in order to contribute to the processing of this resource."

source/api/presentation/3.0/index.md
azaroth42: Fixed. Bad cut/paste error.

zimeon: And I see you fixed the plural that I missed :+1:

Issue #1386, opened 2018-02-08
start with selector rather than a fragment?


azaroth42:
In #1338 we came to consensus that selectors are to be recommended and fragments are allowed ... but the implication on #1320 is that `start` SHOULD thus refer to a `SpecificResource` rather than a `Canvas`... which is not the current pattern.

The value is already an object with id and type rather than a URI, so we've already bitten off that level of verbosity.  Using a selector would avoid the semantic issue of whether uri#fragment is still a Canvas, rather than needing a new type... the type is SpecificResource.

This wouldn't be backwards incompatible, as startCanvas could previously only refer to the entire Canvas. 

#1338 is also inconsistent with #953.

azaroth42:
```
{
  "start": {
    "type": "SpecificResource",
    "source": {
      "id": "https://example.org/iiif/id1/canvas1",
      "type": "Canvas"
    },
    "selector": {
      "type": "PointSelector",
      "conformsTo": "http://iiif.io/api/annex/annotation/pointSelector/",
      "value" : "tp=10"
  }
}
```

azaroth42:
Decision from #1338 (etc) makes this uncontroversial.

azaroth42:
Closed by #1500

Issue #1338, opened 2017-12-03
Fragment selectors or fragment-URI shortcuts when annotating images onto canvas?


zimeon:
In Web Annotation, [fragment selectors](https://www.w3.org/TR/annotation-model/#fragment-selector) are the RECOMMENDED way to reference a fragment. Should this be the IIIF RECOMMENDED way to paint onto a portion of the canvas (and use fragments in annotations in other places)? If so then I'd assume a direct URI-with-fragment-ed would also be allowed (per Web Annnotation) _OR_ Is this all too long winded, we we instead profile Web Annotation to say that the short URI-with-fragment-ed form SHOULD or MUST be used?

Taking a short version of the second example from http://iiif.io/api/presentation/2.1/#segments:

```
{
  "@context": ".../presentation/2/context.json",
  "@id": "http://ex.org/anno1",
  "@type": "oa:Annotation",
  "motivation": "sc:painting",
  "resource": {
    "@id": "http://ex.org/image.jpg",
    ...
   },
  "on": "http://ex.org/canvas/p1#xywh=0,0,600,900"
}
```

we would in Presentation 3 have the following if we follow the RECOMMENDED fragment selector form:

```
{
  "@context": [
    "http://www.w3.org/ns/anno.jsonld",
    ".../presentation/3/context.json"
  ],
  "@id": "http://ex.org/anno1",
  "@type": "Annotation",
  "motivation": "painting",
  "body": {
    "@id": "http://ex.org/image.jpg",
    ...
   },
  "target": {
    "source": "http://ex.org/canvas/p1",
    "selector": {
      "type": "FragmentSelector",
      "conformsTo": "http://www.w3.org/TR/media-frags/",
      "value": "xywh=0,0,600,900"
    }
  }
}
```

azaroth42:
Impacts #1332 as the fragment selector would be more compatible with the proposed AudioTrackSelector as well as the regular SvgSelector.

azaroth42:
Consensus on AV call of 2018-02-06, and Editors on 2018-02-07 was that the selector model should be recommended but not required, and the fragment URI pattern documented as also possible.

azaroth42:
I think this needs further discussion in light of #1386

tomcrane:
Use the full form in examples, explain that the hash-fragment URI form _MAY_ be used. But where in the spec does this explanation go?

aisaac:
>But where in the spec does this explanation go?

In 5.5 Annotations?

[...]

Issue #1332, opened 2017-11-29
how to extract/reference the audio of a video


jronallo:
I'm moving a discussion from another issue onto a new issue. I'll add what I've written below. Other comments begin here on this other issue: https://github.com/IIIF/api/issues/1190#issuecomment-347654763

---

Is there a way to play only the audio from video bitstreams? Or rather what happens if the canvas has only a duration and you paint on video bitstreams/items? 

If you have a duration-only canvas, what happens if you paint on video items that target the whole canvas? Is this different from what happens when you have a duration-only canvas and paint on video items that explicitly target only the duration of the canvas through a media fragment?

The bitstreams/items do not have any indication of their height and width and so could just be treated as audio items. Also in HTML you can make a video the src of an audio element and it plays the audio just fine without displaying the video.

I agree that if you have a duration-only canvas and you try to target the canvas with a media fragment that includes a `xywh` then it should be an error since the canvas does not have any space for painting on the visual component of the video. What I'm presenting is not that case where you have a clear mismatch between explicit dimensions.

This is instead a question of what happens when you intent to reduce a video to just its duration--to just its audio. Use case is that as a manifest creator I want to use the audio from a video without having to extract the audio and publish the audio files myself, especially since I may not have the rights to redistribute the resource myself. Another use case is a video oral history where the video is boring and so the manifest creator only wants to play the audio.

---

The Media Fragments specification (basic) does mention and somewhat define the `track` dimension and give some examples, it just gives no guidance on how to handle this in HTTP or how to process the values. Some of that guidance is in the advanced version but that never got to recommendation status. 

"The way the user agent retrieves information regarding the available tracks of a particular resource is out of scope for this specification." https://www.w3.org/TR/media-frags/#media-fragment-display 
Is there any reason why we couldn't use the `track` name and specify which values a IIIF client ought to accept/reject? 

The basic specification does give guidance on media fragment servers and the case where you're retrieving audio from a video: https://www.w3.org/TR/media-frags/#media-fragment-servers
Since these are in the fragment hash of course you get back a video even if you're just requesting the audio fragment be played. No transformation of the underlying media is suggested.

If leaving off the w,h dimensions from a canvas isn't the way to play audio from a video, then it will probably be easy enough to game this by saying the canvas is 1pixel wide and high which ought to result in a player which looks like an audio player even if it is still trying to render the video. That seems silly. There's no reason to force the client to do the work to render the video component of the video if it isn't really intended to be used.

[...]

PR #1500, opened 2018-03-08
1332 selectors

### LINKED ISSUE #953: Should ranges be able to have non-rectangular regions?
For example to pick out freeform regions on a large artwork, rather than rectangular regions more typical in printed works like newspapers and books.  Currently regions of canvases for Ranges are required to be xywh associated with the Canvas URI as a fragment.

A region might represent, for example, a set of ships in a naval battle painting, the areas where a story is depicted in a scroll, or the set of rondels in a manuscript illumination for a family tree of the royal family of France.

azaroth42:
Closes #953, #1332, #1338, #1386, #1387.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -737,7 +737,7 @@ The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id` and
 
 A link from this Manifest or Range, to a Canvas, or part of a Canvas, that is contained within it. The reference to part of a Canvas is handled in the same way that Ranges reference parts of Canvases. When processing this relationship, a client _SHOULD_ advance to the specified Canvas, or specified segment of the Canvas, when beginning navigation through the Range. This allows the client to begin with the first Canvas that contains interesting content rather than requiring the user to manually navigate to find it.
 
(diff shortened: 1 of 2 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -737,17 +737,31 @@ The value _MUST_ be an array of JSON objects. Each item _MUST_ have the `id` and
 
 A link from this Manifest or Range, to a Canvas, or part of a Canvas, that is contained within it. The reference to part of a Canvas is handled in the same way that Ranges reference parts of Canvases. When processing this relationship, a client _SHOULD_ advance to the specified Canvas, or specified segment of the Canvas, when beginning navigation through the Range. This allows the client to begin with the first Canvas that contains interesting content rather than requiring the user to manually navigate to find it.
 
-The value _MUST_ be a JSON object, which _MUST_ have the `id` and `type` properties.  The object _MUST_ be either a Canvas, or a Specific Resource with a Selector and a `source` property where the value is a Canvas.
+The value _MUST_ be a JSON object, which _MUST_ have the `id` and `type` properties.  The object _MUST_ be either a Canvas (example 1), or a Specific Resource with a Selector and a `source` property where the value is a Canvas (example 2).
```
zimeon: Since the examples aren't numbered, should we say "first example below" and "second example below" instead of "example 1" etc. ?

azaroth42: True. Will fix.

azaroth42:
:clock1: :clock2: :clock3: ...

PR #1504, opened 2018-03-19
1479 rights

### LINKED ISSUE #1464: Consider order of property descriptions
I think we're implicitly ordered by perceived importance -- everything should have a label, so we put label first.

Is this true? Does it align with the Clients MUST/SHOULD/MAY statements?

### LINKED ISSUE #1479: Reconsider `rights` in light of `requiredStatement`
Given 
  * the change in scope from just a license to also include a rights statement, 
  * the inability to associate an internationalized label, per the change to `attribution`,
  * that requiredStatement can use HTML to include links
  * that the URI is not necessarily machine readable / actionable,
  * it's unclear as to the scope of the assertion (just this resource, all included resources, even from other organizations),
  * and it's the _Presentation_ API ...

I think we should consider dropping `rights` in favor of just putting such content into `requiredStatement` as HTML.

azaroth42:
* Create a registry for rights
* Rework rights to be enumerated list
* Add CC and RS.o entries
* Move summary up (#1464)
* Add extraneous pedagogical text to label and summary.

Closes #1479 ; Closes #1464

### DIFF
--- a/source/_includes/links.md
+++ b/source/_includes/links.md
@@ -7,13 +7,6 @@
 [auth-api-1]: {{ site.url }}{{ site.baseurl }}/api/auth/1.0/
 [auth-api]: {{ site.url }}{{ site.baseurl }}/api/auth/
 [cc-licenses]: https://creativecommons.org/licenses/
-[cc-0]: https://creativecommons.org/share-your-work/public-domain/cc0
-[cc-by]: https://creativecommons.org/licenses/by/4.0/
-[cc-by-sa]: https://creativecommons.org/licenses/by-sa/4.0/
-[cc-by-nd]: https://creativecommons.org/licenses/by-nd/4.0/
-[cc-by-nc]: https://creativecommons.org/licenses/by-nc/4.0/
-[cc-by-nc-sa]: https://creativecommons.org/licenses/by-nc-sa/4.0/
-[cc-by-nc-nd]: https://creativecommons.org/licenses/by-nc-nd/4.0/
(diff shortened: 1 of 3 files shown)


source/_includes/links.md
```diff
@@ -6,6 +6,14 @@
 [auth-api-1-token]: {{ site.url }}{{ site.baseurl }}/api/auth/1.0/#access-token-service
 [auth-api-1]: {{ site.url }}{{ site.baseurl }}/api/auth/1.0/
 [auth-api]: {{ site.url }}{{ site.baseurl }}/api/auth/
+[cc-licenses]: https://creativecommons.org/licenses/
+[cc-0]: https://creativecommons.org/share-your-work/public-domain/cc0
```
zimeon: I don't think we need all the cc-XYZ and rs-XYZ refs anymore

source/api/annex/registry/rights/index.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Table of Contents
+{:.no_toc}
+
+* Table of Discontent (will be replaced by macro)
+{:toc}
+
+## 1. Introduction
+
+This is one of a number of [IIIF registries][registry-home]. It lists a set of rights statements and licenses for use across the IIIF APIs that have been identified as useful for implementations.  They may be defined by the IIIF community, or outside of it.
```
zimeon: Should we add in the intro a note that CC and RS statements are already called out in the spec and therefore not listed here?

zimeon: Agree next to table is good and addressed by 2b5b331

PR #1509, opened 2018-03-21
Add paged example

### LINKED ISSUE #1485: Annotation Paging example
We should have an example of next/prev on Annotation Pages.  Notably, Annotation Collections do not have `items` that list all of the pages, by default in the Web Annotation Model.

(They could, as they are AS2 OrderedCollections, ...)

azaroth42:
Closes #1485

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1411,6 +1411,33 @@ Annotation Collections _MUST_ have a URI, and it _SHOULD_ be an HTTP(S) URI. The
 }
 ```
 
+For Annotation Collections with many Annotations, there will be many pages. The Annotation Collection refers to the first and last page, and then the pages refer to the previous and next pages in the ordered list.  Each page is part of the Annotation Collection.
+
+``` json-doc
+{
+  "@context": [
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -1411,6 +1411,33 @@ Annotation Collections _MUST_ have a URI, and it _SHOULD_ be an HTTP(S) URI. The
 }
 ```
 
+For Annotation Collections with many Annotations, there will be many pages. The Annotation Collection refers to the first and last page, and then the pages refer to the previous and next pages in the ordered list.  Each page is part of the Annotation Collection.
```
zimeon: _"pages refer to the previous and next pages in the ordered list"_... is not borne out by the example. Should we say "The example below shows the first page in a list, and thus has a`next` but no `previous` property."?

tomcrane: The example immediately before this shows `AnnotationCollection` with `first` and `last`, then this is an example of one page, so I think it's OK in context.

zimeon: OK, that makes sense

PR #1511, opened 2018-03-24
Moving events to the top level

glenrobson:
This change is to make the API website consistent with the iiif.io website with Events at the top level. Currently they are buried in the community section and see very little use. 

The events page now contains a list of IIIF events people are attending and so should change more frequently. 

This pull request needs to be done close to https://github.com/IIIF/website/pull/71.

### DIFF
--- a/source/_data/navigation.yml
+++ b/source/_data/navigation.yml
@@ -11,5 +11,8 @@
   name: Community
   url: /community/
 -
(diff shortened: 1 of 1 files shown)


glenrobson:
See the following website example: http://preview.iiif.io/root/events-top-level/event/

PR #1513, opened 2018-03-27
Fix floating point description

### LINKED ISSUE #1468: Consider advice/formatting of floating points in Image API spec
We have:

> Size and region parameters in percentages and the rotation parameter _MUST_ be positive floating point numbers or integers. These should have at most 10 decimal digits and consist only of decimal digits and "." with a leading zero if less than 1.0. Intermediate calculations may use floating point numbers and the rounding method is implementation specific.

do we want or need to specify like this? or is there something else to refer to?

zimeon:
Closes #1468

### DIFF
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -142,7 +142,7 @@ The API places no restrictions on the form of the identifiers that a server may 
 
 All parameters described below are required for compliant construction of a IIIF Image API URI. The sequence of parameters in the URI _MUST_ be in the order described below. The order of the parameters is also intended as a mnemonic for the order of the operations by which the service should manipulate the image content. Thus, the requested image content is first extracted as a region of the full image, then scaled to the requested size, mirrored and/or rotated, and finally transformed into the requested color quality and format. This resulting image is returned as the representation for the URI.
 
-Size and region parameters in pixels _MUST_ be non-negative integers. Size and region parameters in percentages and the rotation parameter _MUST_ be positive floating point numbers or integers. See also [Canonical URI Syntax][canonical-uri-syntax].
+Size and region parameters in pixels _MUST_ be non-negative integers. Size and region parameters in percentages and the rotation parameter _MUST_ be positive floating point numbers or integers. For details of the representation of floating point numbers in IIIF URIs, see the [Canonical URI Syntax][canonical-uri-syntax] section.
 
 Servers _SHOULD_ support [CORS][cors-response] on image responses.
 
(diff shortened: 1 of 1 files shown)


zimeon:
Currently there is a travis issue not introduced by this change. Have attempted to address with change of `~0.4.5` to `^0.4.5` in `package.json` but that didn't fix it.

source/api/image/3.0/index.md
```diff
@@ -142,7 +142,7 @@ The API places no restrictions on the form of the identifiers that a server may
...
 
-Size and region parameters in pixels _MUST_ be non-negative integers. Size and region parameters in percentages and the rotation parameter _MUST_ be positive floating point numbers or integers. These should have at most 10 decimal digits and consist only of decimal digits and "." with a leading zero if less than 1.0. Intermediate calculations may use floating point numbers and the rounding method is implementation specific.
+Size and region parameters in pixels _MUST_ be non-negative integers. Size and region parameters in percentages and the rotation parameter _MUST_ be positive floating point numbers or integers. See also [Canonical URI Syntax][canonical-uri-syntax].
```
azaroth42: A little more explanation as to why to see the canonical uri syntax section?

How about:
> For further details on the representation of floating point numbers in IIIF URIs, please see the [Canonical URI Syntax section][canonical-uri-syntax].

source/api/image/3.0/index.md
```diff
@@ -435,11 +435,13 @@ In order to support the above requirements, clients _SHOULD_ construct image req
 | --------- | --------------- |
 | region    | "full" if the full image is requested<br/>otherwise the `x,y,w,h` syntax. |
 | size      | "max" if the maximum size is requested,<br/>otherwise the `w,h` syntax. |
-| rotation  | "!" if the image is mirrored, followed by an integer if possible, and trimming any trailing zeros in a decimal value, and a leading 0 if the value is below 1. |
+| rotation  | "!" if the image is mirrored, followed by an integer if possible, otherwise a floating point value. |
 | quality   | "default" if the server's default quality is requested,<br/>otherwise the quality string. |
 | format    | An explicit format string is always required. |
 {: .api-table}
 
+Size and region parameters given as percentages and the rotation parameter allow positive floating point number values. Integer values _SHOULD_ be use where possible. When floating point values are used, they _SHOULD_ consist only of decimal digits and "." (e.g. 0.9 not +0.9), be represented with a leading 0 if less than 1 (e.g. 0.9 not .9), and not include trailing zeros (e.g. 0.9 not 0.90). Intermediate calculations may use floating point numbers and the rounding method is implementation specific.
```
jpstroop: > _SHOULD_ consist only of decimal digits and a "."

Why not MUST?

zimeon: We could do MUST for the first clause (decimal digits and "."), however then we'd probably want to relax to SHOULD for the other clauses. I did consider this but don't feel strongly. Interested in other opinions

PR #1514, opened 2018-03-27
WIP Image rights

### LINKED ISSUE #1415: Rename attribution to requiredStatement in Image API
See #1287 (closed by #1413 as complete for Presentation API 3), we need to replicate this for Image API 3.

zimeon:
Closes #1415

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ scripts/factory.py
 scripts/factory.pyc
 .ruby-version
 package-lock.json
+.DS_Store
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,13 @@ install:
   - npm install grunt grunt-html --save-dev
 
 script:
+  - set   
   - rake ci
 
 # environment
 language: ruby
 rvm:
-  - "2.5.1"
+  - "2.4.1"
 jdk:
   - oraclejdk8
 dist: trusty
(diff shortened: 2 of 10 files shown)


zimeon:
Don't review yet -- will wait until we get Travis happy and then do another careful read before asking for review

glenrobson:
This is now building OK. So after some extensive investigation it looks like grunt-html was updated to 9.0.0 but not sure why as it should have been pinned to 3.0.0 or a semantically similar version. 

Anyway version 9 wasn't working due to the node version we were using was too old to support it. On the image-rights branch I've got it working by changing:
 * the travis build server we are using to trusty
 * increasing the versions for ruby and grunt-html 
 * removing the JAVA_OPTIONS which was causing grunt-html to fail in the travis file
 * Fixing some validation errors that the new grunt-html picked up (no longer necessary to have the type attribution in ```<script type="text/javascript">```)
 * Added an ignore when grunt-html thought http://preview.iiif.io/api/image-rights/api/annex/ was written in French...

Now that grunt-html `9.0.0` is out it will affect any branches that don't have this fix.  I suggest you merge this branch into prezi3 then update other branches from that.

zimeon:
Thanks for all your work on this @glenrobson. I'll try to do a careful read
and get this ready to merge Sunday or Monday

zimeon:
Closing this one, will create new PR with just Image spec changes

PR #1516, opened 2018-04-04
grunt fixes

zimeon:
Port of @glenrobson's changes in #1514 but without the API spec changes:

This is now building OK. So after some extensive investigation it looks like grunt-html was updated to 9.0.0 but not sure why as it should have been pinned to 3.0.0 or a semantically similar version. 

Anyway version 9 wasn't working due to the node version we were using was too old to support it. On the image-rights branch I've got it working by changing:
 * the travis build server we are using to trusty
 * increasing the versions for ruby and grunt-html 
 * removing the JAVA_OPTIONS which was causing grunt-html to fail in the travis file
 * Fixing some validation errors that the new grunt-html picked up (no longer necessary to have the type attribution in ```<script type="text/javascript">```)
 * Added an ignore when grunt-html thought http://preview.iiif.io/api/image-rights/api/annex/ was written in French...

Now that grunt-html `9.0.0` is out it will affect any branches that don't have this fix.  I suggest you merge this branch into prezi3 then update other branches from that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ scripts/factory.py
 scripts/factory.pyc
 .ruby-version
 package-lock.json
+.DS_Store
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: false
+
 env:
   global:
     - ROOT_BRANCHES="master prezi3"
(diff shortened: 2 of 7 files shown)


zimeon:
Non-normative change to fix API website, need 2 more editor +1 to merge

PR #1519, opened 2018-04-10
update link header docs

### LINKED ISSUE #1365: Compliance link header usage is unclear
In http://iiif.io/api/image/2.1/#compliance-levels it states that 

> The compliance level URI _MAY_ also be given in the HTTP Link header (RFC5988) with the parameter rel="profile", and thus a complete header might look like:

But is unclear as to which responses this link header would be returned on.

The link header is most valuable on the image bitstream responses, as the compliance level is embedded within the info.json already. It would not be wrong, however, to assert against the info.json as well.  The rationale for the inclusion of the link header at all is to keep the image response a little more self-contained.

Proposal:

Replace the text with:

> The compliance level URI _MAY_ also be given in the HTTP Link header (RFC5988) with the parameter rel="profile", on both Image and Image Information responses. A complete header might look like: [...] 

Thanks to @jbaiter for bringing this up!

azaroth42:
Closes #1365


azaroth42:
Will add the `backticks` back in, and should review for consistency in a different PR.

PR #1521, opened 2018-04-11
clarify return w,h can be < requested for bang w,h

### LINKED ISSUE #1372: Clarify !w,h can return both dimensions less than request
If the client asks for !1000000,1000000 and the image can only be delivered at 500,500, that is a valid response.  This should be noted in the API documentation for clarity.

azaroth42:
Closes #1372

### DIFF
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -210,16 +210,16 @@ Examples:
 
 ###  4.2. Size
 
-The size parameter specifies the dimensions to which the extracted region is to be scaled. With the exception of the `w,h` form, the returned image maintains the aspect ratio of the extracted region as closely as possible.
+The size parameter specifies the dimensions to which the extracted region, which might be the complete image, is to be scaled. With the exception of the `w,h` form, the returned image maintains the aspect ratio of the extracted region as closely as possible.
 
 | Form   | Description |
 |--------|-----------------------|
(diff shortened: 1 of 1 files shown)


source/api/image/3.0/index.md
```diff
@@ -219,7 +219,7 @@ The size parameter specifies the dimensions to which the extracted region is to
 | ,h     | The image or region should be scaled so that its height is exactly equal to h. |
 | pct:n  | The width and height of the returned image is scaled to n% of the width and height of the extracted region. |
 | w,h    | The width and height of the returned image are exactly w and h. The aspect ratio of the returned image _MAY_ be significantly different than the extracted region, resulting in a distorted image. |
-| !w,h   | The image is scaled for the best fit such that the resulting width and height are less than or equal to the requested width and height. The exact scaling _MAY_ be determined by the service provider, based on characteristics including image quality and system performance. |
+| !w,h   | The width and height of the returned image are the greatest allowed width and height, such that the aspect ratio of the returned image is that of the extracted region. The returned image _MAY_ have both height and width less than the requested height and width. The exact scaling _MAY_ be determined by the service provider, based on characteristics including image quality and system performance. |
```
zimeon: This doesn't tie w and h values to the text. I think it should be:

> The extracted region is scaled such that the width and height of the returned image and less than or equal to w and h, while maintaining the aspect ratio. The returned image _MAY_ have both height and width less than the requested height and width. The exact scaling _MAY_ be determined by the service provider, based on characteristics including image quality and system performance.

zimeon: (I note the the previous text also failed to tie to w and h)

PR #1524, opened 2018-04-11
clarify IIIF contexts only for JSON-LD design patterns

### LINKED ISSUE #1515: IIIF JSON-LD patterns and external specs
From pull #1510 we have the principle

> #### 3.4.1. Always Use Arrays for Multiple Value Properties 
> If a property can ever have multiple values, it will always be an array even if it has only one value. This means that code can always iterate over the values without checking first to see if it's an array.  Thus we use `"language": [ "en" ]` and not `"language": "en"`.

https://github.com/IIIF/api/pull/1510/files#diff-5e2e11585f5945b5b9c59190b0a2fc4aR157

In Presentation API section 5.6 there's an example annotation, that includes this:

```
  ...
  "type": "Annotation",
  "motivation": "painting",
  ...
```

Now, `motivation` is not from the IIIF spec - but when when serialised in a IIIF representation, should it follow 3.4.1 and be `"motivation": ["painting"]`?

azaroth42:
Closes #1515

### DIFF
--- a/source/api/annex/notes/design_patterns.md
+++ b/source/api/annex/notes/design_patterns.md
@@ -122,7 +122,7 @@ The representation of the information in JSON must be:
   * Understandable by developers without special training or documentation ;
   * Usable in a variety of application environments to ensure the widest adoption.
 
-These design patterns have been selected in order to maximize these features in the resulting APIs.
+These design patterns have been selected in order to maximize these features in the resulting APIs. Note that these patterns are not enforced for externally defined JSON-LD based systems which are then imported for use within the IIIF context.  For example, the use of `motivation` for Annotations exactly follows the requirements of the Web Annotation Data Model which allows for both a JSON array or a single string, and does not follow the pattern recommended by 3.4.1 below.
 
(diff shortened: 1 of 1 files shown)


source/api/annex/notes/design_patterns.md
```diff
@@ -122,7 +122,7 @@ The representation of the information in JSON must be:
   * Understandable by developers without special training or documentation ;
   * Usable in a variety of application environments to ensure the widest adoption.
 
-These design patterns have been selected in order to maximize these features in the resulting APIs.
+These design patterns have been selected in order to maximize these features in the resulting APIs. Note that these patterns are not enforced for externally defined JSON-LD based systems which are then imported for use within the IIIF context.  For example, the use of `motivation` for Annotations exactly follows the requirements of the Web Annotation Data Model which allows for both a JSON array or a single string, and does not follow the pattern recommended by 3.4.1 below.
```
zimeon: Link for 3.4.1? (and I think I see a double space ;-) )

PR #1525, opened 2018-04-12
Move reqs for referenced content in Collection to 5.1

### LINKED ISSUE #1502: Add recommendation for thumbnails in referenced collections
As a developer of mobile Presentation API client I'd like to propose addition of simple recommendation along with brief explanation to the Presentation API specification. The recommendation is that collection references are recommended to contain collection thumbnail if present. The reason is following:

_Client sends request for a collection and gets list of references to other collections. To enhance user experience, client is required to display a thumbnail for each item, but references only require to contain link and title. Displaying thumbnail therefore means to send at least one additional request for each item. If thumbnail is missing, client displays first image resource from item's content. In case of collections this leads to DFS like algorithm. This may result in lot of network requests just in order to display single thumbnail._

Typical response as for Presentation API 2.1.1:
```
{
  "@context": "http://iiif.io/api/presentation/2/context.json",
  "@id": "http://www.example.org/path/to/collection/top",
  "@type": "sc:Collection",
  "label": "Top level collection",

  "members": [
    {
      "@id": "http://www.example.org/path/to/collection/one",
      "@type": "sc:Collection",
      "label": "Collection one"
    },
    ...
  ]
}
```

Proposed response:
```
{
  "@context": "http://iiif.io/api/presentation/2/context.json",
  "@id": "http://www.example.org/path/to/collection/top",
  "@type": "sc:Collection",
  "label": "Top level collection",

  "members": [
    {
      "@id": "http://www.example.org/path/to/collection/one",
      "@type": "sc:Collection",
      "label": "Collection one",
      "thumbnail": "http://www.example.org/path/to/thumbnail_1.png"
    },
    ...
  ]
}
```

azaroth42:
Closes #1502

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1028,7 +1028,7 @@ Note that while the Collection _MAY_ reference Collections or Manifests from pre
 
 ###  5.2. Manifest
 
-The Manifest resource typically represents a single object and any intellectual work or works embodied within that object. In particular it includes descriptive, rights and linking information for the object. It embeds the Canvases that should be rendered to the user as views of the object. The Manifest response contains sufficient information for the client to initialize itself and begin to display something quickly to the user.
+The Manifest resource typically represents a single object and any intellectual work or works embodied within that object. In particular it includes descriptive, rights and linking information for the object. The Manifest embeds the Canvases that should be rendered as views of the object and contains sufficient information for the client to initialize itself and begin to display something quickly to the user.
 
 The identifier in `id` _MUST_ be able to be dereferenced to retrieve the JSON description of the Manifest, and thus _MUST_ use the HTTP(S) URI scheme.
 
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -1024,7 +1032,7 @@ The Manifest resource typically represents a single object and any intellectual
...
+The Manifest _MUST_ have an `items` property, which is an array of JSON-LD objects. Each object is a Canvas, with requirements as described in the next section. It _MAY_ also have a `structures` property listing one or more [Ranges][prezi-api-3-range] which describe additional structure of the content, such as might be rendered as a table of contents.  The Manifest _MAY_ have an `annotations` property, which includes Annotation Page resources where the Annotations have the Manifest as their `target`. These will typically be comment style annotations, and _MUST NOT_ have `painting` as their `motivation`.
```
zimeon: Not introduced by this commit but I think the third sentence should start _The Manifest_ rather than _It_.

jpstroop: 👍

PR #1582, opened 2018-04-19
Unpin Kramdown

jpstroop:
Test branch w/ latest Jekyll and Kramdown. Right now it branches from `hard_code_headings`, which branches from the latest `master`, so I think CSS etc. should be up to date. The only thing I'm seeing is that the CSS for inline examples is messed up, e.g.:

screen shot 2018-04-18 at 7 58 59 pm

I know @zimeon and @tomcrane were playing w/ CSS earlier today--would that be the cause? I can't think why else the CSS would suddenly be screwy.


jpstroop:
CSS now updated as well:

screen shot 2018-04-24 at 9 00 34 pm

Will make a PR on the website, after which this should be good to go.

glenrobson:
Website pull request is merged.

mikeapp:
http://preview.iiif.io/api/unpin_kramdown/api/presentation/3.0/#b-example-manifest-response lost highlighting

jpstroop:
Hmm, good catch. Looks like all of the other highlighting may be a little
different as well?

azaroth42:
Yeah, looks like the highlighting css is messed up

jpstroop:
I'm have a look this afternoon or tonight, probably just another css context thing--looks like all the styling is there

jpstroop:
Here's the issue. Can't yet figure out why though:

screen shot 2018-04-25 at 1 15 37 pm

jpstroop:
OK--found the latest possible version where keys are still a different color. Not sure if it's a bug or a feature, but I will try diffing the branches later to see. If it is a bug, it was introduced some time ago and there have been several releases since.

For the record: `2.0.5` works, `2.0.6` doesn't. List of releases is [here](https://rubygems.org/gems/rouge/versions).

jpstroop:
A little 💚?

PR #1590, opened 2018-04-28
Discovery 0.1

azaroth42:
PR for merging the discovery API into master. Do Not Merge before Wednesday May 2nd.

### DIFF
--- a/source/api/discovery/0.1/index.md
+++ b/source/api/discovery/0.1/index.md
@@ -94,7 +94,7 @@ This specification describes three levels of conformance that build upon each ot
 ### 2.1. Listing Resources and their Changes
 {: #listing-resources-and-their-changes}
 
-There are three levels of conformance at which changes can be described. Level 0 is simply a list of the resources available.  Level 1 adds timestamps and ordering from earliest change to most recent, allowing the consumer to stop processing once it encounters a change that it has already processed. Level 2 adds in additional clarity about the types of activities, enabling the explicit creation and deletion of resources.
+There are three levels of conformance at which changes can be described. Level 0 is simply a list of the resources available.  Level 1 adds timestamps and ordering from earliest change to most recent, allowing the consumer to stop processing once it encounters a change that it has already processed. Level 2 adds information about the types of activities, enabling the explicit creation and deletion of resources.
 
 #### 2.1.1. Level 0: Basic Resource List
 {: #level-0-basic-resource-list}
@@ -167,7 +167,7 @@ Activities are collected together into pages that together make up the entire se
   "id": "https://example.org/activity/page-1",
   "type": "OrderedCollectionPage",
   "partOf": {
-    "id": "https://example.org/activity/top-collection",
+    "id": "https://example.org/activity/all-changes",
     "type": "OrderedCollection"
   },
   "prev": {
@@ -311,7 +311,7 @@ OrderedCollections _MAY_ have a `totalItems` property.  The value _MUST_ be a no
     "http://iiif.io/api/discovery/0/context.json",
     "https://www.w3.org/ns/activitystreams"
   ],
-  "id": "https://example.org/activity/top-collection",
+  "id": "https://example.org/activity/all-changes",
   "type": "OrderedCollection",
   "totalItems": 21456,
   "first": {
@@ -361,7 +361,7 @@ Ordered Collection Pages _SHOULD_ have a `partOf` property. The value _MUST_ be 
 ```
 {
   "partOf": {
-    "id": "https://example.org/activity/top-collection",
+    "id": "https://example.org/activity/all-changes",
     "type": "OrderedCollection"
   }
 }
@@ -441,7 +441,7 @@ Ordered Collection Pages _MUST_ have a `orderedItems` property.  The value _MUST
   "type": "OrderedCollectionPage",
   "startIndex": 20,
   "partOf": {
-  	"id": "https://example.org/activity/top-collection",
+  	"id": "https://example.org/activity/all-changes",
   	"type": "OrderedCollection"
   },
   "prev": {


source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+This specification leverages existing techniques, specifications, and tools in order to promote widespread adoption. It enables the collaborative development of global or thematic search engines and portal applications that allow users to easily find and engage with content available via existing IIIF APIs.
+
+### 1.1. Objectives and Scope
+
+The objective of the IIIF Change Discovery API is to provide the information needed to discover and subsequently make use of existing IIIF resources.  The intended audience is other IIIF aware systems that can proactively work with the content and APIs. While this work may benefit others outside of the IIIF community directly or indirectly, the objective of the API is to specify an interoperable solution that best and most easily fulfills the discovery needs within the community of participating organizations.
```
zimeon: delete "existing" ?
"proactively work with" seems to imply too much, maybe something simpler like "use" or "leverage", or "process the content or use the APIs"?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+The discovery of IIIF resources requires a consistent and well understood pattern for content providers to publish lists of links to their available content. This does not include the transmission of the content itself, for example transferring any source media or data between systems, only for the discovery of that content.  This allows a baseline implementation of discovery systems that process the list, looking for resources that have been added or changed.
```
zimeon: delete "for" or perhaps delete all of ", only for the discovery of that content"?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+This process can be optimized by allowing the content providers to publish descriptions of when their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes no longer available. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll all of the systems to see if anything has changed.
```
zimeon: "otherwise becomes no longer available" is awkward, suggest either "otherwise becomes unavailable" or maybe simpler "otherwise ceases to be available"?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+This process can be optimized by allowing the content providers to publish descriptions of when their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes no longer available. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll all of the systems to see if anything has changed. 
+
+Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. These are out of scope as the diverse domains represented within the IIIF community already have different standards in these spaces, and previous attempts to reconcile these specifications across domains have not been successful.
```
zimeon: Could remove "These are out of scope as" start to second sentence as context seems adequately implied by previous sentence

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+__Notification of Changes__<br>This draft version of the specification does not include the subscription mechanism for enabling notifications to be pushed to remote systems about changes.  The current specification only enables the polling pattern where the set of changes must be periodically reprocessed. Notifications are likely to be added in a future version before 1.0.
```
zimeon: "about changes" is dangling, could be removed as notifications seems enough on its own

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+The resources available via the Presentation API do not have descriptive metadata fields suitable for indexing beyond a full text search. Instead, they link to external documents using the `seeAlso` property that can have richer and domain-specific information about the content being presented. For example, a museum object might have a `seeAlso` reference to a CIDOC-CRM or LIDO description, while a bibliographic resource might reference a BibFrame or MODS description. These external descriptions should be used when possible to provide interfaces giving access to more precise matching algorithms.
+
+There are three levels of detail that can be implemented by a conforming system. The levels build upon each other in terms of functionality enabled and precision of the information published. Sets of these changes are published in [pages](#pages-of-changes), which are then collected together in a [collection](#collections-of-pages) per publisher. Care has been take to allow the implementation of all levels to be done in as static a way as possible, rather than requiring dynamic access to a database.
```
zimeon: First sentence is repeated in following paragraph so suggest trimming here, suggest starting 

> This specification describes three levels of conformance that build upon each other in terms of functionality enabled and precision of the information published.

the following sentence then starts "Sets of these changes ... " where "these" should I think be deleted

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+### 2.1. Listing Resources and their Changes
+
+There are three levels of completeness at which changes can be described. Level 0 is simply a list of the resources available.  Level 1 adds timestamps and ordering from earliest change to most recent, allowing the consumer to stop processing once it encounters a change that it has already processed. Level 2 adds in additional clarity about the types of activities, enabling the explicit creation and deletion of resources.
+
+#### 2.1.1. Level 0: Basic Resource List
+
+The basic information required to provide a minimally effective set of links to IIIF resources is just the URIs of those resources. However, with the addition of a little JSON template around those URIs, we can be on the path towards a robust set of information that clients can use to optimize their processing.
```
zimeon: delete "basic" ?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. These are out of scope as the diverse domains represented within the IIIF community already have different standards in these spaces, and previous attempts to reconcile these specifications across domains have not been successful.
+
+__Notification of Changes__<br>This draft version of the specification does not include the subscription mechanism for enabling notifications to be pushed to remote systems about changes.  The current specification only enables the polling pattern where the set of changes must be periodically reprocessed. Notifications are likely to be added in a future version before 1.0.
+{: .warning}
+
+### 1.2. Terminology
+
+The specification uses the following terms:
+
+* __HTTP(S)__: The HTTP or HTTPS URI scheme and internet protocol.
```
zimeon: Should "Activity" be added with link to AS ? This is used in 2.1.1 and many times thereafter

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+#### 2.1.1. Level 0: Basic Resource List
+
+The basic information required to provide a minimally effective set of links to IIIF resources is just the URIs of those resources. However, with the addition of a little JSON template around those URIs, we can be on the path towards a robust set of information that clients can use to optimize their processing.  
+
+Starting with the IIIF resource URIs, we add an "Update" Activity wrapper around them.  The order of the resources in the resulting list is unimportant, but each should only appear once. In terms of optimization, this approach provides no additional benefit over any other simpler list format, but is compatible with the following levels which introduce significant benefits.  This is the minimum level for interoperability, while being compatible with the more detailed patterns described below.
+
+If resources are deleted after being refered to in the resource list, the entire list should be republished without the reference to the deleted resource. Clients should also expect to encounter links that are out of date and no longer resolve to a IIIF Manifest or Collection.
```
zimeon: s/encounter links/encounter resource URIs/ to be consistent?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+Example Level 0 Activity:
+
+```
+{ 
+  "type": "Update",
+  "object": {
+  	"id": "https://example.org/iiif/1/manifest",
```
zimeon: `id` and `type` indented too far

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+  	"type": "Manifest"  	
+  }
+}
+```
+
+#### 2.1.2. Level 1: Basic Change List
+
+The most effective information to add beyond the basic resource list is the datestamp at which the resource was last modified (including the initial modification that created it).  If we know these dates, we can add them to the Activities and order the list such that the most recent activities occur last. The timestamp is given in the `endTime` property -- the time at which the document update process finished. Consumers will then process the list of Activities in reverse order, from last to first, stopping when they encounter an Activity they have already processed in a previous run.
```
zimeon: Suggest changing

> The timestamp is given in the endTime property – the time at which the document update process finished.

to 

> The time at which the document update process finished is given in the `endTime` property

(I think one could actually remove all 4 instances of the word "timestamp")

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+#### 2.1.2. Level 1: Basic Change List
+
+The most effective information to add beyond the basic resource list is the datestamp at which the resource was last modified (including the initial modification that created it).  If we know these dates, we can add them to the Activities and order the list such that the most recent activities occur last. The timestamp is given in the `endTime` property -- the time at which the document update process finished. Consumers will then process the list of Activities in reverse order, from last to first, stopping when they encounter an Activity they have already processed in a previous run. 
+
+Example Level 1 Activity:
+
+```
+{
+  "type": "Update",
+  "object": {
```
zimeon: This object doesn't have an `updated` which seems at odd with the statement above "If we know these [last modified] dates, we can add them to the Activities"

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+    "id": "https://example.org/iiif/1/manifest",
+    "type": "Manifest"
+  },
+  "endTime": "2017-09-20T00:00:00Z"
+}
+```
+
+### 2.2. Pages of Changes
+
+These Activities are collected together into pages that together make up the entire set of changes that the publishing system has made.  Pages reference the previous and next pages in that set, and the overall collection that they are part of.  The Activities are then listed in time order.
```
zimeon: Delete leading "These"?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+
+```
+{
+  "id": "https://example.org/activity/page-1",
+  "type": "OrderedCollectionPage",
+  "partOf": {
+  	"id": "https://example.org/activity/top-collection",
```
zimeon: mixed indentation

zimeon: would `.../entire-set-of-changes` be better than `.../top-collection` to tie to the text?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+}
+```
+
+### 2.5. Collections of Pages
+
+As the number of Activities is likely too many to usefully be represented in a single Page, they are collected together into a Collection as the initial entry point.  The Collection references the URIs of the first and last pages.
+
+```
+{
+  "id": "https://example.org/activity/top-collection",
```
zimeon: suggest `.../entire-set-of-changes` instead of `.../top-collection` as above

source/api/discovery/0.1/index.md
azaroth42: :+1:
:+1:

source/api/discovery/0.1/index.md
azaroth42: :+1: Also moved it to the last section on out of scope work.

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: :+1: Added "change" before notifications

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+* __HTTP(S)__: The HTTP or HTTPS URI scheme and internet protocol.
+
+The terms _array_, _JSON object_, _number_, and _string_ in this document are to be interpreted as defined by the [Javascript Object Notation (JSON)][org-rfc-8259] specification.
+
+The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _SHOULD NOT_, _RECOMMENDED_, _MAY_, and _OPTIONAL_ in this document are to be interpreted as described in [RFC 2119][org-rfc-2119].
+
+
+## 2. Overview of IIIF Resource Discovery
+
+In order to discover IIIF resources, the state of the publishing system needs to be communicated succinctly and easily to a consuming system.  The consumer can then use that information to retrieve and process the resources of interest.  This communication takes place via the IIIF Change Discovery API, which uses the W3C [Activity Streams][org-w3c-activitystreams] specification to describe and serialize changes to resources, where the semantics of those changes and the interactions between publishers and consumers are based on those identified by the [ResourceSync][org-openarchives-rsync] framework.
```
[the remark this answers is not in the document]
azaroth42: :+1: I tweaked the current wording of the first two sentences.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+
+The resources available via the Presentation API do not have descriptive metadata fields suitable for indexing beyond a full text search. Instead, they link to external documents using the `seeAlso` property that can have richer and domain-specific information about the content being presented. For example, a museum object might have a `seeAlso` reference to a CIDOC-CRM or LIDO description, while a bibliographic resource might reference a BibFrame or MODS description. These external descriptions should be used when possible to provide interfaces giving access to more precise matching algorithms.
```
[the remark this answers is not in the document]
azaroth42: I DO NOT SHOUT NON ACRONYMS.  We can replace with another specification?

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: "basic" -> "core"

source/api/discovery/0.1/index.md
azaroth42: Propose to make an issue for this and discuss on call.

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: Bleh. Jekyll seems to treat tab as 4 spaces not two, like my editor.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,608 @@
...
+__Latest Stable Version:__ None
+
+__Previous Version:__ None
+
+**Editors**
+
+  * **[Michael Appleby](https://orcid.org/0000-0002-1266-298X)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0002-1266-298X), [_Yale University_](http://www.yale.edu/)
+  * **[Tom Crane](https://orcid.org/0000-0003-1881-243X)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0003-1881-243X), [_Digirati_](http://digirati.com/)
+  * **[Robert Sanderson](https://orcid.org/0000-0003-4441-6852)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0003-4441-6852), [_J. Paul Getty Trust_](http://www.getty.edu/)
+  * **[Jon Stroop](https://orcid.org/0000-0002-0367-1243)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0002-0367-1243), [_Princeton University Library_](https://library.princeton.edu/)
```
jpstroop: I'm happy to be associated with this, but TBH I've done very little for this initiative over the past year and I'm not sure it's fair to the rest of the group. Will let others decide.

source/api/discovery/0.1/index.md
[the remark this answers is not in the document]
zimeon: Fine by me, could just have MODS  too

source/api/discovery/0.1/index.md
azaroth42: It has the `endTime` of the update?

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+  "object": {
+    "id": "https://example.org/iiif/1/manifest",
+    "type": "Manifest"  	
+  }
+}
+```
+
+#### 2.1.2. Level 1: Basic Change List
+
+The most effective information to add beyond the basic resource list is the timestamp at which the resource was last modified (including the initial modification that created it). If we know these dates, we can add them to the Activities and order the list such that the most recent activities occur last. The time at which the document update process finished is given in the `endTime` property. Consumers will then process the list of Activities in reverse order, from last to first, stopping when they encounter an Activity they have already processed in a previous run.
```
zimeon: Suggest changes to make it clear that the `endTime` is the timestamp we are talking about:

> The most effective information to add beyond the basic resource list is the timestamp at which the resource was last modified (including the initial modification that created it). If we know these timestamps, we can add them to the Activities using the `endTime` property, the time at which the resource update process finished. Lists with multiple activities are then ordered such that the most recent activities occur last. Consumers will then process the list of Activities in reverse order, from last to first, stopping when they encounter an Activity they have already processed in a previous run.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+        "id": "https://example.org/iiif/2/manifest",
+        "type": "Manifest"
+      },
+      "endTime": "2018-03-11T16:30:00Z"
+    }
+  ]
+}
+```
+
+### 2.5. Collections of Pages
```
zimeon: 2.3 ?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+    "id": "https://example.org/iiif/1/manifest",
+    "type": "Manifest"
+  },
+  "endTime": "2017-09-20T00:00:00Z"
+}
+```
+
+### 2.2. Pages of Changes
+
+Activities are collected together into pages that together make up the entire set of changes that the publishing system has made.  Pages reference the previous and next pages in that set, and the overall collection that they are part of.  The Activities are then listed in time order.
```
zimeon: Remind which time order, suggest replace

> The Activities are then listed in time order.

with

> The Activities are then listed in time order such that the most recent activities occur last.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+```
+{ "id": "https://example.org/activity/all-changes" }
+```
+
+##### type
+
+The class of the Ordered Collection.
+
+Ordered Collections _MUST_ have a `type` property.  The value _MUST_ be "OrderedCollection".
```
zimeon: are we going to adopt the new code-font styling for literals like `OrderedCollection`?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+```
+{ "type": "OrderedCollection" }
+```
+
+##### first
+
+A link to the first Ordered Collection Page for this Collection.
+
+Ordered Collections _SHOULD_ have a `first` property.  The value _MUST_ be a JSON object, with the `id` and `type` properties.  The value of the `id` property _MUST_ be a string, and it _MUST_ be the HTTP(S) URI of the first page of items in the Collection. The value of the `type` property _MUST_ be a string, and _MUST_ be "OrderedCollectionPage".
```
zimeon: `“OrderedCollectionPage”`

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+    "type": "OrderedCollectionPage"
+  }
+}
+```
+
+##### last
+
+A link to the last Ordered Collection Page for this Collection.  As the client processing algorithm works backwards from the most recent to least recent, the inclusion of `last` is _REQUIRED_, but `first` is only _RECOMMENDED_.
+
+Ordered Collections _MUST_ have a `last` property.  The value _MUST_ be a JSON object, with the `id` and `type` properties.  The value of the `id` property _MUST_ be a string, and it _MUST_ be the HTTP(S) URI of the last page of items in the Collection. The value of the `type` property _MUST_ be a string, and _MUST_ be "OrderedCollectionPage".
```
zimeon: `“OrderedCollectionPage”`

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+```
+{ "totalItems": 21456 }
+```
+
+##### Complete Ordered Collection Example
+
+```
+{
+  "@context": "",
```
zimeon: create a dummy context in the logical location perhaps?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+##### type
+
+The type of Activity. 
+
+This specification uses the types described in the table below.
+
+| Type   | Definition |
+| ------ | ---------- |
+| Create | The initial creation of the resource.  Each resource _MUST_ have at most one `Create` Activity in which it is the `object`. |
```
zimeon: code font for `Create` etc.?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+This process can be optimized by allowing the content providers to publish descriptions of when their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes unavailable. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll all of the systems to see if anything has changed. 
+
+Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. The diverse domains represented within the IIIF community already have different standards in these spaces, and previous attempts to reconcile these specifications across domains have not been successful. Also out of scope is optimization of the transmission of content, for example recommendations about transferring any source media or data between systems.
```
jpstroop: > already have different standards in these spaces, and previous attempts to reconcile these specifications across domains have not been successful.

I think I'd prefer:

...already have successful standards fulfilling these use cases, and previous attempts to reconcile these standards across domains have not seen widespread adoption.

mikeapp: 👍

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+Activities are used to describe the state of the publisher by recording each individual change, in the order that they occur. The changes described are the creation, modification and deletion of IIIF Presentation API resources, primarily Collections and Manifests.  If the consuming application is aware of all of the changes that took place at the publisher, it would have full knowledge of the set of resources available.  The focus on IIIF Collections and Manifests is because these are the main access points to published content and references to descriptive metadata about that content, however Activities describing changes to other resources could also be published in this way.
```
jpstroop: > primarily Collections and Manifests.

primarily, but not necessarily, or really only ever? I know the answer, but I think this confuses things and is better left implicit by the use cases and examples that surface throughout the rest of the spec.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+The resources available via the Presentation API do not have descriptive metadata fields suitable for indexing beyond a full text search. Instead, they link to external documents using the `seeAlso` property that can have richer and domain-specific information about the content being presented. For example, a museum object might have a `seeAlso` reference to a CIDOC-CRM or LIDO description, while a bibliographic resource might reference a Dublin Core or MODS description. These external descriptions should be used when possible to provide interfaces giving access to more precise matching algorithms.
+
+This specification describes three levels of conformance that build upon each other in terms of functionality enabled and precision of the information published. Sets of changes are published in pages, which are then collected together in a collection per publisher. Care has been take to allow the implementation of all levels to be done in as static a way as possible, rather than requiring dynamic access to a database.
+
```
jpstroop: > Care has been take to allow the implementation of all levels to be done in as static a way as possible, rather than requiring dynamic access to a database.

To reduce barriers to implementation, care has been taken to make it possible to implement all levels using static files on a web server, rather than requiring resources to be created dynamically, through this should be equally possible.

source/api/discovery/0.1/index.md
jpstroop: I also think introducing this w/ a use case might smooth things out:

When dealing with large sets of resources, it can be useful to work with only those resources that have changed since the last time the list was processed. This can be facilitated by the addition of a time stamp that indicates when a resource was last modified using the `endTime` property. Lists with multiple activities are then ordered such that the most recent activities occur last. Consumers will then process the list of Activities in reverse order, from last to first, stopping when they encounter an Activity they have already processed in a previous run.

jpstroop: This has never been my forté, but why

> Lists with multiple activities are then ordered such that the most recent activities occur last [...] Consumers will then process the list of Activities in reverse order, from last to first ...

Why not reverse the order (i.e. to _most recent first_) so that the client can process the list from top to bottom rather than having to pull the whole thing into memory and reverse it? Following the 'working with large sets' use case, it seems friendlier to me to favor the client by enabling it to use a push/streaming/SAX-like parser.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+    "type": "Manifest"
+  },
+  "endTime": "2017-09-20T00:00:00Z"
+}
+```
+
+#### 2.1.3. Level 2: Complete Change List
+
+At the most detailed level, a log of all of the Activities that have taken place can be recorded, with the likelihood of multiple Activities per IIIF resource.  This allows the additional types of "Create" and "Delete", enabling a synchronization process to remove resources as well as add or update them. The list might end up very long if there are many changes to resources, however this is not a typical situation, and the cost is still reasonable as each entry is short and can be compressed both on disk and at the HTTP(S) transport layer.
+
```
jpstroop: > The list might end up very long if there are many changes to resources, however this is not a typical situation, and the cost is still reasonable as each entry is short and can be compressed both on disk and at the HTTP(S) transport layer.

I don't think it's necessary to say this.

source/api/discovery/0.1/index.md
jpstroop: > that they are part of.

Sentence ending w/ preposition. Suggest:

of which they are a part.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+      },
+      "endTime": "2018-03-11T16:30:00Z"
+    }
+  ]
+}
+```
+
+### 2.5. Collections of Pages
+
+As the number of Activities is likely too many to usefully be represented in a single Page, they are collected together into a Collection as the initial entry point. The Collection references the URIs of the first and last pages.
```
jpstroop: > be represented in a single Page

capital P on page?

source/api/discovery/0.1/index.md
azaroth42: This was more explicit in the document as to why, but there was some pushback it was too much detail as to why rather than what.

The rationale is that the first pages become static resources once finished. Whereas a constantly shifting list is always dynamic.  So publishing is push() meaning that consumption is pop().

jpstroop: OK I get it, and it definitely makes sense for supporting the static implementation/level 0 use case.  That said--and I'm the last to argue for hypothetical use-cases--I do think that client streaming, especially of complete change sets, isn't a stretch. Would it be too fussy to have a flag that indicates which way the list is sorted? (probably)

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: :+1: to the changes. Have tried to synthesize, starting from Jon's.

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: Yes, will do this through out. :+1:

source/api/discovery/0.1/index.md
azaroth42: There's an issue for this https://github.com/IIIF/discovery/issues/28 but I put in the likely statement.

source/api/discovery/0.1/index.md
azaroth42: 👍

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: Oh, this used to work until the CSS change. Will fix.

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: This is the door ajar wording for other resources being notified about.  We should wait until https://github.com/IIIF/discovery/issues/12 is resolved. Prefer to leave this as is for now.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+The resources available via the Presentation API do not have descriptive metadata fields suitable for indexing beyond a full text search. Instead, they link to external documents using the `seeAlso` property that can have richer and domain-specific information about the content being presented. For example, a museum object might have a `seeAlso` reference to a CIDOC-CRM or LIDO description, while a bibliographic resource might reference a Dublin Core or MODS description. These external descriptions should be used when possible to provide interfaces giving access to more precise matching algorithms.
```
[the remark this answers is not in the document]
azaroth42: Yep, will work on it.

source/api/discovery/0.1/index.md
azaroth42: :+1:

azaroth42: (Give or take)

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+### 2.1. Listing Resources and their Changes
+
+There are three levels of conformance at which changes can be described. Level 0 is simply a list of the resources available.  Level 1 adds timestamps and ordering from earliest change to most recent, allowing the consumer to stop processing once it encounters a change that it has already processed. Level 2 adds in additional clarity about the types of activities, enabling the explicit creation and deletion of resources.
+
+#### 2.1.1. Level 0: Basic Resource List
+
+The core information required to provide a minimally effective set of links to IIIF resources is just the URIs of those resources. However, with the addition of a little JSON template around those URIs, we can be on the path towards a robust set of information that clients can use to optimize their processing.
```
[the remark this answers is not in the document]
azaroth42: The intent is that 2.1.1 is just an array of URIs, with a bunch of boilerplate around each item. The extra information doesn't actually help at all, it's just setting up the possibility for the higher levels.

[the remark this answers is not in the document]
zimeon: Maybe just change from "template" to something else, perhaps "However, simple JSON embedding of those URIs is the first step along a path towards a robust set of information that clients can use to optimize their processing."?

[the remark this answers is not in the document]
azaroth42: Have changed the wording, but kept the intent.

source/api/discovery/0.1/index.md
azaroth42: :+1:

zimeon: 👍

source/api/discovery/0.1/index.md
azaroth42: :+1:

source/api/discovery/0.1/index.md
azaroth42: Shorthand for Ordered Collection Page. There also isn't a Collection, and no type in the data that is an Activity.  So it's all a bit fuzzier than the Prezi set of classes.

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,633 @@
...
+
+The Presentation API does not directly include descriptive metadata fields suitable for indexing beyond a full text search. The data intentionally lacks the semantics needed to construct indexes needed to enable advanced or fielded search. Instead, they link to external documents using the `seeAlso` property that can have richer and domain-specific information about the content being presented. For example, a museum object might have a `seeAlso` reference to a CIDOC-CRM or LIDO description, while a bibliographic resource might reference a Dublin Core or MODS description. These external descriptions should be used when possible to provide interfaces giving access to more precise matching algorithms.
```
mikeapp: In sentence three I'd reorder:  "Instead, they use the `seeAlso` property to link to external documents that can have richer and ..."

azaroth42: :+1:

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,633 @@
...
+  	"id": "https://example.org/activity/page-214",
+  	"type": "OrderedCollectionPage"
+  }
+}
+```
+
+
+## 3. Activity Streams Details
+
+The W3C [Activity Streams][org-w3c-activitystreams] specification defines a "model for representing potential and completed activities", and is compatible with the [design patterns][annex-patterns] established for IIIF APIs. It is defined in terms of [JSON-LD][org-w3c-json-ld], and can be seamlessly integrated with the existing IIIF APIs. The model can be used to represent activities of creating, updating and deleting (or otherwise de-publishing) IIIF resources, carried out by content publishers.
```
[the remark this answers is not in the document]
azaroth42: :+1:

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,633 @@
...
+
+The Presentation API does not directly include descriptive metadata fields suitable for indexing beyond a full text search. The data intentionally lacks the semantics needed to construct indexes needed to enable advanced or fielded search. Instead, they use the `seeAlso` property to link to external documents that can have richer and domain-specific information about the content being presented. For example, a museum object might have a `seeAlso` reference to a CIDOC-CRM or LIDO description, while a bibliographic resource might reference a Dublin Core or MODS description. These external descriptions should be used when possible to provide interfaces giving access to more precise matching algorithms.
```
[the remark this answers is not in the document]
azaroth42: :+1:

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,633 @@
...
+  "object": {
+  	"id": "http://example.org/iiif/1/manifest",
+  	"type": "Manifest"
+  }
+}
+```
+
+##### endTime
+
+The time at which the Activity was finished. It is up to the implementer to decide whether the Activity includes the publication of the IIIF resource online, or only the internal data modification, but the decision _MUST_ be consistently applied.
```
tomcrane: Should the spec's use for discovery require that the endTime must correspond to a generally available change in the resource published at the object id? How does it otherwise know that a reindex of the published resource right now will pick up a change?

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,633 @@
...
+```
+
+##### endTime
+
+The time at which the Activity was finished. It is up to the implementer to decide whether the Activity includes the publication of the IIIF resource online, or only the internal data modification, but the decision _MUST_ be consistently applied. 
+
+Activities _SHOULD_ have the `endTime` property.  The value _MUST_ be a datetime expressed in UTC in the [xsd:dateTime][org-w3c-xsd-datetime] format.
+
+```
+{ "endTime": "2017-09-21T00:00:00Z" }
```
tomcrane: Can an endTime be in the future?

tomcrane: (almost certainly not, but...)

azaroth42: I think there's a hypothetical use case where it is -- you're alerting that some update /will/ happen, rather than has already happened.  Not sure we need to specify either way?

source/api/discovery/0.1/index.md
azaroth42: I don't think it must be the time at which the publication happened, but it must be published. Does that need to be explicit?

source/api/discovery/0.1/index.md
tomcrane: The first comment is the bigger issue. I publish my updates, you come along and harvest them, notice that resource X has changed since you were last here, and fetch the resource from the CDN. The change has been made in my internal system, but the cache hasn't expired yet at the CDN. Yes, I know I could mitigate against this, but the spec should say that I need to ensure the resource is really ready for reindexing at its public URL.

azaroth42: I think the important point is that the changed resource must be available at the URI before the time listed in the activity.  If it's in the future, then it's a promise. If it's in the past, then it must already be available.

zimeon: I agree with the above, the expectation is that the "resource" (which is web speak means the thing at URI that you can GET a representation of -- irrespective of whether served by CDN or other means) should be available by `endTime`

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,656 @@
...
+
+This specification describes three levels of conformance that build upon each other in terms of functionality enabled and precision of the information published. Sets of changes are published in pages, which are then collected together in a collection per publisher. To reduce barriers to implementation, care has been take to allow the implementation of all levels using only static files on a web server, rather than requiring dynamic access to a database.
+
+### 2.1. Listing Resources and their Changes
+{: #listing-resources-and-their-changes}
+
+There are three levels of conformance at which changes can be described. Level 0 is simply a list of the resources available.  Level 1 adds timestamps and ordering from earliest change to most recent, allowing the consumer to stop processing once it encounters a change that it has already processed. Level 2 adds in additional clarity about the types of activities, enabling the explicit creation and deletion of resources.
```
zimeon: > adds in additional clarity about

seems both awkward to read and a little bold (clarity), perhaps:

> adds information about

source/api/discovery/0.1/index.md
```diff
@@ -0,0 +1,656 @@
...
+```
+{
+  "@context": [
+    "http://iiif.io/api/discovery/0/context.json",
+    "https://www.w3.org/ns/activitystreams"
+  ],
+  "id": "https://example.org/activity/page-1",
+  "type": "OrderedCollectionPage",
+  "partOf": {
+    "id": "https://example.org/activity/top-collection",
```
zimeon: s/top-collection/all-changes/

zimeon: actually there are several instances of this

zimeon:
LGTM modulo 2 minor comments above

source/api/discovery/0.1/index.md
mikeapp: I understand but was really thinking about future ‘endTimes’ - my hypothetical case was that an implementation behind a CDN might benefit from publishing ’endTime’s in the future when the resource would be universally available in its updated form - for example, a simple implementation that’s serializing the Update entries as the new Manifests are written.   It’s easy enough to work around, but was something I hadn’t considered until I read @tomcrane ‘s comment.

tomcrane: I am OK with the new version. I just have the observation that it puts more burden on the consumer, when chunking through new activities with endTimes later than the previous crawl, it has to be aware that the endTime might be in the future and it can't go and get that resource right now - it might have to shelve that reindex, queue it somewhere etc - which requires a more sophisticated harvester capable of deferring fetches for non-trivial amounts of time.

azaroth42:
@jpstroop @tomcrane ?

PR #1528, opened 2018-04-15
Add Prezi3 Changelog

azaroth42:
Changelog for Prezi3.

PR #1597, opened 2018-05-07
Add prezi3 changelog

zimeon:
Instead of #1528 (where all the comments are), merge against `master` since we are no longer working on `prezi3`


jpstroop:
Faith-based :+1: because we should get this out and won't be able to look until late this afternoon.

PR #1610, opened 2018-05-16
Fix link to sizes deprecation warning in Image 3.0 change log

zimeon:
Trivial fix to fix link that currently goes to wrong deprecation warning


azaroth42:
All +1s, merging.

Issue #1601, opened 2018-05-09
Image API requiredStatement example should use array


mikeapp:
http://iiif.io/api/image/3.0/#55-rights-related-properties has 
```
"label": { "en": "Attribution" }
"value": { "en": "Provided by Example Organization" }
```

azaroth42:
:-1: `requiredStatement`was decided to be singular (IIRC) to avoid publishers putting in ALL their metadata as required.

azaroth42:
Oh the *values* ... sorry, you're completely correct!

azaroth42:
Should be `["Attribution"]` (etc)

mikeapp:
This is already done, see merge above.

PR #1611, opened 2018-05-18
Add arrays around strings in language maps

zimeon:
Examples of `requiredStatement` in image 3.0 draft are wrong. Values of `label` and `value` _MUST_ have an array for every language key per http://iiif.io/api/presentation/3.0/#43-language-of-property-values


azaroth42:
Closes #1601

PR #1635, opened 2018-07-06
Updating links to work with https

glenrobson:
Adding test to check for mixed content

Pointing fonts to use https versions.

Plan is to run the iiif.io site in parallel as both http and https while we look at what needs to change in the site. Things that might need to change are:

 * Fixtures (moving from http://iiif.io to https://iiif.io
 * Specs? do we need to change the link to the context to use the https link?

I've checked the validators and they should work OK as they don't have any client side code to retrieve http manifests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -19,6 +19,7 @@ task :ci do
   build_site
   sh 'grunt test'
   sh 'scripts/check_json.py -v'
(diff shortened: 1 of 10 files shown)


glenrobson:
Thanks all.

PR #1648, opened 2018-09-17
id everywhere plus deref explanation

### LINKED ISSUE #1607: Just require id for everything and be done with it
Only embedded AnnotationPages do not require `id`, and instead they're SHOULD, as per http://iiif.io/api/presentation/3.0/#id

We should just make `id` a MUST for consistency.  The exception here is pointless and will catch implementers out.

azaroth42:
Closes #1607

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -342,22 +342,10 @@ The URI that identifies the resource. If the resource is only available embedded
 
 The value _MUST_ be a string, and the value _MUST_ be an HTTP(S) URI for resources defined in this specification. If the resource is retrievable via HTTP(S), then the URI _MUST_ be the URI at which it is published. External resources, such as profiles, _MAY_ have non-HTTP(S) URIs defined by other communities.
 
- * A Collection _MUST_ have the `id` property. <br/>
-   Clients _SHOULD_ render `id` on a Collection.
- * A Manifest _MUST_ have the `id` property.<br/>
-   Clients _SHOULD_ render `id` on a Manifest.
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -342,22 +342,10 @@ The URI that identifies the resource. If the resource is only available embedded
...
-   Clients _MAY_ render `id` on content resources.
- * A Range _MUST_ have the `id` property.<br/>
-   Clients _MAY_ render `id` on a Range.
- * An Annotation Collection _MUST_ have the `id` property.<br/>
-   Clients _MAY_ render `id` on an Annotation Collection.
- * An Annotation Page _SHOULD_ have the `id` property.<br/>
-   Clients _MAY_ render `id` on an Annotation Page.
- * An Annotation _MUST_ have the `id` property.<br/>
-   Clients _MAY_ render `id` on an Annotation.
+The existence of an HTTP(S) URI in the `id` property does not mean that the URI _MUST_ be dereferencable. The [definitions of the Resources][prezi30-resource-structure] give guidance as to this, as does the [table in Appendix A][prezi30-appendixa].
```
jpstroop: > ... does not mean that the URI _MUST_ ...

Slightly weird use of an RFC _MUST_. I think a regular "must" would be less confusing, or replace with something simpler, like "The HTTP(S) URI given in the `id` property does not have to be dereferencable" or ...is not required to be dereferencable. 

(or RFC 6919 [_OUGHT TO_](https://tools.ietf.org/html/rfc6919#section-4)) 😉

azaroth42: :+1: Fixed

Issue #1606, opened 2018-05-09
Recommend application/ld+json for Image info.json


azaroth42:
In Presentation 3.0 we recommend application/ld+json, with a fallback if necessary to application/json.  We should do the same in Image 3.0: http://iiif.io/api/image/3.0/#51-image-information-request

urrape:
please do it!!! json-ld is the way

zimeon:
I agree that Image 3.0 should have he same recommendation as Presentation 3.0

zimeon:
Discussed on [2018-08-29 Technical Community Call](https://docs.google.com/document/d/1Rx-sUh2STNf6_GxRC1ghqsEuoj5LymRDsxi0_WpqCJ4/edit) - consensus in favor of `application/ld+json`

azaroth42:
Closed by #1652

PR #1654, opened 2018-09-18
Note other parameters allowed

### LINKED ISSUE #1587: Image API: more specific content type
At http://iiif.io/api/image/3.0/#image-information-request it is said that the content type must be either 
`application/json` or `application/ld+json` but the example includes a content type parameter:

    Content-Type: application/ld+json;profile="http://iiif.io/api/image/3/context.json"

It might be made explicit that content type parameters are optional (e.g. JSON can be extended with `charset=utf-8`) and that `profile="http://iiif.io/api/image/3/context.json` is RECOMMENDED parameter for JSON-LD mime type.

zimeon:
Closes #1587 when combined with PR #1652 (for issue #1606)

### DIFF
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -486,7 +486,7 @@ Content-Type: application/ld+json;profile="http://iiif.io/api/image/3/context.js
 ```
 {: .urltemplate}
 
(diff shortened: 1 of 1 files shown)


zimeon:
@tomcrane we have the hyphenated version earlier in the same section so I followed that style

PR #1655, opened 2018-09-18
Improve wording in 5.1. Image Information Request

### LINKED ISSUE #1591: Clarify CORS statement regarding Image Information Requests
In [5.1. Image Information Request](http://iiif.io/api/image/3.0/#image-information-request) we have:

> Servers should also support CORS on image information responses.

which I think should be:

> Servers SHOULD support CORS on image information responses.

to parallel the earlier _Servers should support CORS on image responses._. The "also" doesn't make sense to me in the current context.

zimeon:
Closes #1591


zimeon:
Well spotted @tomcrane , fixed

PR #1656, opened 2018-09-18
Recommend support for HEAD in Image and Presentation 3.0

### LINKED ISSUE #1609: Consider HEAD as part of protocol requirements
@mejackreed on Slack:
As a Presentation API harvester, it might be good to know up front with a HEAD request what the headers for a response might be. Could there be a `SHOULD` guidance that Presentation API providers respond to `HEAD` requests? https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4

(Same applies to info.json, but putting it in Prezi milestone)

zimeon:
Closes #1609

### DIFF
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -814,6 +814,8 @@ A recipe for setting this header on the Apache HTTP Server is shown in the [Apac
 
 ## 7. Server Responses
 
+Servers _MUST_ support the HTTP GET method for retrieval of Image API resources. Servers _SHOULD_ support the HTTP OPTIONS method as part of the [CORS][image30-cors-response] preflight request pattern, and it is _RECOMMENDED_ that implementations also support the HTTP HEAD method.
+
 ### 7.1. CORS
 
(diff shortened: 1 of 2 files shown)


source/api/image/3.0/index.md
```diff
@@ -814,6 +814,8 @@ A recipe for setting this header on the Apache HTTP Server is shown in the [Apac
 
 ## 7. Server Responses
 
+Servers _MUST_ support the HTTP GET method for retrieval of Image API resources. Servers _SHOULD_ support the HTTP OPTIONS method as part of the [CORS][image30-cors-response] preflight request pattern, and it is _RECOMMENDED_ that implementations also support the HTTP HEAD method.
```
azaroth42: Both SHOULD and RECOMMENDED (that mean the same thing) in the same sentence reads a little strange to me. Not a -1 but how about ...
> ... and SHOULD also support the HTTP HEAD method.

And then a little clause explaining why (e.g. to allow clients to test for the existence of the resource).

azaroth42: Discussed, not really necessary.

PR #1658, opened 2018-09-18
1615 poster canvas

### LINKED ISSUE #1605: posterCanvas use is ambiguous for audio only
Should the audio of the poster be played while the main content is buffering, or should it be played as ambient content while the user interacts with the main canvases?

Seems like we need a behavior on the canvas? Or ( 😖 ) two separate properties?

### LINKED ISSUE #1615: posterCanvases can have posterCanvases
For canvas level fractal generation while loading.

Should be disallowed.

azaroth42:
Closes #1615 ; Closes #1605

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -301,32 +301,62 @@ The value _MUST_ be an [XSD dateTime literal][org-w3c-xsd-datetime]. The value _
 { "navDate": "2010-01-01T00:00:00Z" }
 ```
 
-##### posterCanvas
+##### placeholderCanvas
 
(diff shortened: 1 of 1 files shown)


azaroth42:
Note that this does NOT touch the tables, to prevent collisions with @tomcrane's forthcoming PRs on them.

azaroth42:
:+1: to leaving discussion of `together` to the cookbook.

tomcrane:
I agree that placeHolder might _occasionally_ be a typo - but it is a widely used English word, like `thumbnail` or `homepage`.

tomcrane:
I think it less likely to be wrongly capitalised than `homepage`.

PR #1664, opened 2018-09-18
Rebase image-prezi-rc2 on current master to get HTML fixes

zimeon:
It seems useful to have valid travis warnings as we work on `image-prezi-rc2`


jpstroop:
(Rebased again)

PR #1668, opened 2018-09-19
Update ffi version

zimeon:
There is a warning for ffi 1.9.23 which is addressed by updating to 1.9.24. I'm not sure why this was pinned but from a look at various of the API docs I don't see anything broken.

I think this should be merged into `master` and then we can later update `image-prezi-rc2`


zimeon:
Preview: https://preview.iiif.io/api/ffi-update/api/

azaroth42:
Merging on 4 as just infrastructure

Issue #1680, opened 2018-10-02
Paged behavior should be valid on Collections with multi-part behavior


azaroth42:
A multi-part collection, as a way to wrap up bound volumes of more than one manifest into a single resource, should be allowed to have the paged behavior to distinguish from the otherwise default behavior of individuals.

azaroth42:
Closed by #1681

Issue #1679, opened 2018-10-02
Unordered behavior should be valid on Collections


azaroth42:
If there are use cases for unordered manifests and ranges, then there are use cases for unordered collections. Here is a set of objects, rather than a list of objects.

Issue #1643, opened 2018-08-29
Interactions between behaviors


zimeon:
What should the Presentation API spec say regarding interactions between multiple [`behavior`](http://iiif.io/api/presentation/3.0/#behavior) values on a resource (either directly specified or inherited)? Should there be documentation so that clients handle non-obvious interactions consistently (e.g. `repeat` with and without `auto-advance` cases are currently described)?

zimeon:
Discussed on [2018-08-29 Technical Community Call](https://docs.google.com/document/d/1Rx-sUh2STNf6_GxRC1ghqsEuoj5LymRDsxi0_WpqCJ4/edit) in relation to #1612 (Inheritance of behaviors).

Two alternative approaches were suggested:
 1. clarify individual definitions but don't define interactions, thus behavior might be client dependent
 2. define the set of possible interactions (note permutation issue) for more consistent client behavior

A suggestion was that in 3.0 we might adopt 1) with a warning that clarifications of interactions (2) are expected in later revisions. It was agreed that such clarification in a 3.x would not be a breaking change.

azaroth42:
As a proposal for unblocking this issue:

* In the specification we note any disjoint behaviors (e.g. a Manifest cannot be both paged and continuous). This would be testable and allow for error conditions to be generated.
* Interaction and the resulting rendering / user interface implications between the permissible permutations are client dependent, but recommendations are made in the algorithms document. 
* Add a warning that further clarifications are to be expected in subsequent minor releases.

azaroth42:
Closed by #1681

Issue #1632, opened 2018-06-26
auto-advance doesn't allow parts of canvases to advance


azaroth42:
A canvas might represent a long stretch of content, such as a tape recording of oral histories with several histories being present.  Auto-advance should allow the particular history to be pieced together from the end of the segment of the canvas, rather than only taking effect at the end of a canvas.

Thus the point at which the play-head advances is determined by the encapsulating resource (e.g. Range) not the Canvas with content, contrary to the definition [here](http://iiif.io/api/presentation/3.0/#behavior).

workergnome:
I worry that this (and `repeat`) are going to create no end of edge cases, because they're describing event-based behaviors, not interface configuration behaviors.  Because of that, we suddenly need to worry about inheritance and precedence, and bubbling, and all the other event-handling stuff that every event definition system needs to worry about.  

Could we not meet this same need by using canvas-on-canvas annotations, with a `loop` if repeat is needed?

azaroth42:
Along with #1612 about inheritance of behaviors, is @workergnome's suggestion (get rid of auto-advance and just use a single canvas with a longer duration) an easy way out of several thorny issues?

tomcrane:
It would be an easy way out, but we would lose some of the power of the model to present the object.

_As a sound archive, it is important that I can convey in the model the distinct physical aspects of the real world object, as well as making it easy to navigate and experience for the web user._

Views vs structures at work for audio - each tape side in this long recording is a canvas:

![view-nav-large](https://user-images.githubusercontent.com/1443575/43100878-9badda9a-8ebe-11e8-8621-0bb5905f79c8.jpg)

But they are auto-advance canvases; they keep playing. We want that continuity of sound, but we also want to convey that there are a whole load of tape sides here.

I've added some notes about this issue to this document:

https://docs.google.com/document/d/1ad4m48FEVUBSuKf8dkwG0_M10kQnvM9g6qVShvMuO8s/edit#heading=h.3wwqry1ozgsw

The heading "Problems with auto-advance" is the target of that link - I didn't want to dump it all here. 

[...]

PR #1681, opened 2018-10-02
behavior interactions

### LINKED ISSUE #1612: inheritance of behaviors across resources
In presentation 2, as a viewer, if I come across a manifest with a `viewingDirection`, I should take that as an intention for the canvases to be displayed accordingly, even if there are multiple sequences. If multiple sequences, each with different `viewingDirection` properties, they trump the Manifests `viewingDirection`.

In a way, sequences by default inherit the `viewingDirection` from the Manifest. Is this true for ranges? If a top level Manifest has a `viewingDirection` of `right-to-left`, should that be inherited by each of the Ranges (if they don't have a `viewingDirection` themselves) or should they default to `left-to-right`. 

Further, in presentation 3, if sequences are now represented as Ranges, their behaviour here will change, and any conversion between presentation 2 and presentation 3 would have to explicitly set the viewing direction, if it was previously implied by the Manifest having the property (if Ranges should not inherit this direction).

azaroth42:
Closes #1612, #1632, #1643, #1679, #1680! 

Preview: https://preview.iiif.io/api/1612-behaviors/api/presentation/3.0/#behavior


source/api/presentation/3.0/index.md
```diff
@@ -531,27 +531,48 @@ The value _MUST_ be a string.
...
+
+The descriptions of the behavior values have a set of which other values they are disjoint with, meaning that the same resource _MUST NOT_ have both of two or more from that set. In order to determine which is in effect, the client _SHOULD_ follow the inheritance rules above, taking the value from the closest resource. The user interface effects of the possible permutations of non-disjoint behavior values are client dependent, and implementers are advised to look for [relevant recipes][annex-cookbook] in the IIIF cookbook.
```
zimeon: Seems odd to hyperlink "relevant recipes" instead of "IIIF cookbook" -- I would expect the former to take me to the relevant ones.

(otherwise looks good)

azaroth42: True! Fixed :)

azaroth42:
@iiif/editors Ping?

source/api/presentation/3.0/index.md
```diff
@@ -531,27 +531,48 @@ The value _MUST_ be a string.
...
 
+In order to determine the behaviors that are governing a particular resource, there are four inheritance rules:
+* Collections inherit behaviors from their parent Collection.
+* Manifests **DO NOT** inherit behaviors from any parent Collections.
+* Canvases inherit behaviors from their parent Manifest, but **DO NOT** inherit behaviors from any parent Ranges.
+* Ranges inherit behaviors from their parent Range and parent Manifest.
+
+When interpreting the Layout Behaviors and `auto-advance` for Ranges, they govern the rendering of the resources when that particular Range is selected but are not always applied.  A Range with the `behavior` value `continuous`, in a Manifest with the `behavior` value `paged`, would mean that the Manifest's Canvases should be rendered in a paged fashion, unless the range is selected to be viewed, and its included Canvases would be rendered in that context only as being virtually stitched together. This might occur, for example, when a physical scroll is cut into pages and bound into a codex with other pages, and the publisher would like to provide the user the experience of the scroll in its original form.
```
tomcrane: I find the first sentence very hard to scan. Not sure this version is any better...

> Clients should interpret behaviors on a Range only when that Range is selected or is in some other way the context for the user's current interaction with the resources. At other times, behaviors on a Range may not apply to the current interaction because the user isn't interacting with the resources in the context of that Range.
> A Range with the `behavior` value `continuous`, in a Manifest with the `behavior` value `paged`, means that the Manifest's Canvases should be rendered in a paged fashion, _unless that Range is selected to be viewed_. Its included Canvases would be rendered as being virtually stitched together in that context only. This might occur, for example, when a physical scroll is cut into pages and bound into a codex with other pages, and the publisher would like to provide the user the experience of the scroll in its original form. 


This whole bit of spec introduces the notion of state history - how did the user come to be looking at resource X? Which is maybe unavoidable but makes me uneasy.

source/api/presentation/3.0/index.md
```diff
@@ -531,27 +531,48 @@ The value _MUST_ be a string.
...
-| `unordered` | Valid on Manifests and Ranges. The Canvases included in resources that have this behavior have no inherent order, and user interfaces _SHOULD_ avoid implying an order to the user. |
+|| **Temporal Behaviors** |
+| `auto-advance` | Valid on Collections, Manifests, Canvases, and Ranges that include or are Canvases with at least the `duration` dimension. When the client reaches the end of a Canvas, or segment there-of as specified in a Range, with a duration dimension that has this behavior, it _SHOULD_ immediately proceed to the next Canvas or segment and render it. If there is no subsequent Canvas in the current context, then this behavior should be ignored. When applied to a Collection, the client should treat the first Canvas of the next Manifest as following the last Canvas of the previous Manifest, respecting any `start` property specified. Disjoint with `no-auto-advance`. |
```
tomcrane: thereof rather than there-of?

source/api/presentation/3.0/index.md
mikeapp: I think splitting the first sentence as above helps with clarity.

source/api/presentation/3.0/index.md
```diff
@@ -531,27 +531,48 @@ The value _MUST_ be a string.
 
 A set of user experience features that the publisher of the content would prefer the client to use when presenting the resource. This specification defines the values in the table below. Others may be defined externally as an [extension][prezi30-ldce].
 
+In order to determine the behaviors that are governing a particular resource, there are four inheritance rules:
+* Collections inherit behaviors from their parent Collection.
+* Manifests **DO NOT** inherit behaviors from any parent Collections.
+* Canvases inherit behaviors from their parent Manifest, but **DO NOT** inherit behaviors from any parent Ranges.
```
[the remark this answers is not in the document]
azaroth42: Changed all "parent" to "referencing" and expanded the intro sentence to the list.

source/api/presentation/3.0/index.md
azaroth42: I think it's unavoidable.  Agree the first sentence was cumbersome.  Not sure the second proposed sentence is useful though, as if Canvases don't inherit behaviors from Ranges, so if you're not interacting with a Range, there are no resources that have inherited behavior.

Replace first sentence with above first sentence.

source/api/presentation/3.0/index.md

azaroth42:
@IIIF/editors Ping ping?

source/api/presentation/3.0/index.md
zimeon: confirm fixed, resolving convo

PR #1695, opened 2018-10-17
Escape footnote asterisks with backslash

tomcrane:
This seems to keep GitHub happy


azaroth42:
Non normative, merging

PR #1699, opened 2018-10-17
Add provider, change logo to 1:1

### LINKED ISSUE #1639: Label for the organization that has the logo
The Presentation API has a `logo` property, with the definition in [2.1](http://iiif.io/api/presentation/2.1/#logo):  
> A small image that represents an individual or organization associated with the resource it is attached to. This could be the logo of the owning or hosting institution.

And almost identical in [3.0](http://iiif.io/api/presentation/3.0/#logo).

So there is the capability to associate a logo of an actor, but not even a label for the actor that has the logo. The label of the logo would be "Logo of X" not just "X".  The attribution/requiredStatement might not be for the institution that has the logo, e.g. the requiredStatement could say "Purchased with the generous donation of J Smith." whereas the logo is for the museum.  Further, the cardinality of logo is 1-many, from the discussion in #664. Even if there were multiple institutionLabels and multiple logos, you still couldn't link them together.  

The logo is at the same level of presentation as `summary` and `thumbnail` -- when the manifest is being summarized in a list, such as a Collection or a search results list.  This is implemented in Mirador (for example) in the "loading window" via a method that's not based on the specifications. It is useful in any context when the resource is used outside of the responsible institution's context. This would include any search/discovery environment, of which the Mirador load window is a preconfigured microcosm.

Given that the only use of `logo` can be for an institution, it seems a very small step to actually instantiate the identity of that institution in the data.  

This is the right time to discuss, as it would be a breaking change for both Presentation and Image. It does not add significant complexity, while solving several use cases from discovery and adding accessibility to the intended use of the existing `logo` property.  The requirement came up in the Discovery TSG on 2018-07-11 with [several use cases being described](https://github.com/IIIF/discovery/issues/37).

Proposal (from Rob, not Discovery TSG):

Where ever we currently use `logo`, instead replace with a property that references a Person or Organization, which in turn has `id`, `type`, `label`, and `logo`.  It would have the same semantics as `logo` does currently -- an individual or organization associated with the resource.

### LINKED ISSUE #1698: Logo should be exactly 1:1 not 1:*
Once `logo` is associated with the Agent (which is 1:\*), it should instead be 1:1. There's no need to have multiple logo images associated per Agent. Previously it needed to be 1:\* as there wasn't the Agent resource between the resource and the logo.

azaroth42:
Closes #1639 
Closes #1698 

Preview: https://preview.iiif.io/api/1639-provider/api/presentation/3.0/#provider

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -265,6 +265,57 @@ The value _MUST_ be a string. If the value is drawn from Creative Commons or Rig
 { "rights": "https://creativecommons.org/licenses/by/4.0/" }
 ```
 
+##### provider
+
+An organization or person that contributed to providing the content of the resource. Clients can then display this information to the user to acknowledge the provider's contributions.  This differs from the `requiredStatement` property, in that the data is structured, allowing the client to do more than just present text but instead have richer information about the people and organizations to use in different interfaces.
+
+The organization or person is represented as an `Agent` resource.  
+
+* Agents _MUST_ have the `id` property, and its value _MUST_ be a string that contains a URI that identifies the agent.
+* Agents _MUST_ have the `type` property, and its value _MUST_ be the string "Agent".
+* Agents _MUST_ have the `label` property, and its value _MUST_ be a JSON object as described in the [languages][prezi30-languages] section.
+* Agents _SHOULD_ have the `homepage` property, and its value _MUST_ be a JSON object as described in the [homepage][prezi30-homepage] section.
+* Agents _SHOULD_ have the `logo` property, and its value _MUST_ be an array of JSON objects as described in the [logo][prezi30-logo] section.
+* Agents _MAY_ have the `seeAlso` property, and its value _MUST_ be an array of JSON object as described in the [seeAlso][prezi30-seealso] section.
+
+The value _MUST_ be an array of JSON objects, where each item in the array conforms to the structure of an Agent, as described above.
+
+* A Collection _SHOULD_ have the `provider` property with at least one item. <br/> Clients _MUST_ render `provider` on a Collection.
+* A Manifest _SHOULD_ have the `provider` property with at least one item. <br/> Clients _MUST_ render `provider` on a Manifest.
+* Other resource types _MAY_ have the `provider` property with at least one item. <br/> Clients _SHOULD_ render `provider` on other resource types.
+
+```json-doc
+{
+  "provider": [
+    {
+      "id": "https://example.org/about",
+      "type": "Agent",
+      "label": {"en": ["Example Organization"]},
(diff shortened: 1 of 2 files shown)


azaroth42:
Ping @jpstroop @tomcrane

source/api/presentation/3.0/index.md
```diff
@@ -607,15 +658,22 @@ The value _MUST_ be a JSON object, which _MUST_ have the `id`, `type`, and `labe
 
 ##### logo
 
-A small external image resource that represents an individual or organization associated with the resource that has the `logo` property. This could be the logo of the owning or hosting institution. The logo _MUST_ be clearly rendered when the resource is displayed or used, without cropping, rotating or otherwise distorting the image. It is _RECOMMENDED_ that a [IIIF Image API][image-api] service be available for this image for other manipulations such as resizing.
+A small external image resource that represents the Agent resource it is associated with. The logo _MUST_ be clearly rendered when the resource is displayed or used, without cropping, rotating or otherwise distorting the image. It is _RECOMMENDED_ that a [IIIF Image API][image-api] service be available for this image for other manipulations such as resizing.
 
-The value _MUST_ be an array of JSON objects, each of which _MUST_ have an `id` and _SHOULD_ have at least one of `type` and `format`.
+The value _MUST_ be a JSON objects, which _MUST_ have an `id` and _SHOULD_ have at least one of `type` and `format`.
```
tomcrane:
```suggestion
The value _MUST_ be a JSON object, which _MUST_ have an `id` and _SHOULD_ have at least one of `type` and `format`.
```

azaroth42:
Missing update to the summary tables - DNM

source/api/presentation/3.0/index.md
```diff
@@ -265,6 +265,57 @@ The value _MUST_ be a string. If the value is drawn from Creative Commons or Rig
 { "rights": "https://creativecommons.org/licenses/by/4.0/" }
 ```
 
+##### provider
+
+An organization or person that contributed to providing the content of the resource. Clients can then display this information to the user to acknowledge the provider's contributions.  This differs from the `requiredStatement` property, in that the data is structured, allowing the client to do more than just present text but instead have richer information about the people and organizations to use in different interfaces.
```
zimeon: should "can then" be _MAY_ ?

azaroth42: No, it's SHOULD in the normative part for Collection and Manifest.  "can" was intentionally chosen to NOT seem like an RFC word.

zimeon: I guess I'm stuck with the old fashioned British permission/ability distinction between may/can ;-) I don't see a simple change so I'll :+1:

PR #1704, opened 2018-10-30
Add 4.1 for case sensitivity

### LINKED ISSUE #1703: Please clarify case sensitive of types (and more generally for property values on which data model depends?)
I pose the question: Are the values for model-dependent properties, e.g. "type", generally case sensitive or not? e.g. @type":"sc:Manifest" == "SC:manifest" ???

azaroth42:
Closes #1703


jpstroop:
I'm 👍🏼, but do we really need the example? Seems like it adds unnecessary length. I think this could be two sentences: 

> (JSON-LD is case-sensitive)[reference?]. Valid IIIF server responses _MUST_ match the cases used throughout this specification.

azaroth42:
https://www.w3.org/TR/json-ld11/#the-context : 
> Simply speaking, a context is used to map terms to IRIs. **Terms are case sensitive** and any valid string that is not a reserved JSON-LD keyword can be used as a term. 

(Emphasis added) Can add a link to that section if you think it's useful? We don't tend to link inherited requirements, but I don't think there was ever a conscious decision not to.

Agree about the example, it's pretty lame. Put it in for consistency with all the other sections, happy to take it out.

azaroth42:
:+1: to zimeon's rewording. Fixed.

tomcrane:
Stale review not dismissed but this is a re-approval anyway.

PR #1714, opened 2018-11-12
Reorder description of conneg and content types in Image and Prezi 3.0

### LINKED ISSUE #1685: JSON-LD and content negotation
Currently trying to implement the Image API v2 (and will tackle v3 later). I'm getting a bit confused by the wording of the `Content-Type` for image info requests (refs https://github.com/elifesciences/fastly-iiif/pull/3). 

In 2.0.0/2.1.1 it's:

> If the client explicitly wants the JSON-LD content-type, then it *must* specify this in an Accept header, otherwise the server *must* return the regular JSON content-type.

This reads to me as if `application/ld+json` _has_ to be present in an `Accept` header, eg `Accept: */*` (which is the same as no `Accept` header) isn't enough. (So the same as the [recommendation for WebP content negotiation](https://developers.google.com/speed/webp/faq#how_can_i_detect_browser_support_for_webp), support has to be explicitly advertised.)

(I note it doesn't mention the potential of `application/ld+json` having a [quality value of 0, which means 'not acceptable'](https://tools.ietf.org/html/rfc7231#section-5.3.1).)

But it seems like (at least some) existing implementations _do_ return `application/ld+json` in this case (see https://github.com/elifesciences/fastly-iiif/pull/3#discussion_r223418490)

The wording in 3.0.0-alpha is:

>  If the server receives a request with one of the content types above in the Accept header, it *should* respond with that content type following the rules of content negotiation. Otherwise, it *must* respond with the `application/json` content type.

Which now mentions content negotiation, but still seems to mean that `application/ld+json` has to be present in an `Accept` header.

https://github.com/IIIF/api/pull/1652 has been added to try and promote `application/ld+json`, but doesn't seem to make much difference: `application/ld+json` still has to be present in an `Accept` header.

(I don't use the other APIs, but they seem to have similar wording.)

To help clarify what should be happening, what is the expected `Content-Type` for the the following `Accept` header values in both image API v2 and v3 image requests when the `jsonldMediaType` feature is available?

1. No header
1. `*/*`
1. `application/*`
1. `application/ld+json`, `*/*`
1. `application/json, */*`
1. `*/*; q=1, application/ld+json; q=0.1`
1. `*/*; q=1, application/ld+json; q=0`

zimeon:
Closes #1685

### DIFF
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -518,16 +518,16 @@ The request for the image information _MUST_ conform to the URI template:
 ```
 {: .urltemplate}
 
-The syntax for the response is [JSON-LD][org-w3c-json-ld]. If the server receives a request with an `Accept` header, it _SHOULD_ respond following the rules of [content negotiation][org-rfc-7231-conneg]. Note that content negotiation requests _MAY_ include `profile` and other parameters with any content type. 
+The syntax for the response is [JSON-LD][org-w3c-json-ld]. If the server receives a request with an `Accept` header, it _SHOULD_ respond following the rules of [content negotiation][org-rfc-7231-conneg]. Note that content types provided in the `Accept` header of the request _MAY_ include parameters, for example `profile` or `charset`.
 
-If the request does not include an `Accept` header, or that header does not include supported content types, the HTTP `Content-Type` header of the response _SHOULD_ have the value `application/ld+json` with the `profile` parameter given as the context document: `http://iiif.io/api/image/3/context.json`.
+If the request does not include an `Accept` header, the HTTP `Content-Type` header of the response _SHOULD_ have the value `application/ld+json` with the `profile` parameter given as the context document: `http://iiif.io/api/image/3/context.json`.
 
 ``` none
 Content-Type: application/ld+json;profile="http://iiif.io/api/image/3/context.json"
 ```
 {: .urltemplate}
 
-If content type `application/ld+json` cannot be generated due to server configuration details, then the `Content-Type` header _SHOULD_ instead be `application/json` (regular JSON), without a `profile` parameter.
+If the `Content-Type` header `application/ld+json` cannot be generated due to server configuration details, then the `Content-Type` header _SHOULD_ instead be `application/json` (regular JSON), without a `profile` parameter.
 
 ``` none
 Content-Type: application/json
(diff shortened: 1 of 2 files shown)


source/api/image/3.0/index.md
```diff
@@ -518,21 +518,22 @@ The request for the image information _MUST_ conform to the URI template:
 ```
 {: .urltemplate}
 
-The syntax for the response is [JSON-LD][org-w3c-json-ld]. The HTTP `Content-Type` header of the response _SHOULD_ have the value `application/ld+json` with the `profile` parameter given as the context document: `http://iiif.io/api/image/3/context.json`.
+The syntax for the response is [JSON-LD][org-w3c-json-ld]. If the server receives a request with an `Accept` header, it _SHOULD_ respond following the rules of [content negotiation][org-rfc-7231-conneg]. Note that content negotiation requests _MAY_ include `profile` and other parameters with any content type.
```
azaroth42: Note that the content type provided in the `Accept` header of the request _MAY_ include parameters, for example `profile` or `charset`.  

?

source/api/image/3.0/index.md
```diff
@@ -518,21 +518,22 @@ The request for the image information _MUST_ conform to the URI template:
...
+
+If the request does not include an `Accept` header, or that header does not include supported content types, the HTTP `Content-Type` header of the response _SHOULD_ have the value `application/ld+json` with the `profile` parameter given as the context document: `http://iiif.io/api/image/3/context.json`.
 
 ``` none
 Content-Type: application/ld+json;profile="http://iiif.io/api/image/3/context.json"
 ```
 {: .urltemplate}
 
-If this cannot be generated due to server configuration details, then the `Content-Type` header _MUST_ instead be `application/json` (regular JSON), without a `profile` parameter.
+If content type `application/ld+json` cannot be generated due to server configuration details, then the `Content-Type` header _SHOULD_ instead be `application/json` (regular JSON), without a `profile` parameter.
```
azaroth42: If the `Content-Type` value `application/ld+json` cannot be generated ... ?

source/api/image/3.0/index.md
zimeon: Agreed, with content types (plural)

source/api/image/3.0/index.md
zimeon: :+1:

PR #1772, opened 2019-01-29
Remove leftover redirect from discovery/0.1

### LINKED ISSUE #1770: Change Discovery API permalink redirects to 0.1
I've just noticed that the Change Discovery API permalink (https://iiif.io/api/discovery/) redirects to 0.1 and not to 0.2. 

I know it is not a stable version but shouldn't we want to give access to the latest one?

zimeon:
The redirect for `/api/discovery/` was left in 0.1 as well as being present in 0.2. The PR removes from 0.1 so that the redirect is to 0.2. Fixes #1770

### DIFF
--- a/source/api/discovery/0.1/index.md
+++ b/source/api/discovery/0.1/index.md
@@ -9,9 +9,6 @@ major: 0
 minor: 1
 patch: 0
 pre: final
-redirect_from:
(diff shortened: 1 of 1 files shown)


jpstroop:
Non-normative. Will merge

PR #1789, opened 2019-03-19
discovery 0.3 update

azaroth42:
Add:
  * `partOf` to OrderedCollection (which will be scoped in, to avoid collisions) to refer to an aggregating OrderedCollection
 * Rename `context` back to `seeAlso` given that we have our own context again
 * Add section 3.4, describing `@context` and use of extensions, following Prezi3 pattern.
 * Add section 4, on protocol related matters, notably:
    * Negotiable resources
    * Authentication / network issues

Updated:
  * links, version number

### DIFF
--- a/source/api/discovery/0.3/index.md
+++ b/source/api/discovery/0.3/index.md
@@ -1,0 +1,781 @@
+---
+title: "IIIF Change Discovery API 0.3"
+title_override: "IIIF Change Discovery API 0.3"
+id: discovery-api
+layout: spec
+cssversion: 3
+tags: [specifications, presentation-api]
+major: 0
+minor: 3
+patch: 0
+pre: final
+redirect_from:
+  - /api/discovery/index.html
+  - /api/discovery/0/index.html
+---
+
+
+## Status of this Document
+{:.no_toc}
+__This Version:__ {{ page.major }}.{{ page.minor }}.{{ page.patch }}{% if page.pre != 'final' %}-{{ page.pre }}{% endif %}
+
+__Latest Stable Version:__ None
+
+__Previous Version:__ [0.2][discovery02]
+
+**Editors**
+
+  * **[Michael Appleby](https://orcid.org/0000-0002-1266-298X)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0002-1266-298X), [_Yale University_](http://www.yale.edu/)
+  * **[Tom Crane](https://orcid.org/0000-0003-1881-243X)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0003-1881-243X), [_Digirati_](http://digirati.com/)
+  * **[Robert Sanderson](https://orcid.org/0000-0003-4441-6852)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0003-4441-6852), [_J. Paul Getty Trust_](http://www.getty.edu/)
+  * **[Jon Stroop](https://orcid.org/0000-0002-0367-1243)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0002-0367-1243), [_Princeton University Library_](https://library.princeton.edu/)
+  * **[Simeon Warner](https://orcid.org/0000-0002-7970-7855)** [![ORCID iD]({{ site.url }}{{ site.baseurl }}/img/orcid_16x16.png)](https://orcid.org/0000-0002-7970-7855), [_Cornell University_](https://www.cornell.edu/)
+  {: .names}
+
+{% include copyright.md %}
+
+__Status Warning__
+This is a work in progress and may change without notice. Implementers should be aware that this document is not stable. Implementers are likely to find the specification changing in incompatible ways. Those interested in implementing this document before it reaches beta or release stages should join the IIIF [mailing list][iiif-discuss] and the [Discovery Specification Group][groups-discovery], take part in the discussions, and follow the [emerging issues][github-discovery-issues] on Github.
+{: .warning}
+
+----
+
+## Table of Contents
+{:.no_toc}
+
+* Table of Discontent (will be replaced by macro)
+{:toc}
+
+## 1. Introduction
+{: #introduction}
+
+The resources made available via the IIIF (pronounced "Triple-Eye-Eff") [Image][image-api] and [Presentation][prezi-api] APIs are only useful if they can be found. Users cannot interact directly with a distributed, decentralized ecosystem but instead must rely on services that harvest and process the available content, and then provide a user interface enabling navigation to that content via searching, browsing or other paradigms. Once the user has discovered the content, they can then display it in their viewing application of choice. Machine to machine interfaces are also enabled by this pattern, where software agents could interact via APIs to discover the same content and retrieve it for further analysis or processing.
+
+This specification leverages existing techniques, specifications, and tools in order to promote widespread adoption. It enables the collaborative development of global or thematic search engines and portal applications that allow users to easily find and engage with content available via existing IIIF APIs.
+
+### 1.1. Objectives and Scope
+{: #objectives-and-scope}
+
+The objective of the IIIF Change Discovery API is to provide the information needed to discover and subsequently make use of IIIF resources.  The intended audience is other IIIF aware systems that can leverage the content and APIs. While this work may benefit others outside of the IIIF community directly or indirectly, the objective of the API is to specify an interoperable solution that best and most easily fulfills the discovery needs within the community of participating organizations.
+
+The discovery of IIIF resources requires a consistent and well understood pattern for content providers to publish lists of links to their available content. This allows a baseline implementation of discovery systems that process the list, looking for resources that have been added or changed.
+
+This process can be optimized by allowing the content providers to publish descriptions of when their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes unavailable. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll all of the systems to see if anything has changed. 
+
+Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. The diverse domains represented within the IIIF community already have successful standards fulfilling these use cases, and the results of previous attempts to reconcile these standards across domains have not seen widespread adoption. Also out of scope is optimization of the transmission of content, for example recommendations about transferring any source media or data between systems.
+
+
+__Notification of Changes__<br>This draft version of the specification does not include the subscription mechanism for enabling change notifications to be pushed to remote systems.  The current specification only enables the polling pattern where the set of changes must be periodically reprocessed. Notifications are likely to be added in a future version before 1.0.
+{: .warning}
+
+### 1.2. Terminology
+{: #terminology}
+
+The specification uses the following terms:
+
+* __HTTP(S)__: The HTTP or HTTPS URI scheme and internet protocol.
+
+The terms _array_, _JSON object_, _number_, and _string_ in this document are to be interpreted as defined by the [Javascript Object Notation (JSON)][org-rfc-8259] specification.
+
+The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _SHOULD NOT_, _RECOMMENDED_, _MAY_, and _OPTIONAL_ in this document are to be interpreted as described in [RFC 2119][org-rfc-2119].
+
+
+## 2. Overview of IIIF Resource Discovery
+{: #overview-of-iiif-resource-discovery}
+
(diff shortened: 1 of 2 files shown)


azaroth42:
Preview link:  https://preview.iiif.io/api/discovery-0.3/api/discovery/0.3/

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+
+OrderedCollections _MAY_ have a `totalItems` property.  The value _MUST_ be a non-negative integer.
+
+```
+{ "totalItems": 21456 }
+```
+
+##### seeAlso
+
+This property is used to refer to one or more documents that semantically describe the set of resources that are being acted upon in the Activities within the OrderedCollection. This would allow the OrderedCollection to refer to, for example, a [DCAT][org-w3c-dcat] description of the dataset. For OrderedCollections that aggregate activities and/or objects from multiple sources, the referenced description should describe the complete aggregation rather than an individual source.
```
mikeapp: Use of "OrderedCollection" (no space) here seems inconsistent with other entries.

zimeon: Fixed by 00a403de623c4993618af80254e0d1c92ed6bb29

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+
+```
+{
+  "@context": "http://iiif.io/api/discovery/0/context.json"
+}
+```
+
+#### 3.4.2. Extensions
+
+If any additional classes or properties are desired beyond the ones defined in this specification or the ActivityStreams specification, then those classes or properties _SHOULD_ be mapped to RDF terms in one or more additional context documents. These extension contexts _SHOULD_ be added to the top level `@context` property, and _MUST_ be before the discovery context. The JSON-LD 1.1 functionality of defining terms only within a specific property, known as [scoped contexts][org-w3c-json-ld-scoped-contexts], _MUST_ be used to minimize cross-extension collisions.  Extensions intended for broad use _SHOULD_ be registered in the [extensions registry][annex-registry].
```
mikeapp: 'Discovery context` (capitalized) or even more verbose (e.g., "URI of..." as in 3.4.1)

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+## Table of Contents
+{:.no_toc}
+
+* Table of Discontent (will be replaced by macro)
+{:toc}
+
+## 1. Introduction
+{: #introduction}
+
+The resources made available via the IIIF (pronounced "Triple-Eye-Eff") [Image][image-api] and [Presentation][prezi-api] APIs are only useful if they can be found. Users cannot interact directly with a distributed, decentralized ecosystem but instead must rely on services that harvest and process the available content, and then provide a user interface enabling navigation to that content via searching, browsing or other paradigms. Once the user has discovered the content, they can then display it in their viewing application of choice. Machine to machine interfaces are also enabled by this pattern, where software agents could interact via APIs to discover the same content and retrieve it for further analysis or processing.
```
zimeon: Not wrong as is, but I think the first sentence is stronger/clearer with "...APIs are useful only if they can be found." (moving "only")

In last sentence suggest s/could/can/ to match preceding sentence.

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+
+This specification describes three levels of conformance that build upon each other in terms of functionality enabled and precision of the information published. Sets of changes are published in pages, which are then collected together in a collection per publisher. To reduce barriers to implementation, care has been take to allow the implementation of all levels using only static files on a web server, rather than requiring dynamic access to a database.
+
+### 2.1. Listing Resources and their Changes
+{: #listing-resources-and-their-changes}
+
+There are three levels of conformance at which changes can be described. Level 0 is simply a list of the resources available.  Level 1 adds timestamps and ordering from earliest change to most recent, allowing the consumer to stop processing once it encounters a change that it has already processed. Level 2 adds information about the types of activities, enabling the explicit creation and deletion of resources.
```
zimeon: the types don't enable creation and deletion, just clear description of these actions. I suggest:
> < Level 2 adds information about the types of activities, enabling the explicit creation and deletion of resources.
>
> \> Level 2 adds information about the types of activities, enabling explicit description of the creation and deletion of resources.

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+    "type": "Manifest"
+  },
+  "endTime": "2017-09-20T00:00:00Z"
+}
+```
+
+#### 2.1.3. Level 2: Complete Change List
+{: #level-2-complete-change-list}
+
+At the most detailed level, a log of all of the Activities that have taken place can be recorded, with the likelihood of multiple Activities per IIIF resource.  This allows the additional types of "Create" and "Delete", enabling a synchronization process to remove resources as well as add or update them.
```
zimeon: "This allows" isn't clear. I suggest second sentence:

| \> Use of the additional types of "Create" and "Delete" allows explicit description of creations and deletions, enabling a synchronization process to remove resources as well as add or update them.

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+        "format": "application/ld+json"
+      }
+    ] 
+  }
+}
+```
+
+##### endTime
+
+The time at which the Activity was finished. It is up to the implementer to decide whether the Activity includes the publication of the IIIF resource online, or only the internal data modification, but the decision _MUST_ be consistently applied. The changed resource given in `object` _MUST_ be available at its URI at or before the timestamp given in `endTime`. The value of `endTime` _SHOULD_ be before the time that the Activity is published as part of its Ordered Collection.
```
zimeon: How can an implementer only do "the internal data modification" (not publish it) _AND_ have "changed resource given in `object` _MUST_ be available at its URI at or before the timestamp given in `endTime`" ? I would strike the second sentence.

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+  }
+}
+```
+
+
+### 3.4. Linked Data Context and Extensions
+
+#### 3.4.1. @context
+
+The top level resource in the response _MUST_ have the `@context` property, and it _SHOULD_ appear as the very first key/value pair of the JSON representation.  This property lets Linked Data processors interpret the document as a graph.  The value of the property _MUST_ be either the URI of the IIIF Discovery context document, `http://iiif.io/api/discovery/0/context.json`, or it _MUST_ be an array of strings, where the URI of the IIIF Discovery context document is the last item in the array.
```
zimeon: Awkward non-parallel construction: .. _MUST_ be either BLAH1, or it _MUST_ be BLAH2 -- remove second "it _MUST_ be" because "_MUST_ be either BLAH1 or BLAH2" is fine and much easier to read (cf. https://preview.iiif.io/api/image-prezi-rc2/api/presentation/3.0/#46-linked-data-context-and-extensions)

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+
+<ol>
+  <li>Initialization:
+    <ol>
+      <li>Let <code class="highlighter-rouge">processedItems</code> be an empty array</li>
+      <li>Let <code class="highlighter-rouge">lastCrawl</code> be the timestamp of the previous time the algorithm was executed</li>
+    </ol>
+  </li>
+  <li>Retrieve the representation of <code class="highlighter-rouge">collection</code> via HTTP(S)</li>
+  <li>Minimally validate that it conforms to the specification</li>
```
zimeon: Wondering what "Minimally validate" means... "At least validate" or "Partially validate"? Would striking "Minimally" be clearer?

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+</ol>
+
+#### 3.5.2. Page Algorithm
+{: #page-algorithm}
+
+Given the URI of an ActivityStreams CollectionPage (`page`) and the date of last crawling (`lastCrawl`) as input, a conforming processor _SHOULD_:
+
+<ol class="ordered-list">
+  <li>Retrieve the representation of <code class="highlighter-rouge">page</code> via HTTP(S)</li>
+  <li>Minimally validate that it conforms to the specification</li>
```
zimeon: Minimally, see my comment above

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+#### 3.5.3. Indexing
+{: #indexing}
+
+If the objective of the consuming application is to find descriptive information that might be used to build an index allowing the resources to be discovered, then the application _SHOULD_ use the IIIF Presentation API `seeAlso` property to discover an appropriate, machine-readable description of the resource.  For different types of resource, and for different domains, the external resources will have different formats and semantics. If there are no external descriptions, or none that can be processed, the data in the Manifest and in other IIIF resources might be used as a last resort, despite its presentational intent. 
+
+## 4. Network Considerations
+
+### 4.1. Activities for Access-Restricted Content
+
+Activities _MAY_ be published about content that has access restrictions. Clients _MUST NOT_ assume that they will be able to access every resource that is the object of an Activity, and _MUST NOT_ assume that if they cannot access it, then it has been permanently deleted.  The content might be protected by some authentication system, or there might simply be a temporary network outage.
```
zimeon: Do we mean "permanently deleted" or simply "deleted"?

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+
+##### type
+
+The type of Activity. 
+
+This specification uses the types described in the table below.
+
+| Type   | Definition |
+| ------ | ---------- |
+| `Create` | The initial creation of the resource.  Each resource _MUST_ have at most one `Create` Activity in which it is the `object`. |
```
zimeon: Is it intentionally illegal to have a `Create` -> `Update` -> `Delete` -> `Create` -> `Update` sequence for a resource? (per max one `Create`, max one `Delete` two lines below). Maybe this is not an editorial question for now but I wonder whether this is a useful/necessary restriction on AS

source/api/discovery/0.3/index.md
azaroth42: :+1:

source/api/discovery/0.3/index.md
azaroth42: :+1:

source/api/discovery/0.3/index.md
azaroth42: :+1:

source/api/discovery/0.3/index.md
azaroth42: :+1:

source/api/discovery/0.3/index.md
azaroth42: Sentence is opaque, I agree.
Clarified, I hope, to: 
> ... whether the value of `endTime` is the timestamp for the publication of the IIIF resource online or is the timestamp of the modification to the data in the managing system if these are different,  ...

source/api/discovery/0.3/index.md
azaroth42: :+1:

source/api/discovery/0.3/index.md
azaroth42: It's intended as a warning to validate the input at least such that it's JSON and approximately correct structure. It could be omitted completely.  But no need to validate all the extensions, and so forth. Replaced with:

> Validate that the retrieved representation contains at least the features required for processing

source/api/discovery/0.3/index.md
azaroth42: Ditto. :)

source/api/discovery/0.3/index.md
azaroth42: The second sentence is the important one. The final sentence is intended as examples for when content might be inaccessible but not gone.  Added For example and messed a bit with the sentence.

source/api/discovery/0.3/index.md
azaroth42: Will make an issue for this, as I think it's normative rather than just editorial.

source/api/discovery/0.3/index.md
```diff
@@ -0,0 +1,784 @@
...
+      "endTime": "2018-03-10T10:00:00Z"
+    }
+  ]
+}
+```
+
+### 3.3. Activities
+{: #activities}
+
+The Activities are the means of describing the changes that have occured in the content provider's system.
```
mikeapp: -> "occurred"

azaroth42: I cannot spell that word. Ever. Fixed.

source/api/discovery/0.3/index.md
zimeon: :+1:

source/api/discovery/0.3/index.md
zimeon: But I think the point is simply that you can't take it that the resource has been deleted -- permanently, temporarily, or otherwise.

source/api/discovery/0.3/index.md
zimeon: Done, https://github.com/IIIF/discovery/issues/50

source/api/discovery/0.3/index.md
azaroth42: Oh, I see. Sure... the "permanently" was supposed to give some justification to why the crawler might think to delete it from the index, but I can just delete (ha) the word.

source/api/discovery/0.3/index.md
zimeon: :+1:

source/api/discovery/0.3/index.md
zimeon: :+1:

PR #1799, opened 2019-04-15
Mark w,h syntax required at level1

### LINKED ISSUE #1785: 'w,h' as canonical size param vs. compliance level requirements
Version 3 will have `w,h`as the canonical size URI parameter syntax, but arbitrary `w,h` requests are only required for level 2 compliance. So at level 1, servers are able to serve tiles proportionally scaled to an arbitrary width or arbitrary height, but if clients want to use that feature, they can't use the canonical URI syntax. So if clients want to request tiles via canonical URIs, they can only request them for the listed `tiles` / `scalefactors`. If no `tiles` are given in the info.json, requests for tiles to level 1 services can't be done in canonical form. Same applies to `sizes`.

On the other hand, having `w,h` required for level 1 would require those image servers to also serve non proportionally scaled images.

zimeon:
Fixes #1785


azaroth42:
Note that this PR (given it needs approval by TRC) should also update the compliance JSON... once that hits.

azaroth42:
@jpstroop?

PR #1802, opened 2019-04-15
Update Image API examples in Prezi

azaroth42:
To use `max` and consistent URI structures


azaroth42:
Note that Travis is confused:  https://travis-ci.org/IIIF/api/builds/520437313  is the real PR build.

The failure above is #1801, when I tried to merge to master, and then closed the PR during the build ... but Travis can't seem to distinguish the two PRs.

PR #1804, opened 2019-04-15
Add 4.7 about context conflicts and note in homepage

### LINKED ISSUE #1764: Unclear if Annotations in IIIF inherit JSON-LD requirements
It's not mentioned in [5.6](https://preview.iiif.io/api/image-prezi-rc2/api/presentation/3.0/#56-annotation) if the design patterns for JSON-LD are inherited in our use of the Web Annotation Data Model or not.  For example, as there can be multiple targets, do we require `target` to always be a list? Similarly `motivation`.

We do (implicitly) require it for IIIF defined properties, as the context says that they're `@container:@set`. Hence the single IIIF Image Service on the Image in the body of an Annotation follows different rules compared to other properties potentially on the same JSON object.

If we do - remote Annotations might fail those requirements, as they're not required to by the W3C spec. We would also need to list out all of the requirements somewhere.
If we don't - implementers need to type check everything when they hit Annotations, with different rules per property at times.

Either way, we should make it clear in that section.

### LINKED ISSUE #1765: Annotation homepage vs IIIF homepage
In changing to use `homepage` from `related` we added a new conflict between IIIF and the Annotation model, which also uses `homepage` for agents.
The context definition will mean that the right definition (assuming that homepage on an Annotation's agent is the Annotation version) but we might need to spell it out in the API docs, like we do with `value`.

Conversely if we want agents to always use the same (IIIF) homepage definition, we would need further changes to the specs.  I'm not averse, but something that should be discussed again perhaps now we're closer to the end of the road.

Related to #1764

azaroth42:
Closes #1765 ; Closes #1764


source/api/presentation/3.0/index.md
```diff
@@ -682,6 +682,10 @@ The value of this property _MUST_ be an array of JSON objects, each of which _MU
  * Any resource type _MAY_ have the `homepage` property.<br/>
    Clients _SHOULD_ render `homepage` on a Collection, Manifest or Canvas, and _MAY_ render `homepage` on other resource types.
 
+__Model Alignment__<br/>
+Please note that the [Web Annotation Data Model][org-w3c-webanno] has less stringent requirements about the model to be used for the `homepage` property. The IIIF requirements are compatible, but the home page of an Agent found might have only a URI, or might be a JSON object with other properties.
+{: .note}
```
jpstroop: Small thing, but I wonder if it might be nicer to say, rather than

> the Web Annotation Data Model has less stringent requirements

That IIIF further refines the WADM?

source/api/presentation/3.0/index.md
```diff
@@ -1093,6 +1097,25 @@ Any additional properties beyond those defined in this specification or the Web
 
 The JSON representation _MUST NOT_ include the `@graph` key at the top level. This key might be created when serializing directly from RDF data using the JSON-LD 1.0 compaction algorithm. Instead, JSON-LD framing and/or custom code should be used to ensure the structure of the document is as defined by this specification.
 
+### 4.7. Term Collisions between Contexts
+
+It must be noted that there are some common terms that are used in more than one JSON-LD context document. Every attempt has been made to minimize these collisions, but some are inevitable. In order to know which specification is in effect at any given point, the class of the resource that has the property is the governing factor. Thus properties on Annotation based resources use the context from the [Web Annotation Data Model][org-w3c-webanno], whereas properties on classes defined by this specification use the IIIF Presentation API context's definition.
```
jpstroop: I would strike "It must be noted that "

source/api/presentation/3.0/index.md
```diff
@@ -1093,6 +1097,25 @@ Any additional properties beyond those defined in this specification or the Web
...
+
+The label property is defined by both and is in direct conflict:
+
+* `label`: The use of `label` in IIIF follows modern best practices for internationalization by allowing the language to be associated with the value using the language map construction [described above][prezi30-languages]. The Web Annotation Data Model uses it only for [Annotation Collections][prezi30-annocoll], and mandates that it must be a string.
+
+The following properties are defined by both, and the IIIF representation is more specific than the Web Annotation Data Model but are not in conflict, or are never used on the same resource:
+
+* `homepage`: In IIIF the home page of a resource must be represented as a JSON object, whereas in the Web Annotation Data Model it may also be a string.
```
jpstroop: must or _MUST_ ?

source/api/presentation/3.0/index.md
azaroth42: Also added forward ref to 4.7

source/api/presentation/3.0/index.md

source/api/presentation/3.0/index.md
azaroth42: I wondered about that. I went with lower case as it's not /the/ normative requirement itself, it's just reflecting the requirement from the normative part in section 3. I'm fine to go either way.

zimeon: If we don't want to use normative MUST then say simply "is" to avoid confusion

jpstroop:
@tomcrane can merge

tomcrane:
I missed one char from a merge conflict fix ("@") so will merge anyway as the PR is the same.

PR #1805, opened 2019-04-15
s/@none/none/ in spec

azaroth42:
Does NOT close the issue by itself, but will be closed with the revision to the context docs. This is the normative spec part.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1018,7 +1018,7 @@ Any of the properties in the API that can have multiple values _MUST_ always be 
 
 Language _MAY_ be associated with strings that are intended to be displayed to the user for the `label` and `summary` properties, plus the `label` and `value` properties of the `metadata` and `requiredStatement` objects.
 
-The values of these properties _MUST_ be JSON objects, with the keys being the [BCP 47][org-bcp-47] language code for the language, or if the language is either not known or the string does not have a language, then the key must be the string `@none`. The associated values _MUST_ be arrays of strings, where each item is the content in the given language.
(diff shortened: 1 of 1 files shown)


jpstroop:
Does this need to be explained somewhere, similar to how we explain `type` vs. `@type`, etc, under [service](https://preview.iiif.io/api/image-prezi-rc2/api/presentation/3.0/#service)

azaroth42:
The use of `@type` is because the external service's context might not define those aliases. We don't directly say that `id` is `@id` or `type` is `@type` (or `rdf:type`). So I don't think we need to explain that `none` is `@none`.

jpstroop:
We sort of do:
> The value must be an array of JSON objects. Each object will have properties depending on the service’s definition, but must have either the id or @id and type or @type properties. Each object should have a profile property.
But I'll go with the consensus of the rest of the group

azaroth42:
Also happy to go with consensus, but the intent of `service` talking about them is that per the example, a v2 image api service will have `@id` not `id`. There aren't language maps, or even `@none` in 1.0, so there's no use of `@none` anywhere else in our APIs to talk about. We could have a subsection in 4 that talks about the context mapping for things like this, as it's a JSON-LD issue not a property of a json object issue. But that seems like extra words for little gain.

zimeon:
I think the places where `@type` and `@id` are currently mentioned are not the right places to mention `@none`.

We could add a short "4.7 Keyword Mappings" along the lines of:

> The JSON-LD keywords `@id`, `@type` and `@none` are mapped to `id`, `type` and `none` by the Presentation API linked data context. Thus in Presentation API 3 content the only JSON key beginning with `@` will be `@context`. However, the content may include data conforming to older specifications or external specifications that use keywords beginning with `@`.

Issue #1490, opened 2018-03-01
Name of the `transcribing` motivation


tomcrane:
Presentation 3 introduces an additional annotation motivation, `transcribing`

> Resources associated with a Canvas by an Annotation with the “transcribing” motivation may be presented to the user as part of the representation of the Canvas, or may be presented in a different part of the user interface.

Source: http://prezi3.iiif.io/api/presentation/3.0/#values

This is an obvious, clear and useful motivation name for familiar transcription use cases - OCRed text, human transcribed manuscript, musical notation and other representations. 

The _mechanism_  also works for many AV use cases like multiple camera angles and other forms of graphical representation that stretch the common meaning of the word "transcribing". The motivation means that this content is part of the representation (not commentary, etc), and is directly associated with regions of the canvas, and _could_ be painted but doesn't have to be, a client can be flexible.

Is the word "transcribing" misleading or too domain specific? Especially if #1480 uses the same word form ("transcriptions"). 

Is there a better term?

See https://github.com/IIIF/api/issues/1480 and https://github.com/IIIF/api/issues/1483

zimeon:
Discussed possible use of Icelandic for "side-car": hliðarvagn, but perhaps this won't be the most obvious term for most readers

azaroth42:
Ed's call: We think that the content of the transcribing annotation is an alternate or additional representation of the primary content displayed in the target canvas / part of canvas.  

Propose:  `supplementing`
And thus #1480 would be `supplementary` as the property name?

tomcrane:
Captain's log: `supplemental`

Although that usage is more US. Compare the British and American English usage of supplementary here: https://www.collinsdictionary.com/dictionary/english/supplementary

jpstroop:
Closed by #1499

PR #1807, opened 2019-04-15
Add transcription/translation use case to supplementary/supplementing

### LINKED ISSUE #1738: Inconsistency in patterns for representing translations (wrt 'supplementary')
Imagine a French manuscript with an English translation.
According to the new 'supplementary/supplementing' pattern, the English translations (annotations) would be 'from' the content, and thus can be grouped into as a Range.
It seems not possible to group the French transcriptions into their own Range though, as Ranges look like they're restricted to refer to collections of supplementary annotations (https://preview.iiif.io/api/image-prezi-rc2/api/presentation/3.0/#supplementary and the diagram in 2.1).
If this understanding is correct, it is an odd situation. While there is a difference between the English and the French text, it is desirable to use similar patterns to represent them.

azaroth42:
Closes #1738


source/api/presentation/3.0/index.md
```diff
@@ -973,7 +973,7 @@ Additional motivations may be added to the Annotation to further clarify the int
...
+| `supplementing` | Resources associated with a Canvas by an Annotation that has the `motivation` value `supplementing`  _MAY_ be presented to the user as part of the representation of the Canvas, or _MAY_ be presented in a different part of the user interface. The content can be thought of as being _from_ the Canvas. The use of this motivation with target resources other than Canvases is undefined. For example, an Annotation that has the `motivation` value `supplementing`, a body of an Image and the target of part of the Canvas is an instruction to present that Image to the user either in the Canvas's rendering area or somewhere associated with it, and could be used to present an easier to read representation of a diagram. Similar, a textual body is to be presented either in the targeted region of the Canvas or otherwise associated with it, and might be a transcription or translation of handwritten text or captions for what is being said in a Canvas with audio content. |
 {: .api-table #table-motivations}
```
jpstroop: 👍 but shouldn't it be "Similar**ly**, a textual body..."

azaroth42: Ha, yes, not introduced by this PR, but will fix it while we're here.

jpstroop: Yeah, meant to point that out. May as well...

azaroth42: Fixed in #372c084

tomcrane:
Just getting thoughts in order...

Presentation 3 is providing the `supplementing` motivation for content (typically text) that is _from_ the canvas. This is a refinement on P2, where the spec only provided `painting`, requiring a client to render a textual body directly on the canvas (although none did, suggesting that the addition of `supplementing` is correct).

We have not gone beyond `supplementing`; sticking with Presentation semantics, we haven't opted to add the motivations `transcribing` and `translating`, despite the obvious use cases.

We think we should see further differentiation of `supplementing` annotations through extensions, and we also think that recipes in the cookbook are the place to demonstrate this (like https://github.com/IIIF/cookbook-recipes/issues/92). 

`"motivation": [ "supplementing", "transcribing" ]`

As extensions, there is nothing stopping `translating` and `transcribing` becoming first-class spec motivations in 3.1, if they are popular and clear patterns emerge. But this is the last chance for them to be part of 3.0.

Against:

* Use in the wild can see them come in in 3.1: one step at a time
* `transcribing` vs `translating` - not Presentation semantics...
* If `translating` provided, how do you use it? Does the translating anno target the Canvas, or does it target the transcribing anno? That _is_ what recipes are for, but by providing the motivation the spec has an additional responsibility.

For:

* There _is_ a Presentation semantics difference between transcribing and translating (although both are supplementing). In a transcription, the content _from_ the image retains its **form** but not its format (e.g., it's the same letters). In a translation, it takes a different form. A client could paint or not-paint a transcribing anno, but refuse to paint a translating anno, always rendering it differently. A client could present the transcription, and know its language, and also present several available translations, knowing their languages. With `supplementing` only, a UI can only use labels and can't separate them off (a similar argument is given for captions in #1258).
* Such a common use case should be encouraged.

For reference:

https://github.com/IIIF/api/issues/1258
https://github.com/IIIF/api/issues/1490

Here's where I stand (today at least) - `transcribing` and `translating` should not be in 3.0, even though I would really want to use them. They will make an excellent demonstration of the extension process if they are introduced that way.

PR #1809, opened 2019-04-15
Image info response must include all qualities beyond default

### LINKED ISSUE #1754: Clarify that level2 image API MUST for extra qualities should be in the info.json
Rather than omitting as they MUST be there at level 2

jpstroop:
Closes #1754

### DIFF
--- a/source/api/image/3.0/compliance.md
+++ b/source/api/image/3.0/compliance.md
@@ -84,7 +84,7 @@ Note that servers may express limits on the sizes available for an image with th
 | `default`     | ![required][icon-req] | ![required][icon-req] | ![required][icon-req] |
 | `color`       | ![optional][icon-opt] | ![optional][icon-opt] | ![required][icon-req] (if applicable) |
 | `gray`        | ![optional][icon-opt] | ![optional][icon-opt] | ![required][icon-req] (if applicable) |
-| `bitonal`     | ![optional][icon-opt] | ![optional][icon-opt] | ![required][icon-req] |
+| `bitonal`     | ![optional][icon-opt] | ![optional][icon-opt] | ![optional][icon-opt] |
 {: .api-table}
 
 ### 3.5 Format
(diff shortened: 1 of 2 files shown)


azaroth42:
Meaning that info.json would always need to list `bitonal` ?

jpstroop:
I went back and forth. It's such an edge-case quality, and the fact that it's only required at level 2; I think this just simplifies things. I can add "or bitonal at compliance level 2" if you want, but my instinct was otherwise.

jpstroop:
Also, think about the developer experience. Having all of the qualities in one place (beyond `default`) is a better experience than having to check the qualities list AND the compliance level, IMO.

azaroth42:
I agree about the ergonomics, it's just a change. The inconsistency is if we leave bitonal in the compliance doc at level 2, then it's repeated ... which we don't do in other places. We could make it optional at level 2, forcing the publisher to put it in info.json

jpstroop:
> We could make it optional at level 2, forcing the publisher to put it in info.json
I like that. @zimeon @tomcrane @mikeapp ?

mikeapp:
I would be OK with making it optional, but I assume then we're needing TRC review?

tomcrane:
I think bitonal is fine to be Optional at level 2.

As an art museum with lovely paintings, I have a level 2 image server but I really don't want bitonal available. It's just silly.

jpstroop:
Yes, I don't think I've ever heard a real world use case for bitonal

tomcrane:
One _possible_ use case is harvesting a huge amount of text-bearing images, for OCR or machine learning purposes. Although I suspect most people would want to manage the conversion of the images into training set material themselves.

jpstroop:
>One possible use case is harvesting a huge amount of text-bearing images, for OCR or machine learning purposes. Although I suspect most people would want to manage the conversion of the images into training set material themselves.

Yes, and we've actually never found bitonal to work as well as grayscale for OCR.

jpstroop:
So we're in general agreement about making bitonal optional at all levels, can we merge before TRC approval, or does this need to sit open?

zimeon:
I think it need to sit open if normative change, we shouldn't get into the habit of merging before approval.

While we are on this subject, should `color` and `gray` me marked simply optional? "Required if appropriate" is the same as optional from a client point of view.

Issue #1784, opened 2019-02-28
extraQualities, extraFormats and extraFeatures should not expand to the same IRI


lutzhelm:
The `extraQualities`, `extraFormats` and `extraFeatures` properties of the version 3 image information response all expand to the same IRI `http://iiif.io/api/image/3#quality`. This will make it difficult to use (and especially distinguish between) them in hypothetical JSON-LD purposes. As I understand it, it will be impossible to keep them apart during framing (except by the `@vocab` definition in `extraFeatures`) or after they have been expanded.
Is there any reason / advantage behind using the same IRI or maybe a way around the problem I described?

zimeon:
Relevant section in https://iiif.io/api/presentation/3/combined-context.json is:

```

"extraQualities": {
  "@id": "image:quality",
  "@container": "@set"
},
"extraFormats": {
  "@id": "image:quality",
  "@container": "@set"
},
"extraFeatures": {
  "@id": "image:quality",
  "@type": "@vocab",
  "@container": "@set"
},
```

which has been only slightly modified in [current draft for RC2](https://github.com/IIIF/api/blob/master/source/api/presentation/3/combined-context.json#L238-L250).

If I recall, the discussion about merging everything into `quality` was that there was no possibility of overlap and it avoided more term pollution. However, we do want to be able to use a JSON-LD frame to create a valid Image API `info.json` so I think it would make sense to split them, perhaps into `qualities`, `formats` and `features`.

azaroth42:
Agree that we should split into three. I think should be singular, as the property doesn't take an rdf:List as the object.

Thus:  `extraFeatures` would expand to `image:feature`

zimeon:
Editors agree: `image:quality`, `image:format`, `image:feature`

zimeon:
Closed by merge of #1810

Issue #1780, opened 2019-02-20
Add provider (and any downstream) to context


azaroth42:
The term for provider is `schema:provider`, but this isn't in the context document.  Thanks to @regisrob for pointing this out.

zimeon:
Closed by merge of #1810

Issue #1779, opened 2019-02-18
Terms defined in API drafts missing in combined-context.json


lutzhelm:
While playing around with version 3 I noticed that the following terms are defined in the Presentation API or Image API drafts but have not yet been defined in the combined-context.json. I don't know at what point any term definition is added to the context document...

Image API types (already fixed via #1774):

- ~~`Tile`~~
- ~~`Size`~~

Image API values:
- `sizeUpscaling`

Presentation API properties:

- `placeholderCanvas`
- `accompanyingCanvas`
- `provider`

Presentation API values:

- `no-auto-advance`
- `no-repeat`

zimeon:
Closed by merge of #1810

Issue #1739, opened 2018-12-03
Should we alias @none to none (or similar)?


jpstroop:
[from Rob]
The advantage of the language map pattern is you can do `manifest.label.en` to get the english labels for the manifest.  We can't do that for no language without using the disliked `['@none']` pattern.  Thus we should alias `@none` to something and discuss with JSON-LD WG if there's a best practice here

azaroth42:
Yes, see json-ld playground: http://tinyurl.com/yd2v4mlu

azaroth42:
Question - can we use scoped contexts to add the alias for @none ONLY within the properties that take a language?

tomcrane:
:+1: to "none"

azaroth42:
Raised upstream clarification issue about the definition of @none in JSON-LD:  https://github.com/w3c/json-ld-syntax/issues/102

mikeapp:
👍 to `none`

zimeon:
Given (hopeful) clarification of JSON-LD description of `@none` to mean more than just "no language", I'm :+1: on `none`

christopher-johnson:
:+1: to none.  This also facilitates GraphQL  (it cannot query fields with `@`).

zimeon:
`none` issues addressed by merge of #1810, the question os `en-us` split off into #1771. Closing

PR #1810, opened 2019-04-15
update contexts for 3.0

azaroth42:
I believe this closes ... #1784, #1780, #1779, #1739 (the context part), and #1571.


tomcrane:
I agree with @zimeon that we should merge to get more eyes on it.

PR #1813, opened 2019-04-15
Use consistent form for normative property descriptions

### LINKED ISSUE #1730: Consistency of normative requirements for properties
Sometimes we collapse the same requirement for the property, eg:

> A Collection, Manifest, Range or Canvas may have the navDate property.

And sometimes we spell it out for each class individually, eg:
> A Collection may have the placeholderCanvas property.
> A Manifest may have the placeholderCanvas property.
> ...

We should be consistent.  The former makes for a slightly shorter specification, the latter is easier for making a checklist and verifying you meet requirements.

zimeon:
Fixes #1730

Also has a couple of line end and other space fixes

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -269,7 +269,7 @@ The value _MUST_ be a string. If the value is drawn from Creative Commons or Rig
 
 An organization or person that contributed to providing the content of the resource. Clients can then display this information to the user to acknowledge the provider's contributions.  This differs from the `requiredStatement` property, in that the data is structured, allowing the client to do more than just present text but instead have richer information about the people and organizations to use in different interfaces.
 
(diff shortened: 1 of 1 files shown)


zimeon:
Question of whether to also split the two cases where the allows clause is common but the client handling is split:

https://github.com/IIIF/api/commit/889716ec35689d7315e18ecff54745a8f8986189#diff-35b93d1a4cd76f02960d65b1ed15692bR692

https://github.com/IIIF/api/commit/889716ec35689d7315e18ecff54745a8f8986189#diff-35b93d1a4cd76f02960d65b1ed15692bR760

Also, how fussy are we about language? I think instead of:

< Other resource types....

we should really say:

> Other types of resource...

tomcrane:
I agree with "Other types of resource"

PR #1814, opened 2019-04-16
^pct:n clarification and canonical values for upscaling

### LINKED ISSUE #1741: Should explicit upscale (^) be required for pct:nnn?
In the current draft, any upscaling operation must be indicated by a "^" (caret) in the size parameter.

For percentage size operations, this is technically redundant: `pct:200` and `^pct:200` can only mean the same thing.

Which is better? Be consistent, so that upscaling is _always_ indicated by a caret?
Or not require it on `pct`, as it is redundant?

mikeapp:
Provides description of upscaling error codes (not implemented vs bad request) and uses example of `^pct:120` as a case where the server might return 501.   

Also note canonical form is `pct:n` for `n` <=100

Closes #1741 and #1824

### DIFF
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -233,7 +233,7 @@ The size parameter specifies the dimensions to which the extracted region, which
 
 Requests for sizes not prefixed with `^` that result in a scaled region with pixel dimensions greater than the pixel dimensions of the extracted region are errors that _SHOULD_ result in a 400 (Bad Request) status code.
 
-Requests for sizes prefixed with `^` that result in upscaling _SHOULD_ result in a 501 (Not Implemented) status code if the server does not support upscaling, while a 400 (Bad Request) status code _SHOULD_ be returned in response to other client request syntax errors.  For example, a request for the size `^pct:120` should result in a 501 status code if the server does not support upscaling.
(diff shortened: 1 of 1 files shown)


azaroth42:
Two comments...

* This should go to TRC, as it changes the normative rules around canonical IIIF Image URIs.
* Shouldn't the note about the canonical form of `^` just be for *all* uses of it? Any time you are upscaling you need to use it, and if you don't need it, then it's not canonical.

mikeapp:
OK, I tried to spell out the alternatives in the size section of the Canonical Values table.  
Preview: https://preview.iiif.io/api/1741_image_pct_n/api/image/3.0/#47-canonical-uri-syntax

tomcrane:
Does _4.2 Size_  need a little extra?

> `max` | The extracted region is returned at the maximum size available. The resulting image will have the pixel dimensions of the extracted region, unless it is constrained to a smaller size by maxWidth, maxHeight, or maxArea as defined in the Technical Properties section.

If the service supports `sizeUpscaling`, I can request both /max/ and /^max/ and get different sized images back. "maximum size available" is ambiguous - it is clarified by the next sentence, but perhaps this could be reinforced as "The extracted region is returned at the maximum size available, but will not be upscaled".

zimeon:
I think the canonical section should not talk of "results in" but of what is requested. From a client point of view -- which is where the decision of what request to make is made -- this must be calculated beforehand. Perhaps:

> `max` if the maximum size without upscaling is requested,
> `^max` if upscaling is supported and the maximum upscaled size is requested, 
> `w,h` if the size requested does not require upscaling, or
> `^w,h` if upscaling is supported and the request requires upscaling of the extracted region.

mikeapp:
Also in 4.2 the size of ^max ends up unspecified if none of the optional max.. properties is present.  I'll make an issue.

PR #1800, opened 2019-04-15
[TRC-DNM] 1787 remove logo reqdstmnt from image

tomcrane:
fixes #1787

PR #1818, opened 2019-04-16
1787 remove logo reqdstmnt from image

### LINKED ISSUE #1787: Remove logo and attribution/requiredStatement from Image 3
Given the (now TRC approved) change to Presentation API to add a provider structure around the logo and other information about the organizations and/or people that should be attributed as providers of the information, there was a discussion about whether that should be done or not.  The result of the discussion on the community calls was that everyone believed that it would be too much of a stretch for the Image API to deal with this, and it would remain unimplemented.

A call for input was sent to iiif-discuss to solicit opinions if there was disagreement, and no-one responded.

Thus this issue is to remove logo and requiredStatement from the info.json of the Image API v 3.0 and instead recommend that the content be represented in the Presentation API, using the `provider` construct.

jpstroop:
Not a DNM. Re-merges accidentally umerged work to close #1787


jpstroop:
Already approved in #1800

PR #1828, opened 2019-04-17
Explain @ keywords

### LINKED ISSUE #1819: Describe mapping of JSON-LD keywords to avoid @ signs?
From #1805 discussion. 

I think the places where `@type` and `@id` are currently mentioned are not the right places to mention `@none`.

We could add a short "4.7 Keyword Mappings" along the lines of:

> The JSON-LD keywords `@id`, `@type` and `@none` are mapped to `id`, `type` and `none` by the Presentation API linked data context. Thus in Presentation API 3 content the only JSON key beginning with `@` will be `@context`. However, the content may include data conforming to older specifications or external specifications that use keywords beginning with `@`.

zimeon:
Fixes #1819

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1136,6 +1136,10 @@ The following properties are defined by both, and the IIIF representation is mor
 
 The `rights`, `partOf`, and `items` properties are defined by both in the same way.
 
+### 4.8. Keyword Mappings
+
+The JSON-LD keywords `@id`, `@type` and `@none` are mapped to `id`, `type` and `none` by the Presentation API [linked data context][prezi30-ldce]. Thus in content conforming to this version of the Presentation API, the only JSON key beginning with `@` will be `@context`. However, the content may include data conforming to older specifications or external specifications that use keywords beginning with `@`.
+
 
 ##  5. Resource Structure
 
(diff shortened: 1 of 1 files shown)


azaroth42:
@jpstroop you can merge

source/api/presentation/3.0/index.md
```diff
@@ -1136,6 +1136,10 @@ The following properties are defined by both, and the IIIF representation is mor
 
 The `rights`, `partOf`, and `items` properties are defined by both in the same way.
 
+### 4.8. Keyword Mappings
+
+The JSON-LD keywords `@id`, `@type` and `@none` are mapped to `id`, `type` and `none` by the Presentation API [linked data context][prezi30-ldce]. Thus in content conforming to this version of the Presentation API, the only JSON key beginning with `@` will be `@context`. However, the content may include data conforming to older specifications or external specifications that use keywords beginning with `@`.
+
```
jpstroop: ...and therefore clients should be prepared to handle both syntaxes?

azaroth42: We do say that already in the Extensions section a couple sections above ... but repeating it certainly wouldn't hurt. :+1:

zimeon: "should be prepared to handle" seems too strong, it would seem reasonable to write a v3 and up client that ignored data from earlier specs or other specs that use `@` keywords. Maybe simply "Clients should expect to encounter both syntaxes."?

(We already say "If a client discovers properties that it does not understand, then it MUST ignore them.")

PR #1842, opened 2019-05-15
Replace color quality with full

### LINKED ISSUE #1839: Quality in Image API v3: Replace `color` with `full`
Following the discussion that developed on https://github.com/IIIF/trc/issues/19, the editors propose that potential ambiguity about whether a server should offer `color`, and what it should do with requests for `color`, could be resolved by using the term `full` to avoid the implication that a `color` quality image should contain non-gray-shade pixels, or use particular color encodings.

Therefore (wording provisional):

* `bitonal`: The image returned is bitonal, where each pixel is either black or white.
* `gray`: The image is returned in grayscale, where each pixel is black, white or any shade of gray in between.
* `full`: The image is returned with as much color information as available.
* `default`: The image is returned using the server’s default quality (e.g. full, gray or bitonal) for the image.

Observation:
A `full` image may contain only gray shades, if that is all that is available: it is legitimate for the server to list `full` in this case even if the response would be identical to that for `gray`.

tomcrane:
fixes #1839


tomcrane:
Some wordsmithing required I think, but the basics are here.

azaroth42:
iiif/trc#25 is pending Ex Officio decision, so DNM.

zimeon:
https://github.com/IIIF/trc/issues/25#issuecomment-500079866 rejected pending further research and discussion, closing this PR

PR #1843, opened 2019-05-15
Be consistent that `type` is required with `thumbnail` and `logo`

### LINKED ISSUE #1835: Description of thumbnail & logo implies type is not mandatory
Both thumbnail and logo say: ... where each item ... SHOULD have at least one of the `type` and `format` properties.
It MUST have the `type` property!

I think this is a holdover from previous spec text. Debatable if this is normative (rfc SHOULD) or editorial ... tagging as normative for now.

zimeon:
Closes #1835


zimeon:
The change that was made to text at line 715 in 67b4cd8 was removed in 1845971 because it had been incorporated a paragraph later by #1850

PR #1847, opened 2019-06-02
example and use case for preferredFormats

### LINKED ISSUE #1838: preferredFormats should have use case, example
The description of the preferredFormats property in the Image API 3.0 beta draft does not have an example in its section (5.5) and should for consistency and clarity.

Also noted on the TRC review of the issue is that the description lacks sufficient informative discussion to know when publishers might want to use the function, and what clients might be expected to do with it when encountered. Some further use case description would help understanding as to its purpose and intended effects.

tomcrane:
fixes #1838


azaroth42:
@jpstroop ?

PR #1848, opened 2019-06-06
Add format property to thumbnail and logo examples

### LINKED ISSUE #1844: We say `thumbnail` and `logo` SHOULD have `format`, but none of the examples show that
I think we should add `format` properties to some of the examples. I think the general principle is that examples should generally include SHOULD properties.

zimeon:
Closes #1844

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1037,7 +1037,11 @@ Any of the properties in the API that can have multiple values _MUST_ always be 
 ``` json-doc
 {
   "thumbnail": [
-    { "id": "https://example.org/images/thumb1.jpg", "type": "Image", "format": "image/jpeg" }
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -1034,7 +1037,7 @@ Any of the properties in the API that can have multiple values _MUST_ always be
 ``` json-doc
 {
   "thumbnail": [
-    { "id": "https://example.org/images/thumb1.jpg", "type": "Image" }
+    { "id": "https://example.org/images/thumb1.jpg", "type": "Image", "format": "image/jpeg" }
```
azaroth42: Should this instead be 3 lines?

zimeon: Done

PR #1852, opened 2019-06-19
Release Image, Prezi 3.0 Beta

azaroth42:
Push Image and Presentation 3.0 beta specs to master before the conference


zimeon:
The red from travis is because in `master` we have `discovery/0.2` and `discovery/0.3` that aren't in the `image-prezi-rc2` branch. The file `source/api/discovery/0.3/index.md` includes a reference to link `[annex-registry]` that we have renamed `[registry]` in the `image-prezi-rc2` branch.

I think the "correct" way to fix this would be to rebase `image-prezi-rc2` onto the current `master` and add a fix for `source/api/discovery/0.3/index.md`. However, that involves lot of manual conflict resolution that I don't have time to work through (and would be error prone). 

I think a simpler way round is to make a new branch `image-prezi-rc2-b` that is `image-prezi-rc2` merged into `master`, add in the fix for `source/api/discovery/0.3/index.md` and then review that as a PR onto `master` to publish the beta 3.0s ---> #1853

PR #1855, opened 2019-06-26
Add sizeByWh to level1.json

lutzhelm:
#1799 as solution for  #1785 did not include the necessary change to the compliance level document.


azaroth42:
Thanks! Good catch @lutzhelm !

PR #1865, opened 2019-08-08
Adding support for 3.0 

glenrobson:
Currently only supports level 0 but working on supporting the other levels.


glenrobson:
Available at: https://preview.iiif.io/api/image-3.0-validator/api/image/validator/

glenrobson:
Example 3.0 tiled image: 

https://glenrobson.github.io/iiif/30-validation-image/info.json

Generated from https://github.com/glenrobson/iiif-tiler

glenrobson:
I've deployed an updated validator which has full support for level 3.0. The only thing left to do is to deploy this change so we can share with people to test.

Issue #758, opened 2016-03-24
Annotation lists with different target specificities (word, line, paragraph)


glenrobson:
How do I share multiple version of the same annotation list with different specificities so for example I might have:
- a word level annotation list for harvesting by Europeana
- a line level annotation list for use in Mirador
- a paragraph annotation list for OCR correction.

I would like to be able to link to these options to allow the client to decide which ones they want to use.

jpstroop:
The only known spec for hOCR:

https://docs.google.com/document/d/1QQnIQtvdAC_8n92-LhwPcjtAUFwBlzE8EWnKAxlgVf0/preview

Typesetting elements seem relevant, though they are not consistent with what you get from Tesseract.

Sample:

```html
    <div class='ocr_page' id='page_1' title='image "crowley/ah8_Vol01_0005_original.jpg"; bbox 0 0 5018 3668; ppageno 0'>
      <div class='ocr_carea' id='block_1_1' title="bbox 449 291 2388 3013">
        <p class='ocr_par' dir='ltr' id='par_1_1' title="bbox 474 291 2388 656">
          <span class='ocr_line' id='line_1_1' title="bbox 474 291 2388 372; baseline -0.009 -13">
            <span class='ocrx_word' id='word_1_1' lang='ita' title='bbox 474 356 476 359; x_wconf 83'>
              <strong>.</strong>
            </span>
            <span class='ocrx_word' dir='ltr' id='word_1_2' lang='ita' title='bbox 498 308 637 365; x_wconf 88'>
              <strong>detto,</strong>
            </span>
            <span class='ocrx_word' dir='ltr' id='word_1_3' lang='ita' title='bbox 649 307 675 357; x_wconf 90'>
              <strong>il</strong>
            </span>
            <span class='ocrx_word' dir='ltr' id='word_1_4' lang='ita' title='bbox 690 305 948 368; x_wconf 79'>
              <strong>_Sanl&#39;ouino,</strong>
            </span>
```

zimeon:
[...]

PR #1870, opened 2019-08-22
Text granularity extension property

mikeapp:
https://preview.iiif.io/api/text-granularity/api/extension/text-granularity/

See issue #758

### DIFF
--- a/source/api/extension/text-granularity/index.md
+++ b/source/api/extension/text-granularity/index.md
@@ -48,7 +48,7 @@ The `textGranularity` property identifies the Text Granularity Level of a resour
 | `line`                       |  A topographic line             |
 | `word`                       |  A single word                  |
 | `glyph`                      |  A single glyph or symbol       |
-{: .api-table #table-type-dfn}
+{: .api-table #table-granularity-level-dfn}
 
 ```json-doc
 { "textGranularity": "line" }
@@ -131,7 +131,7 @@ The Text Granularity Levels defined above were derived from a survey of commonly
 | `line`      | `ocr_line`<br/>`ocrx_line` | `TextLine` | `line` | - |
 | `word`      | `ocrx_word`    | `String`      | `charParams[@wordStart=”1”]` | `Word` |
 | `glyph`     | `ocrx_word[@cuts]` | `Glyph`   | `charParams`        | `Symbol` |
-{: .api-table #table-type-dfn}
+{: .api-table #table-ocr-granularity-map}
 
 ## Appendices
 
--- a/source/api/extension/text-granularity/index.md
+++ b/source/api/extension/text-granularity/index.md
@@ -23,7 +23,7 @@ Changes will be tracked within the document.
 
 ### 1.1 Objectives and Scope
 
-One use of annotation in IIIF is to associate text with IIIF resources, with the text originating from optical character recognition (OCR) software, manual transcription, existing digitized text, or other sources. Furthermore, it is a common practice to produce sets of annotations at a particular level of ‘text granularity’; that is to say that each annotation in the set references the same unit of text, such as a character, line, or page. This extension recommends a pattern for indicating the level of text granularity of an annotation.
+One use of annotations in IIIF is to associate text with IIIF resources, with the text originating from many potential sources, including optical character recognition (OCR) software, manual transcription, and existing digitized text. Furthermore, it is a common practice to produce sets of annotations at a particular level of ‘text granularity’, where each annotation in the set references the same unit of text, such as a character, line, or page. This extension recommends a pattern for indicating the level of text granularity for an annotation.
 
 ### 1.2 Motivating Use Cases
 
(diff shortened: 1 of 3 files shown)


source/api/extension/text-granularity/index.md
```diff
@@ -0,0 +1,158 @@
...
+{:.no_toc}
+
+* Table of Discontent (will be replaced by macro)
+{:toc}
+
+## 1. Introduction
+
+### 1.1 Objectives and Scope
+
+One use of annotation in IIIF is to associate text with IIIF resources, with the text originating from optical character recognition (OCR) software, manual transcription, existing digitized text, or other sources. Furthermore, it is a common practice to produce sets of annotations at a particular level of ‘text granularity’; that is to say that each annotation in the set references the same unit of text, such as a character, line, or page. This extension recommends a pattern for indicating the level of text granularity of an annotation.
```
jpstroop: One use of annotations in IIIF is to associate text with IIIF resources, with the text originating from many potential sources, including optical character recognition (OCR) software, manual transcription, and existing digitized text. Furthermore, it is a common practice to produce sets of annotations at a particular level of ‘text granularity’, where each annotation in the set references the same unit of text, such as a character, line, or page. This extension recommends a pattern for indicating the level of text granularity for an annotation.

source/api/extension/text-granularity/index.md
```diff
@@ -0,0 +1,158 @@
...
+### 1.2 Motivating Use Cases
+
+A number of common workflows can result in annotation sets with discrete levels of text granularity:
+
+- OCR text is frequently available at a range of granularities, from page to character.
+- Manual transcription user interfaces may constrain user input to the level of the line,
+block, or page.
+- Transcriptions may be produced without coordinate data and thus have a very coarse level of granularity, such as page- or block-level.
+
+Identification of the level of text granularity in published annotations can facilitate the reuse of their textual content and target regions in other applications. Search is a primary use case, with the word-level annotations allowing for the accurate highlighting of search result terms in the user interface. Software designed for user interaction and input, such as an application designed to allow users to correct OCR text, might suppress finer-grained levels and make use of paragraph- or block-level text to simplify editing.
```
jpstroop: >Search is a primary use case, with the word-level annotations allowing for the accurate highlighting of search result terms in the user interface.

This bothers me a little because it leads with a suggestion that could be read as the spec being concerned with server-side tech.  Could we reverse and say something are:

> Interfaces that require highlighting of search results in context are a primary use case.

?

jpstroop: >Software designed for user interaction and input, such as an application designed to allow users to correct OCR text, might suppress finer-grained levels and make use of paragraph- or block-level text to simplify editing.

Seems like a little bit of a non-sequitur? Previous sentence is a use case; this seems more like an implementation detail?

source/api/extension/text-granularity/index.md
```diff
@@ -0,0 +1,158 @@
...
+| `glyph`                      |  A single glyph or symbol       |
+{: .api-table #table-granularity-level-dfn}
+
+```json-doc
+{ "textGranularity": "line" }
+```
+
+## 3. Use of the `textGranularity` Property with Annotations
+
+An Annotation _MAY_ have the `textGranularity` property. An Annotation that has the property _SHOULD_ target an IIIF Presentation API Canvas or segment and the identified Text Granularity Level _SHOULD_ describe that of the textual content represented by the content resources painted on the Target.
```
jpstroop: "target" -- is that a word we use? "reference"? "be associated with"?

source/api/extension/text-granularity/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+The Annotation Body’s textual content _SHOULD_ be equivalent to the textual content represented by the content resources painted on the Target. For example, the Body of the Annotation might be a [TextualBody](https://www.w3.org/TR/annotation-model/#embedded-textual-body) that contains the transcription of the Target, which is painted with the image of a page of a medieval manuscript.
```
jpstroop: > ..._SHOULD_ be equivalent...

What about _logically_ equivalent? Or some other way of making it clearer that, e.g., the text and the image should represent the same resource. I'm not sure that the "For example..." really captures it.

jpstroop: There are a few inline links in the doc that should go in `links.md`

source/api/extension/text-granularity/index.md
```diff
@@ -0,0 +1,158 @@
...
+  "@context": [
+     "http://iiif.io/api/extension/text-granularity/context.json",
+     "http://iiif.io/api/presentation/3/context.json"
+  ]
+}
+```
+
+## 5. Implementation Note
+
+The Text Granularity Levels defined above were derived from a survey of commonly used formats for OCR output such as hOCR, ALTO, and ABBYY, as well the output produced by the Google Cloud Vision API. The findings are available in the [Granularities in Different OCR Formats document](https://docs.google.com/document/d/13R7Dk-AA-ALZ5i3fzAS3g66umWjiYaD7NVr8QiYb19E/edit#heading=h.q6phhnqlv5sf). The following table maps Text Granularity Levels to structures found in these output formats and is intended as an aid to implementers.
```
jpstroop: Should the Google doc be translated to Markdown included on the website?

mikeapp: I'm going to remove this, after reflection I don't think the document adds much.

source/api/extension/text-granularity/index.md
mikeapp: Changing to `..._SHOULD_ reference a IIIF Presentation API Canvas or segment in the `target` property and...`

source/api/extension/text-granularity/index.md
mikeapp: Reworked to emphasize these are use cases. 

>  A primary use case is search: clients could use word-level transcription annotations to provide a search function capable of accurate hit highlighting.  Other common use cases derive from crowdsourced text input, such as the correction of OCR transcription, which may require the user to work with a text at a single level of granularity, such as line- or block-level.

source/api/extension/text-granularity/index.md
mikeapp: Regarding "equivalent", I've created an issue since this is more than just a gloss and should probably be reviewed by the TRC.  https://github.com/IIIF/api/issues/1889   I'd suggest we keep the current text, since the document isn't semantically versioned, and adjust once we have agreed on a suggested change.

I think the links are working, some break because they reference the community part of the site and so aren't present in the branch preview.

PR #1877, opened 2019-10-17
Presentation validator 3.0

glenrobson:
Changes to the validator html code to add 3.0 as an option. Includes changes to the javascript as the return from the 3.0 jsonschema is different to the 2.x responses.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ scripts/factory.pyc
 .ruby-version
 package-lock.json
 .DS_Store
+*.swp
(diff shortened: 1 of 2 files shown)


glenrobson:
This needs to be merged first: https://github.com/IIIF/presentation-validator/pull/73

zimeon:
https://preview.iiif.io/api/presentation_validator_3.0/api/presentation/validator/service/index.html

glenrobson:
This is good to go if someone can merge it. I can't test until its live but the html page will behave similar to this:

http://iiif-presentation-validator.us-east-1.elasticbeanstalk.com/

PR #1881, opened 2019-10-30
Allow shared services pattern

### LINKED ISSUE #1873: Use JSON-LD 1.1 @included for services
*Background:*

In version 2.x, feedback about attaching services to the object(s) that the service relates to was that this was painful to use, as only the first occurrence of a service was included in full -- other occurrences simply have the URI, and the client is expected to search through the entire document to find the previous reference.

This feedback was taken to the JSON-LD 1.1 Working Group in the W3C, and made it into the specification (though only just, it was the last issue to be resolved). The solution adopted is to allow a key at the root of the document tree called `@included`. This is described in the [latest draft spec](https://www.w3.org/TR/json-ld11/#included-nodes), but is an array of nodes.

*Issue:*

It's hard to find shared nodes in the document, such as Authentication services. They should be somewhere convenient for implementers.

*Proposal:*

Replace the current `service` pattern with the `@included` pattern, such that all services are described in full in the `@included` property (which can be aliased to get rid of the `@`), and all references to them are only by URI.

*Issue:*

There is a significant issue with this approach though.  The `service` property is used for all services, and the most common one is likely to be the IIIF Image API being referenced from the Presentation API. It makes no sense to take all of the Image API services out, along with the Auth services. I don't think we can use scoped contexts, as the same image might have both auth and image API services (or any other service). We could have a `shared` property that just references included things ... but then you need to look in two places on the resource ... `service` and `shared`.

Thoughts?

azaroth42:
Closes #1873

Note that this is a substantive, normative change and requires TRC approval.
Approval request: https://github.com/IIIF/trc/issues/31

Do not merge before TRC approves.

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1,6 +1,6 @@
 ---
-title: "Presentation API 3.0 BETA DRAFT"
-title_override: "IIIF Presentation API 3.0 BETA DRAFT"
+title: "Presentation API 3.0 BETA-2 DRAFT"
+title_override: "IIIF Presentation API 3.0 BETA-2 DRAFT"
 id: presentation-api
 layout: spec
 cssversion: 3
@@ -8,7 +8,7 @@ tags: [specifications, presentation-api]
 major: 3
 minor: 0
 patch: 0
-pre: BETA
+pre: BETA-2
 redirect_from:
   - /api/presentation/3/index.html
 ---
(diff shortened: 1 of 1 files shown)


jpstroop:
Ummm:
>A list of one or more service definitions on the top-most resource of the document, that are typically shared by more than one subsequent resource. This allows for these shared services to be collected together in a single place, rather than either having their information duplicated potentially many times throughout the document, or requiring a consuming client to traverse the entire document structure to find the information.

https://www.w3.org/2018/jsonld-cg-reports/json-ld/#id-maps :
>An id map is used to associate an IRI with a value that allows easy programmatic access. 

[What goes around](https://github.com/IIIF/api/issues/1072)...

azaroth42:
Haha! :) Yes, the included block and id-maps serve a similar but not identical function.

id-maps are a JSON object, where the key MUST be the URI of the rest of the resource that's the value.  They're related directly to the containing resource via a predicate, and thus appear embedded in that resource.

The included list is an array of JSON objects that must be resources (e.g. the object in the array has an `@id` key). The property in the JSON does **not** map to any predicate, it's just a bucket to throw resources into. We added it to avoid having to use a named graph container (an id-map where the ids are for graphs rather than resources).

The advantages of the included pattern are, thus, that it falls into the main document at the top level without needing a predicate to relate the top resource to the array. So there's no has_referenced_service_somewhere_in_this_manifest fake property needed. The expansion algorithm simply includes the resources in the set of known resources.  So the service resource is referenced in the resource that has it via `service` still, but the description of it is in a known place at the top of the tree.

source/api/presentation/3.0/index.md
```diff
@@ -811,6 +811,41 @@ Implementations _SHOULD_ be prepared to recognize the `@id` and `@type` property
...
+
+A client encountering a `service` property where the definition consists only of an `id` and `type` _SHOULD_ then check the `services` property on the top-most resource for an expanded definition.  If the service is not present in the `services` list, and the client requires more information in order to use the service, then it _SHOULD_ dereference the `id` (or `@id`) of the service in order to retrieve a service description. 
+
+The value _MUST_ be an array of JSON objects. Each object _MUST_ a service resource, as described above.
```
regisrob: Isn't there a verb missing in the second sentence?

zimeon: Agreed --> _"... MUST **be** a service resource."_

zimeon:
Approved by [TRC](https://github.com/IIIF/trc/issues/31), merging

PR #1885, opened 2019-10-31
fix 6.3 para 1

### LINKED ISSUE #1866: Mention of protocol requirements being summarized in Appendix A is obsolete
Section "6.3. Responses" of P-API 3.0 still refers to requirements regarding the dereferencibility of resources, even though the "Protocol Behavior" table has been removed (see #1618):

> The different requirements for which resources must provide a response is summarized in Appendix A.

I may have not seen it, but it seems there is no mention of such requirements in Appendix A at all anymore...

So the first paragraph of 6.3. should be reviewed a little bit to reflect that... Maybe we should just keep the last sentence "While some resources do not require their URI to provide the description, it is good practice if possible", if this is desirable.

azaroth42:
Remove reference to removed table of protocol requirements

Closes #1866

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1679,7 +1679,7 @@ Clients _MUST NOT_ attempt to construct resource URIs by themselves, instead the
 
 ###  6.3. Responses
 
-The format for all responses is JSON, as described above. The different requirements for which resources _MUST_ provide a response is summarized in [Appendix A][prezi30-appendixa]. While some resources do not require their URI to provide the description, it is good practice if possible.
+The format for all responses is JSON, as described above. It is good practice for all resources with an HTTP(S) URI to provide their description when the URI is dereferenced. If the resource is referenced, rather than being embedded in its entirety, then it _MUST_ be able to be dereferenced.
 
 If the server receives a request with an `Accept` header, it _SHOULD_ respond following the rules of [content negotiation][org-rfc-7231-conneg]. Note that content types provided in the `Accept` header of the request _MAY_ include parameters, for example `profile` or `charset`.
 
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -1679,7 +1679,7 @@ Clients _MUST NOT_ attempt to construct resource URIs by themselves, instead the
 
 ###  6.3. Responses
 
-The format for all responses is JSON, as described above. The different requirements for which resources _MUST_ provide a response is summarized in [Appendix A][prezi30-appendixa]. While some resources do not require their URI to provide the description, it is good practice if possible.
+The format for all responses is JSON, as described above. It is good practice for all resources with an HTTP(S) URI to provide their description when the URI is dereferenced. If the resource is referenced, rather than being embedded in its entirety, then it _MUST_ be able to be dereferenced.
```
jpstroop: >If _the_ resource is referenced

We're not talking about a specific instance, so I think the article should be _a_, instead of _the_:

> If _a_ resource is referenced 

Also, is 'embedded in its entirety' clear, or should it be something more explicit, like, 'contained with the current resource' or something like that?

tomcrane: Agree that could do with a tiny bit more here to indicate that "embedded" is being used in the formal sense defined in https://iiif.io/api/presentation/3.0/#12-terminology

tomcrane: This might just be removing "in its entirety" to match other uses of the term elsewhere in the spec.

azaroth42: Linked "referenced" and "embedded" to the terminology section and removed "in its entirety". Swapped "the" for "a".

azaroth42:
It's not against a protected branch

azaroth42:
Someone other than me can merge...

PR #1893, opened 2019-11-09
Tone down beta warning

### LINKED ISSUE #1891: Tone down warning messages on the BETA drafts of Image and Presentation APIs
We currently have big red warning box:

> Status Warning - This is a work in progress and may change without any notices. Implementers should be aware that this document is not stable. Implementers are likely to find the specification changing in incompatible ways. Those interested in implementing this document before it reaches beta or release stages should join the mailing list and take part in the discussions, and follow the emerging issues on Github.

which is certainly incorrect because these _are_ beta drafts!

zimeon:
Closes #1891 

This is against `master` because we want to update the live beta docs. If merged it might also be good to merge that back into the `image_presi_rc3` branch to avoid any possible later conflicts.


jpstroop:
I added "the" before "final release"

PR #1903, opened 2019-11-27
Use http for CC URIs

### LINKED ISSUE #1874: clarify CC canonical URLs
The new `rights` statement states that URL must be drawn from the set of Creative Commons license URIs, the RightsStatements.org rights statement URIs, etc.

For RightStatements, there's no issue as each license has a canonical URI indicated on each pages with "`URI for this statement:`", example: http://rightsstatements.org/vocab/InC/1.0/ (note the `http`, not `https`), and there's even a nice rdf file accessible at

```
curl -L -H "Accept: text/turtle" http://rightsstatements.org/vocab/1.0/
```

Now, for CC, it's a bit more confusing as there are many URLs that point to very similar pages, and no clear indication of a canonical link. Although when visiting an RDF version ([example](https://creativecommons.org/publicdomain/zero/1.0/rdf)) or the [index](https://creativecommons.org/licenses/index.rdf) it becomes quite clear that the canonical URIs are `http`, which is probably a bit counter intuitive.

I'm thinking that giving an example or pointing to the RDF index could be useful to avoid users using non-canonical URLs

zimeon:
Closes #1874 per https://github.com/IIIF/trc/issues/32

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -263,6 +263,10 @@ The value _MUST_ be a string. If the value is drawn from Creative Commons or Rig
 { "rights": "http://creativecommons.org/licenses/by/4.0/" }
 ```
 
+__Machine actionable URIs and links for users__<br/>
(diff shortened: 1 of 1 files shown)


zimeon:
https://preview.iiif.io/api/issue-1874/api/presentation/3.0/#rights

PR #1904, opened 2019-11-27
remove prezi context in example as all terms are covered by image con…

### LINKED ISSUE #1892: Presentation Context in Image response example
The second example in [section 5.8](https://iiif.io/api/image/3.0/#58-linking-properties) has the Presentation API context.

The requirement is not expressed in the prose of the specification.  Either this is a bug in the example (because the Image API context has all the necessary terms) or a bug in the spec and should be explained as a requirement (because the terms are not in the Image context).

tomcrane:
…text

closes #1892


azaroth42:
@mikeapp @jpstroop can you +1 please :)

PR #1905, opened 2019-11-27
add service example to image, add list of types to annex

### LINKED ISSUE #1869: Show type on service annex and give type for Auth as an example
In the linking properties the table says that a service must have a `type` or `@type` from:

https://iiif.io/api/image/3.0/#58-linking-properties

and links to the Service Registry:

https://iiif.io/api/annex/services/

but most of these examples don't have a type. The text also mentions this service could be used to link to an authentication service but a type is present in the auth spec and auth isn't in the annex services document.

azaroth42:
Closes #1869.


azaroth42:
@tomcrane ?

PR #1907, opened 2019-11-27
add entry for @context

### LINKED ISSUE #1864: Add @context change to 3.0 image change log
Just looking through the change-log and can't see the `@context` change from `http://iiif.io/api/image/2/context.json` to `http://iiif.io/api/image/3/context.json` in:

https://iiif.io/api/image/3.0/change-log/

azaroth42:
Add changelog entry for `@context.  Closes #1864


azaroth42:
:+1: I considered property, but went for field as it's not a /property/ of the resource, it's a meta- thing in the representation ... but no one actually cares about that, so have reverted to property :)

azaroth42:
@tomcrane ?

Issue #1831, opened 2019-04-18
Update Presentation API 3.0 Changelog


azaroth42:
Leave open until TRC issues are approved and included in change-log.

zimeon:
#1832 was a step toward this

zimeon:
Keeping this issue but moving forward to RC3 (hopefully final)

zimeon:
We think these are now up to date but need a review. @azaroth42 and @zimeon will review

zimeon:
We did review and looks good

PR #1908, opened 2019-11-27
Note context change in prezi3

zimeon:
#1864 for presentation change log, see also #1831

Shamelessly cribbing from @azaroth42's #1907


azaroth42:
And ... @tomcrane ?

PR #1942, opened 2020-04-29
clarify URI construction to Prezi API

### LINKED ISSUE #1918: Request directives needs clarification
In the v3 spec, section 6.2 it says:

"Clients must not attempt to construct resource URIs by themselves, instead they must follow links from within retrieved descriptions or elsewhere."

I think this needs further clarification. IIIF viewers must, by necessity, take image IDs such as `"id": "https://example.org/iiif/book1/page1/full/max/0/default.jpg"` and construct derived URLs for the purposes of making tiling work. Any tiling-capable viewer needs to remove the `full` and `max` parameters and rewrite the URL with specific tile sizes, which seems to fall foul of the guidance given here.

A note clarifying request behaviours for clients' dynamically-generated Image API requests would be helpful, particularly since there is now a requirement that the `id` of the Image can no longer simply be a request for the Image service, and it must be a full URL to an image (section 5.7).

azaroth42:
Closes #1918

### DIFF
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1738,7 +1738,7 @@ While any HTTP(S) URI is technically acceptable for any of the resources in the 
 
 ###  6.2. Requests
 
-Clients _MUST NOT_ attempt to construct URIs for Presentation API resources by themselves, instead they _MUST_ follow links from within retrieved descriptions or elsewhere. Other services _MAY_ require constructing URIs according to their specifications.
+Clients are only expected to follow links to Presentation API resources. Unlike [IIIF Image API][image-api] requests, or other parameterized services, the URIs for Presentation API resources cannot be assumed to follow any particular pattern. 
 
 ###  6.3. Responses
 
--- a/source/api/presentation/3.0/index.md
+++ b/source/api/presentation/3.0/index.md
@@ -1738,7 +1738,7 @@ While any HTTP(S) URI is technically acceptable for any of the resources in the 
 
 ###  6.2. Requests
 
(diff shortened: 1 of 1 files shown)


source/api/presentation/3.0/index.md
```diff
@@ -1738,7 +1738,7 @@ While any HTTP(S) URI is technically acceptable for any of the resources in the
 
 ###  6.2. Requests
 
-Clients _MUST NOT_ attempt to construct resource URIs by themselves, instead they _MUST_ follow links from within retrieved descriptions or elsewhere.
+Clients _MUST NOT_ attempt to construct URIs for Presentation API resources by themselves, instead they _MUST_ follow links from within retrieved descriptions or elsewhere. Other services _MAY_ require constructing URIs according to their specifications.
```
jpstroop, line 1742: Do you want to call out the Image API specifically, something like:
>Other IIIF services (e.g. the [Image API](https://iiif.io/api/image)) _MAY_ require ...

zimeon: How about something less normative:

> Clients are expected to follow links to Presentation API resources from within retrieved descriptions or elsewhere. Unlike Image API requests, the URIs for Presentation API resources cannot be assumed to follow any particular pattern.

I don't really like using untestable normative language about clients any more than we have to.

azaroth42:
How about:

> Clients are only expected to follow links to Presentation API resources. Unlike [IIIF Image API][image-api] requests, or other parameterized services, the URIs for Presentation API resources cannot be assumed to follow any particular pattern.

azaroth42:
@jpstroop? With your :+1: we can merge :)

PR #1943, opened 2020-04-29
Clarfiy that color may still be grayscale, etc.

### LINKED ISSUE #1922: clarification on "if applicable"
The [Image API 3.0 compliance level spec](https://iiif.io/api/image/3.0/compliance/#34-quality) has a puzzling and unexplained wording in section `3.4 Quality`, where `color` and `gray` at `level 2` are `required (if applicable)`. What does this mean? Who determines if it's applicable and how is communicated? In other words: if I have a client that fetches an image info and sees that it is compliant with `level 2`, how does it determine if `gray` or `color` are `applicable`?

related to https://github.com/IIIF/api/issues/1920

**Edit:** I suspect this means "if `color` / `gray` are listed in `extraQualities`"? But in that case that makes the info redundant, and the `quality` section of the compliance level useless

jpstroop:
This does not feel right, but is consistent with the decision in https://github.com/IIIF/trc/issues/25. Closes #1922.

### DIFF
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -395,9 +395,7 @@ The `default` quality exists to support [level 0 compliant implementations][imag
 
 Regardless of level of compliance, services that make additional qualities beyond `default` available _MUST_ list those qualities in the [`extraQualities` property][image3-extra-functionality] of the response to the [Image Information Request][image3-information]. A request for an image with an unsupported quality value _SHOULD_ return a 400 (Bad Request) status code.
 
-__Qualities and Color Information__<br />
 A request for the `color` quality of an image _MUST_ return an image. This image may be comprised solely of grayscale or bitonal pixels if this is all of the color information the server has available. The server _SHOULD NOT_ include `color` in the `extraQualities` list in such cases, but a request for the `color` quality _MUST NOT_ result in an error. Similarly, a request for the grayscale quality may return an image comprised solely of bitonal pixels.
-{: .note}
 
 Examples:
 
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -395,7 +395,9 @@ The `default` quality exists to support [level 0 compliant implementations][imag
 
 Regardless of level of compliance, services that make additional qualities beyond `default` available _MUST_ list those qualities in the [`extraQualities` property][image3-extra-functionality] of the response to the [Image Information Request][image3-information]. A request for an image with an unsupported quality value _SHOULD_ return a 400 (Bad Request) status code.
 
-**N.B.** A request for the `color` quality of an image may return an image comprised solely of grayscale or bitonal pixels if this is all of the color information the server has available. The server _SHOULD NOT_ include `color` in the `extraQualities` list when this is a case, but a request of this nature _MUST NOT_ result in an error.
+__Qualities and Color Information__<br />
+A request for the `color` quality of an image may return an image comprised solely of grayscale or bitonal pixels if this is all of the color information the server has available. The server _SHOULD NOT_ include `color` in the `extraQualities` list in such cases, but a request for the `color` quality _MUST NOT_ result in an error. Similarly, a request for the grayscale quality may return an image comprised solely of bitonal pixels.
+{: .note}
 
 Examples:
 
--- a/source/api/image/3.0/index.md
+++ b/source/api/image/3.0/index.md
@@ -395,6 +395,8 @@ The `default` quality exists to support [level 0 compliant implementations][imag
 
 Regardless of level of compliance, services that make additional qualities beyond `default` available _MUST_ list those qualities in the [`extraQualities` property][image3-extra-functionality] of the response to the [Image Information Request][image3-information]. A request for an image with an unsupported quality value _SHOULD_ return a 400 (Bad Request) status code.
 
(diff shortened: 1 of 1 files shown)


azaroth42:
Agree that it is consistent with trc#25 decision. I don't really understand the difference between default and color, apart from the ability that no one has asked for to have a default which is not the same as `color`, but we tried to resolve it already.

jpstroop:
Yes, the semantic difference between `color` and `default` is blurry at best.

`/blurry.jpg`? 🤔

source/api/image/3.0/index.md
```diff
@@ -395,6 +395,8 @@ The `default` quality exists to support [level 0 compliant implementations][imag
 
 Regardless of level of compliance, services that make additional qualities beyond `default` available _MUST_ list those qualities in the [`extraQualities` property][image3-extra-functionality] of the response to the [Image Information Request][image3-information]. A request for an image with an unsupported quality value _SHOULD_ return a 400 (Bad Request) status code.
 
+**N.B.** A request for the `color` quality of an image may return an image comprised solely of grayscale or bitonal pixels if this is all of the color information the server has available. The server _SHOULD NOT_ include `color` in the `extraQualities` list when this is a case, but a request of this nature _MUST NOT_ result in an error result in an error.
```
ahankinson, line 398: "result in an error" is repeated twice at the end of the paragraph.

jpstroop: Oops, thanks. Fixed.

source/api/image/3.0/index.md
```diff
@@ -395,6 +395,8 @@ The `default` quality exists to support [level 0 compliant implementations][imag
 
 Regardless of level of compliance, services that make additional qualities beyond `default` available _MUST_ list those qualities in the [`extraQualities` property][image3-extra-functionality] of the response to the [Image Information Request][image3-information]. A request for an image with an unsupported quality value _SHOULD_ return a 400 (Bad Request) status code.
 
+**N.B.** A request for the `color` quality of an image may return an image comprised solely of grayscale or bitonal pixels if this is all of the color information the server has available. The server _SHOULD NOT_ include `color` in the `extraQualities` list when this is a case, but a request of this nature _MUST NOT_ result in an error.
```
zimeon, line 398: I think to be consistent with formatting we should use a note like "Model Alignment" in https://iiif.io/api/presentation/3.0/#331-external-links , along the lines of:

```
__Qualities Note__<br />
...stuff in here...
{: .note}
```

zimeon: Also, the last sentence is really hard to parse with two "this". Perhaps:

> The server _SHOULD NOT_ include `color` in the `extraQualities` list in such cases, but a request for the `color` quality _MUST NOT_ result in an error.

Should there also be:

> Similarly, a request for the `greyscale` quality may return an image comprised solely of bitonal pixels.

azaroth42:
:+1: to @zimeon's requested changes

mikeapp:
Also 👍

source/api/image/3.0/index.md
```diff
@@ -395,7 +395,9 @@ The `default` quality exists to support [level 0 compliant implementations][imag
...
 
-**N.B.** A request for the `color` quality of an image may return an image comprised solely of grayscale or bitonal pixels if this is all of the color information the server has available. The server _SHOULD NOT_ include `color` in the `extraQualities` list when this is a case, but a request of this nature _MUST NOT_ result in an error.
+__Qualities and Color Information__<br />
+A request for the `color` quality of an image may return an image comprised solely of grayscale or bitonal pixels if this is all of the color information the server has available. The server _SHOULD NOT_ include `color` in the `extraQualities` list in such cases, but a request for the `color` quality _MUST NOT_ result in an error. Similarly, a request for the grayscale quality may return an image comprised solely of bitonal pixels.
+{: .note}
```
jpstroop, line 401: We agree that these are not RFC MAYs, correct? i.e.:

> the `color` quality of an image may return an image comprised solely of grayscale or bitonal pixels
...
> a request for the grayscale quality may return [...] bitonal pixels

If RFC `MAY` is a synonym for OPTIONAL, then using an RFC MAY could imply that the server does not have to return an image, which would contradict:

> a request for the `color` quality _MUST NOT_ result in an error

tomcrane: I agree that they are not RFC MAYs.

Is the first, though, an RFC MUST? To avoid the error the next bit says MUST NOT happen?

... MUST return an image comprised solely of grayscale or bitonal pixels **if** this is all of the color information the server has available. 

(i.e., `color` doesn't mean colorize it :) )

zimeon: How about something like:

> the `color` quality MUST return an image, although the image need not contain color pixels.

zimeon: or to combine with @tomcrane's comment:

| the `color` quality MUST return an image, although the image need not contain color pixels if no color information is available.

jpstroop: >(i.e., color doesn't mean colorize it :) )

Unless it does? I have no idea what that would mean, but I think I'm more comfortable with the fuzzy semantics.

Pushing up another change...

jpstroop: >I have no idea what that would mean
https://arxiv.org/abs/1406.2661

PR #1944, opened 2020-05-01
JS editorial readthrough

jpstroop:
Wordsmithy and typo stuff only

### DIFF
--- a/source/api/discovery/0.5/index.md
+++ b/source/api/discovery/0.5/index.md
@@ -102,21 +102,20 @@ The subsections below describe first how to construct the description of the cha
 
 The core information required to provide a minimally effective set of links to IIIF resources is the URIs of those resources. The order of the resources in the resulting list is unimportant, but each _SHOULD_ only appear once. This is the minimum level for any effective interoperability.
 
+This minimal level 0 API approach is compatible with the levels 1 and 2 which introduce significant benefits that allow clients to better optimize their interactions.
+
 If resources are deleted after being referred to in the resource list, the entire list should be republished without the reference to the deleted resource. Clients should also expect to encounter resource URIs that are out of date and no longer resolve to a IIIF Manifest or Collection.
 
 Example Level 0 Activity:
 
 ```json-doc
-[
-  {
-    "type": "Update",
-    "object": {
-      "id": "https://example.org/iiif/1/manifest",
-      "type": "Manifest"
-    }
-  },
-  // ...
-]
+{
+  "type": "Update",
+  "object": {
+    "id": "https://example.org/iiif/1/manifest",
+    "type": "Manifest"
+  }
+}
 ```
 
 #### 2.1.2. Level 1: Resource Changes
@@ -127,17 +126,14 @@ When dealing with large sets of resources, it can be useful to work with only th
 Example Level 1 Activity:
 
 ```json-doc
-[
-  {
-    "type": "Update",
-    "object": {
-      "id": "https://example.org/iiif/1/manifest",
-      "type": "Manifest"
-    },
-    "endTime": "2017-09-20T00:00:00Z"
+{
+  "type": "Update",
+  "object": {
+    "id": "https://example.org/iiif/1/manifest",
+    "type": "Manifest"
   },
-  // ...
-]
+  "endTime": "2017-09-20T00:00:00Z"
+}
 ```
 
 #### 2.1.3. Level 2: Resource Creation, Change and Deletion
@@ -150,17 +146,14 @@ A complete change history is not required, and sometimes not even desirable. If 
 Example Level 2 Activity:
 
 ```json-doc
-[
-  {
-    "type": "Create",
-    "object": {
-      "id": "https://example.org/iiif/1/manifest",
-      "type": "Manifest"
-    },
-    "endTime": "2017-09-20T00:00:00Z"
+{
+  "type": "Create",
+  "object": {
+    "id": "https://example.org/iiif/1/manifest",
+    "type": "Manifest"
   },
-  // ...
-]
+  "endTime": "2017-09-20T00:00:00Z"
+}
 ```
 
 #### 2.1.4. Aggregating Activities
@@ -177,43 +170,37 @@ Whenever an "Add" Activity is encountered, it is semantically similar to a "Crea
 Example Add Activity:
 
 ```json-doc
-[
-  {
-    "type": "Add",
-    "summary": "Added newly discovered manifest to stream",
-    "object": {
-      "id": "https://example.org/iiif/1/manifest",
-      "type": "Manifest"
-    },
-    "target": {
-      "id": "https://example.org/activity/aggregated-changes",
-      "type": "OrderedCollection"
-    },
-    "endTime": "2019-11-22T00:00:00Z"
+{
+  "type": "Add",
+  "summary": "Added newly discovered manifest to stream",
+  "object": {
+    "id": "https://example.org/iiif/1/manifest",
+    "type": "Manifest"
   },
-  // ...
-]
+  "target": {
+    "id": "https://example.org/activity/aggregated-changes",
+    "type": "OrderedCollection"
+  },
+  "endTime": "2019-11-22T00:00:00Z"
+}
 ```
 
 Example Remove Activity:
 
 ```json-doc
-[
-  {
-    "type": "Remove",
-    "summary": "Removed manifest from stream due to being out of scope",
-    "object": {
-      "id": "https://example.org/iiif/1/manifest",
-      "type": "Manifest"
-    },
-    "origin": {
-      "id": "https://example.org/activity/aggregated-changes",
-      "type": "OrderedCollection"
-    },
-    "endTime": "2017-11-23T00:00:00Z"
+{
+  "type": "Remove",
+  "summary": "Removed manifest from stream due to being out of scope",
+  "object": {
+    "id": "https://example.org/iiif/1/manifest",
+    "type": "Manifest"
   },
-  // ...
-]
+  "origin": {
+    "id": "https://example.org/activity/aggregated-changes",
+    "type": "OrderedCollection"
+  },
+  "endTime": "2017-11-23T00:00:00Z"
+}
 ```
 
 
--- a/source/api/discovery/0.5/index.md
+++ b/source/api/discovery/0.5/index.md
@@ -60,7 +60,7 @@ The objective of the IIIF Change Discovery API is to provide a machine to machin
 
 The discovery of IIIF resources requires a consistent and well understood pattern for content providers to publish lists of links to their available content. This allows a baseline implementation of discovery systems that process the list, looking for resources that have been added or changed.
 
-This process can be optimized by allowing the content providers to publish descriptions of how their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes unavailable. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll multiple sources to see if anything has changed.
+This process can be optimized by allowing the content providers to publish descriptions of how their content has changed, enabling consuming systems to retrieve only the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes unavailable. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll multiple sources to see if anything has changed.
 
 Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. The diverse domains represented within the IIIF community already have successful standards fulfilling these use cases. Also out of scope is optimization of the transmission of content, for example recommendations about transferring any source media or data between systems.
 
@@ -532,7 +532,7 @@ Ordered Collection Pages _MUST_ have a `prev` property, unless they are the firs
 
 The Activities that are listed as part of this page.
 
-Ordered Collection Pages _MUST_ have a `orderedItems` property. The value _MUST_ be an array, with at least one item. Each item _MUST_ be a JSON object, conforming to the requirements of an Activity.
+Ordered Collection Pages _MUST_ have an `orderedItems` property. The value _MUST_ be an array, with at least one item. Each item _MUST_ be a JSON object, conforming to the requirements of an Activity.
 
 ```json-doc
 {
@@ -617,7 +617,7 @@ This specification uses the types described in the table below.
 | `Remove` | The removal of an object from a stream, outside of any of the above types of activity, such as a third party aggregator removing resources from a stream that are no longer considered to be in scope. Each resource _SHOULD_ have at most one `Remove` activity, but if the resource is re-added and then re-removed, then there _MAY_ be more than one. The `Remove` Activity _SHOULD NOT_ be present if a `Delete` Activity is present at the same time.|
 {: .api-table #table-type-dfn}
 
-Activities _MUST_ have the `type` property. The value _MUST_ be a registered Activity class, and _SHOULD_ be one of `Create`, `Update`, or `Delete`.
+Activities _MUST_ have the `type` property. The value _MUST_ be a registered Activity type, and _SHOULD_ be one of `Create`, `Update`, or `Delete`.
 
 ```json-doc
 { "type": "Update" }
@@ -761,7 +761,7 @@ The top level resource in the response _MUST_ have the `@context` property, and 
 
 #### 3.4.2. Extensions
 
-If any additional classes or properties are desired beyond the ones defined in this specification or the ActivityStreams specification, then those classes or properties _SHOULD_ be mapped to RDF terms in one or more additional context documents. These extension contexts _SHOULD_ be added to the top level `@context` property, and _MUST_ be before the URI of the Discovery context. The JSON-LD 1.1 functionality of defining terms only within a specific property, known as [scoped contexts][org-w3c-json-ld-scoped-contexts], _MUST_ be used to minimize cross-extension collisions. Extensions intended for broad use _SHOULD_ be registered in the [extensions registry][registry].
+If any additional types or properties are desired beyond the ones defined in this specification or the ActivityStreams specification, then those types or properties _SHOULD_ be mapped to RDF terms in one or more additional context documents. These extension contexts _SHOULD_ be added to the top level `@context` property, and _MUST_ be before the URI of the Discovery context. The JSON-LD 1.1 functionality of defining terms only within a specific property, known as [scoped contexts][org-w3c-json-ld-scoped-contexts], _MUST_ be used to minimize cross-extension collisions. Extensions intended for broad use _SHOULD_ be registered in the [extensions registry][registry].
 
 ```json-doc
 {
@@ -867,13 +867,13 @@ Given an array (`collections`) of collection URIs as input,
 
 Activities _MAY_ be published about content that has access restrictions. Clients _MUST NOT_ assume that they will be able to access every resource that is the object of an Activity, and _MUST NOT_ assume that it has been deleted if it is inaccessible. For example, the content might be protected by an authentication system that is denying access (an end user might be able to provide the right credentials to gain access) or there may be a temporary network outage preventing the content from being retrieved (the network will eventually be restored).
 
-Content may also change from being available to being protected by access restrictions, or become available having previously been protected. There are no more specific activity classes for these situations, and the publisher _SHOULD_ issue the regular `Update` or `Delete` activities accordingly.
+Content may also change from being available to being protected by access restrictions, or become available having previously been protected. There are no more specific activity types for these situations, and the publisher _SHOULD_ issue the regular `Update` or `Delete` activities accordingly.
 
 ### 4.2. Negotiable Resources
 
 Some HTTP(S) URIs are able to respond with different representations of the same content in response to requests with different headers, such as the same URI being able to return both version 2 and version 3 of the IIIF Presentation API based on the Accept header. This is known as "content negotiation", and such resources are known as "negotiable resources". The representations that can be negotiated for are known as "variants".
 
-Negotiable resources are not supported by the Discovery API, only variants. This means that there would be one Activity entry for each of the representations that are available, and that each representation _MUST_ have its own URI, even if it can also be reached via the negotiable resource. In the case of negotiating for different versions of the IIIF Presentation API, the `format` property can be used to include the full media type of the resource, where the version specific context document is given in the `profile` parameter. If there is additional descriptive resources available, each such resource would describe all of the variants, and thus the `seeAlso` property of each Activity would refer to the same descriptions, allowing the variants to be connected together.
+Negotiable resources are not supported by the Discovery API, only variants. This means that there would be one Activity entry for each of the representations that are available, and that each representation _MUST_ have its own URI, even if it can also be reached via the negotiable resource. In the case of negotiating for different versions of the IIIF Presentation API, the `format` property can be used to include the full media type of the resource, where the version-specific context document is given in the `profile` parameter. If there are additional descriptive resources available, each such resource would describe all of the variants, and thus the `seeAlso` property of each Activity would refer to the same descriptions, allowing the variants to be connected together.
 
 Two variants of the same negotiable resource can be represented as follows.
 
--- a/source/api/discovery/0.5/index.md
+++ b/source/api/discovery/0.5/index.md
@@ -56,15 +56,15 @@ This specification leverages existing techniques, specifications, and tools in o
 ### 1.1. Objectives and Scope
 {: #objectives-and-scope}
 
-The objective of the IIIF Change Discovery API is to provide a machine to machine API that provides the information needed to discover and subsequently make use of IIIF resources.  The intended audience is other IIIF aware systems that can leverage the content and APIs. While this work may benefit others outside of the IIIF community directly or indirectly, the objective of the API is to specify an interoperable solution that best and most easily fulfills the discovery needs within the community of participating organizations.
+The objective of the IIIF Change Discovery API is to provide a machine to machine API that provides the information needed to discover and subsequently make use of IIIF resources. The intended audience is other IIIF-aware systems that can leverage the content and APIs. While this work may benefit others outside of the IIIF community directly or indirectly, the objective of the API is to specify an interoperable solution that best and most easily fulfills the discovery needs within the community of participating organizations.
 
 The discovery of IIIF resources requires a consistent and well understood pattern for content providers to publish lists of links to their available content. This allows a baseline implementation of discovery systems that process the list, looking for resources that have been added or changed.
 
-This process can be optimized by allowing the content providers to publish descriptions of when their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes unavailable. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll all of the systems to see if anything has changed.
+This process can be optimized by allowing the content providers to publish descriptions of how their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes unavailable. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll multiple sources to see if anything has changed.
 
-Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. The diverse domains represented within the IIIF community already have successful standards fulfilling these use cases, and the results of previous attempts to reconcile these standards across domains have not seen widespread adoption. Also out of scope is optimization of the transmission of content, for example recommendations about transferring any source media or data between systems.
+Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. The diverse domains represented within the IIIF community already have successful standards fulfilling these use cases. Also out of scope is optimization of the transmission of content, for example recommendations about transferring any source media or data between systems.
 
-__Notification of Changes__<br>This draft version of the specification does not include a subscription mechanism for enabling change notifications to be pushed to remote systems.  The current specification only enables the polling pattern where the set of changes must be periodically reprocessed. A subscription/notification pattern might be added in a future version after implementation experience with the polling pattern has demonstrated that it would be valuable.
+__Change Notifications__<br>This specification does not include a subscription mechanism for enabling change notifications to be pushed to remote systems. Only periodic polling for the set of changes that must be processed is supported. A subscription/notification pattern may be added in a future version after implementation experience with the polling pattern has demonstrated that it would be valuable.
 {: .warning}
 
 ### 1.2. Terminology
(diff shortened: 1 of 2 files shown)


source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
+  {
+    "type": "Update",
+    "object": {
+      "id": "https://example.org/iiif/1/manifest",
+      "type": "Manifest"
+    }
+  },
+  // ...
+]
 ```
```
jpstroop, line 120: The 2.1.x text refers to lists over and over; seems like the examples should be lists as well.

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
+Secondly, resources might become available or unavailable without being created or deleted. This could happen because of changes to their permissions, such as an embargo period after creation, or when resources are temporarily removed in order to be edited as part of some data cleaning or migration process.
 
-Whenever an "Add" Activity is encountered, it is similar in nature to a "Create" Activity in that it is the first time that resource is mentioned in the stream. It is, therefore, expected that if there is an "Add", then there may not be a "Create" for the same resource occurring before it. Similarly, there will be no more references to the resource after a "Remove" activity, in the same way as for "Delete". 
+Whenever an "Add" Activity is encountered, it is semantically similar to a "Create" Activity in that it is the first time that resource is mentioned in the stream. It is, therefore, expected that if there is an "Add", then there may not be a "Create" for the same resource occurring before it. Similarly, there will be no more references to the resource after a "Remove" or "Delete".
```
jpstroop, line 176: It's a little strange to me that Add and Remove come out of nowhere here after talking about C, U, D, but I'm not sure a forward reference to `type` (where the list of supported activities lives) is appropriate either.

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
-  "type": "Add",
-  "summary": "Added newly discovered manifest to stream",
-  "object": {
-    "id": "https://example.org/iiif/1/manifest",
-    "type": "Manifest"
-  },
-  "target": {
-    "id": "https://example.org/activity/aggregated-changes",
-    "type": "OrderedCollection"
+```json-doc
```
jpstroop, line 179: Note that I added `json-doc` everywhere, for consistency with our other markdown docs.

source/api/discovery/0.5/index.md
```diff
@@ -279,15 +290,15 @@ As the number of Activities is likely too many to usefully be represented in a s
 
 The W3C [Activity Streams][org-w3c-activitystreams] specification defines a "model for representing potential and completed activities", and is compatible with the [design patterns][annex-patterns] established for IIIF APIs. It is defined in terms of [JSON-LD][org-w3c-json-ld] and can be seamlessly integrated with the existing IIIF APIs. The model can be used to represent activities carried out by content publishers of creating, updating, and deleting (or otherwise de-publishing) IIIF resources.
 
-This section is a summary of the properties and types used by this specification, and defined by Activity Streams.  This is intended to ease implementation efforts by collecting the relevant information together.
```
jpstroop, line 282: I see you sneaking in double spaces after stops. Don't think I won't catch you ;-)

source/api/discovery/0.5/index.md
```diff
@@ -279,15 +290,15 @@ As the number of Activities is likely too many to usefully be represented in a s
...
+Properties, beyond those described in this specification, that the consuming application does not have code to process _MUST_ be ignored. Other properties defined by Activity Streams _MAY_ be used, such as `origin` or `instrument`, but there are no current use cases that would warrant their inclusion in this specification.
 
 
 ### 3.1. Ordered Collection
 {: #orderedcollection}
 
-The top-most resource for managing the lists of Activities is an Ordered Collection, broken up into Ordered Collection Pages. This is the same pattern that the Web Annotation model uses for Annotation Collections and Annotation Pages. The Collection does not directly contain any of the Activities, instead it refers to the `first` and `last` pages of the list.
+The top-most resource for managing the lists of Activities is an Ordered Collection, broken up into Ordered Collection Pages. The Collection does not directly contain any of the Activities, instead it refers to the `first` and `last` pages of the list.
```
jpstroop, line 301: Web Annotation isn't mentioned anywhere else. Seems like a wasted sentence.

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
 
 #### 2.1.1. Level 0: Resource References
 {: #level-0-basic-resource-list}
 
-The core information required to provide a minimally effective set of links to IIIF resources is just the URIs of those resources. However, with the addition of some additional JSON structure wrapped around those URIs, the document can also be on the path towards a robust set of information that clients can use to optimize their processing.
-
-Starting with the IIIF resource URIs, we add an "Update" Activity wrapper around them.  The order of the resources in the resulting list is unimportant, but each should only appear once. In terms of optimization, this approach provides no additional benefit over any other simpler list format, but is compatible with the following levels which introduce significant benefits.  This is the minimum level for any effective interoperability.
+The core information required to provide a minimally effective set of links to IIIF resources is the URIs of those resources. The order of the resources in the resulting list is unimportant, but each _SHOULD_ only appear once. This is the minimum level for any effective interoperability.
```
jpstroop, line 103: This read too much like a tutorial for me, and the tone wasn't consistent w/ the rest of the spec.

source/api/discovery/0.5/index.md
```diff
@@ -446,27 +458,28 @@ The identifier of the Collection Page.
...
 
-```
+```json-doc
 { "type": "OrderedCollectionPage" }
 ```
 
 ##### partOf
 
-The Ordered Collection that this Page is part of.
+The Ordered Collection of which this Page is a part.
```
jpstroop, line 478: I've managed to keep us from ending sentences w/ prepositions up to now...

source/api/discovery/0.5/index.md
```diff
@@ -662,39 +676,39 @@ The new location of the IIIF resource, after it was affected by a `Move` activit
...
+```json-doc
 { "summary": "admin updated the manifest, fixing reported bug #15." }
 ```
 
 ##### actor
 
 The organization, person, or software agent that carried out the Activity.
 
-Activities _MAY_ have the `actor` property.  The value _MUST_ be a JSON object, with the `id` and `type` properties. The `id` _SHOULD_ be an HTTP(S) URI. The `type` _MUST_ be one of `Application`, `Organization`, or `Person`.
+Activities _MAY_ have the `actor` property. The value _MUST_ be a JSON object, with the `id` and `type` properties. The `id` _SHOULD_ be an HTTP(S) URI. The `type` _MUST_ be one of `Application`, `Organization`, or `Person`.
```
jpstroop, line 709: Just confirming: no `name` property?  This is inconsistent with https://www.w3.org/TR/activitystreams-vocabulary/#actor-types

source/api/discovery/0.5/index.md
```diff
@@ -56,15 +56,15 @@ This specification leverages existing techniques, specifications, and tools in o
...
 
-This process can be optimized by allowing the content providers to publish descriptions of when their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes unavailable. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll all of the systems to see if anything has changed.
+This process can be optimized by allowing the content providers to publish descriptions of how their content has changed, enabling consuming systems to only retrieve the resources that have been modified since they were last retrieved. These changes might include when content is deleted or otherwise becomes unavailable. Finally, for rapid synchronization, a system of notifications pushed from the publisher to a set of subscribers can reduce the amount of effort required to constantly poll multiple sources to see if anything has changed.
```
mikeapp, line 63: Small additional suggestion: `to only retrieve` -> `to retrieve only`

source/api/discovery/0.5/index.md
```diff
@@ -851,19 +865,19 @@ Given an array (`collections`) of collection URIs as input,
...
+Activities _MAY_ be published about content that has access restrictions. Clients _MUST NOT_ assume that they will be able to access every resource that is the object of an Activity, and _MUST NOT_ assume that it has been deleted if it is inaccessible. For example, the content might be protected by an authentication system that is denying access (an end user might be able to provide the right credentials to gain access) or there may be a temporary network outage preventing the content from being retrieved (the network will eventually be restored).
 
-Content may also change state from being available to being protected by access restrictions, or become available having previously been protected. There are no new activity classes for these situations, and the publisher _SHOULD_ issue the regular `Update` or `Delete` activities, if that is the behavior desired for harvesters.
+Content may also change from being available to being protected by access restrictions, or become available having previously been protected. There are no more specific activity classes for these situations, and the publisher _SHOULD_ issue the regular `Update` or `Delete` activities accordingly.
```
jpstroop, line 870: Should it be "activity classes" or "activity types"?

source/api/discovery/0.5/index.md
jpstroop: Note also the addition of an RFC _SHOULD_

source/api/discovery/0.5/index.md
```diff
@@ -56,15 +56,15 @@ This specification leverages existing techniques, specifications, and tools in o
 ### 1.1. Objectives and Scope
 {: #objectives-and-scope}
 
-The objective of the IIIF Change Discovery API is to provide a machine to machine API that provides the information needed to discover and subsequently make use of IIIF resources.  The intended audience is other IIIF aware systems that can leverage the content and APIs. While this work may benefit others outside of the IIIF community directly or indirectly, the objective of the API is to specify an interoperable solution that best and most easily fulfills the discovery needs within the community of participating organizations.
+The objective of the IIIF Change Discovery API is to provide a machine to machine API that provides the information needed to discover and subsequently make use of IIIF resources. The intended audience is other IIIF-aware systems that can leverage the content and APIs. While this work may benefit others outside of the IIIF community directly or indirectly, the objective of the API is to specify an interoperable solution that best and most easily fulfills the discovery needs within the community of participating organizations.
```
azaroth42, line 59: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -519,9 +532,9 @@ Ordered Collection Pages _MUST_ have a `prev` property, unless they are the firs
 
 The Activities that are listed as part of this page.
 
-Ordered Collection Pages _MUST_ have a `orderedItems` property.  The value _MUST_ be an array, with at least one item.  Each item _MUST_ be a JSON object, conforming to the requirements of an Activity.
+Ordered Collection Pages _MUST_ have a `orderedItems` property. The value _MUST_ be an array, with at least one item. Each item _MUST_ be a JSON object, conforming to the requirements of an Activity.
```
mikeapp, line 536: 'have a' -> 'have an'

source/api/discovery/0.5/index.md
azaroth42, line 63: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -56,15 +56,15 @@ This specification leverages existing techniques, specifications, and tools in o
...
 
-Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. The diverse domains represented within the IIIF community already have successful standards fulfilling these use cases, and the results of previous attempts to reconcile these standards across domains have not seen widespread adoption. Also out of scope is optimization of the transmission of content, for example recommendations about transferring any source media or data between systems.
+Work that is out of scope of this API includes the recommendation or creation of any descriptive metadata formats, and the recommendation or creation of metadata search APIs or protocols. The diverse domains represented within the IIIF community already have successful standards fulfilling these use cases. Also out of scope is optimization of the transmission of content, for example recommendations about transferring any source media or data between systems.
```
azaroth42, line 65: :-1: This is from the discussions that led to the Discovery for Humans CG. It's there to cut short any proposal that a reconciliation of existing standards is not out of scope.

source/api/discovery/0.5/index.md
```diff
@@ -56,15 +56,15 @@ This specification leverages existing techniques, specifications, and tools in o
...
 
-__Notification of Changes__<br>This draft version of the specification does not include a subscription mechanism for enabling change notifications to be pushed to remote systems.  The current specification only enables the polling pattern where the set of changes must be periodically reprocessed. A subscription/notification pattern might be added in a future version after implementation experience with the polling pattern has demonstrated that it would be valuable.
+__Change Notifications__<br>This specification does not include a subscription mechanism for enabling change notifications to be pushed to remote systems. Only periodic polling for the set of changes that must be processed is supported. A subscription/notification pattern may be added in a future version after implementation experience with the polling pattern has demonstrated that it would be valuable.
```
azaroth42, line 67: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
+In order to discover IIIF resources, the state of those resources in the systems that publish them needs to be communicated succinctly and easily to a consuming system. The consuming system can then use that information to retrieve and process the resources of interest and provide platforms that enable discovery and use. This communication takes place via the IIIF Change Discovery API, which uses the W3C [Activity Streams][org-w3c-activitystreams] specification to describe and serialize changes to resources. The conceptual model for those changes and the interactions between publishers and consumers are based on those identified by the [ResourceSync][org-openarchives-rsync] framework. It is not necessary to understand either of these standards in order to use or implement the IIIF Change Discovery API, which is fully specified in this document.
```
azaroth42, line 85: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -851,19 +865,19 @@ Given an array (`collections`) of collection URIs as input,
...
+Negotiable resources are not supported by the Discovery API, only variants. This means that there would be one Activity entry for each of the representations that are available, and that each representation _MUST_ have its own URI, even if it can also be reached via the negotiable resource. In the case of negotiating for different versions of the IIIF Presentation API, the `format` property can be used to include the full media type of the resource, where the version specific context document is given in the `profile` parameter. If there is additional descriptive resources available, each such resource would describe all of the variants, and thus the `seeAlso` property of each Activity would refer to the same descriptions, allowing the variants to be connected together.
```
mikeapp, line 876: "If there is ... resources available" -> "If there are ... "

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
+Activities are used to describe the state of the publishing system by recording each individual change, in the order that they occur. The changes described are the creation, modification and deletion of IIIF Presentation API resources. A consuming application that is aware of all of the changes that took place in the publishing system has full knowledge of the set of resources available. IIIF [Collection][prezi30-collection]s and [Manifest][prezi30-manifest] are the main access points to published content and references to descriptive metadata about that content, however, Activities describing changes to other resources, such as the IIIF Image API endpoints or the descriptive metadata about the object represented, could also be published in this way. As the intended use is to inform a consuming system, rather than to display information, the Change Discovery API does not include any content intended to be rendered to end users.
```
azaroth42, line 87: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
+The Presentation API does not include descriptive metadata fields, and intentionally lacks the semantics needed to implement advanced or fielded search. Instead, the Presentation API uses the [`seeAlso`][prezi30-seealso] property to link to external documents that can have richer and domain-specific information about the content being presented. For example, a museum object might have a `seeAlso` reference to a CIDOC-CRM or LIDO description, while a bibliographic resource might reference a Dublin Core or MODS description. These external descriptions should be used when possible to provide interfaces giving access to more precise matching algorithms.
```
azaroth42, line 89: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
 
-This specification describes three levels of conformance that build upon each other in terms of functionality enabled and precision of the information published. Sets of changes are published in pages, which are then collected together in a collection per publisher. To reduce barriers to entry, care has been taken to allow the implementation of all levels using only static files on a web server, rather than requiring dynamic access to a database.
+This specification describes three levels of conformance that build upon each other in terms of functionality and precision of the information published. Sets of changes are published in pages, which are then aggregated into a collection per publisher. To reduce barriers to entry, care has been taken to allow for the possibility of implementing all levels using only static files on a web server, rather than requiring dynamic access to a database.
```
azaroth42, line 91: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
 
 ### 2.1. Listing Resources and their Changes
 {: #listing-resources-and-their-changes}
 
-There are three levels of conformance at which changes can be described. Level 0 is simply a list of the resources available.  Level 1 adds timestamps and ordering from earliest change to most recent, allowing a consuming application to work backwards through the list and stop processing once it encounters a change that it has already seen from a previous run. Level 2 adds information about the types of activities, enabling the explicit description of the creation and deletion of resources.
+There are three levels of conformance at which changes can be described. Level 0 is simply a list of the resources available. Level 1 adds timestamps and ordering from earliest change to most recent, allowing a consuming application to work backwards through the list and stop processing once it encounters a change that it has already seen from a previous run. Level 2 adds information about the types of activities, enabling the explicit description of the creation and deletion of resources.
```
azaroth42, line 96: :+1:

source/api/discovery/0.5/index.md
mikeapp: "version specific" -> "version-specific"

source/api/discovery/0.5/index.md
azaroth42: A bit tutorial-like but it provides the rationale for the inclusion of the level at all, and a way to understand the approach. I think the issue is that the two paragraphs both have bits of normative spec-y text and introductory/background/rationale text.

How about your paragraph, plus:

> This approach provides no additional benefit over any simple list format, however is compatible with the following levels which introduce significant benefits. By implementing a minimal level 0 API, a provider can make a start towards later publishing a more robust and useful API that allows for clients to better optimize their interactions.

source/api/discovery/0.5/index.md
azaroth42: -0. With an array, the example is an array with an update resource in it. I would then expect some property that has the array as its value ... which exists (`items`) but isn't the point of this part of the spec. Faint unease about this change.

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
+When dealing with large sets of resources, it can be useful to work with only those resources that have changed since the last time the list was processed. This can be facilitated by the addition of a time stamp that indicates when a resource was last modified or initially created. This is included using the `endTime` property, representing the time at which the activity of publishing the resource was finished. Lists with multiple activities are then ordered such that the most recent activities occur last. Consumers will then process the list of Activities in reverse order, from last to first, stopping when they encounter an Activity they have already processed.
```
azaroth42, line 125: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
 
-A complete change history is not required, and sometimes not even desirable. If there are many, frequent changes to a resource, then an implementation _MAY_ omit any number of individual changes, but _SHOULD_ always  have the most recent change included in the list. Changes that are deemed to be insignificant to the publisher of the list _MAY_ also be omitted, such as changes that affect only the syntax of the document but not the content.
+A complete change history is not required, and sometimes not even desirable. If there are many or frequent changes to a resource, an implementation _MAY_ omit any number of individual changes, but _SHOULD_ always  have the most recent change included in the list. Changes that are deemed to be insignificant to the publisher of the list _MAY_ also be omitted, such as changes that affect only the syntax of the document but not the content.
```
azaroth42, line 148: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
-}
+  // ...
+]
 ```
 
 #### 2.1.4. Aggregating Activities
 {: #aggregated-activity-streams}
 
-Activities can be added to or removed from streams at times other than when the resources they refer to are created, modified, deleted or moved. There are several use cases for this pattern, however they are all outside of the normal publication of a stream by the content owners and thus are unlikely to affect most publishing implementations. They do affect consuming applications, and the implications are documented in the section on processing streams.
+Activities can be added to or removed at times other than when the resources they refer to are created, modified, deleted or moved. There are several use cases for this pattern, however they are all outside of the normal publication of a stream by the content owners and thus are unlikely to affect most publishing implementations. They do affect consuming applications, and the implications are documented in [3.5. Activity Streams Processing Algorithm][discovery-processing-streams].
```
azaroth42, line 169: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -82,127 +82,138 @@ The key words _MUST_, _MUST NOT_, _REQUIRED_, _SHALL_, _SHALL NOT_, _SHOULD_, _S
...
 
-The first use case is when third-party aggregators add activities harvested from different institutions, and likely filtered according to some criteria, into a composite stream. For example, a dinosaur aggregator might filter several natural history museum streams for only those manifests that are about dinosaurs.
+The first use case is when third-party aggregators add activities harvested from different institutions, filter those activities or objects, and aggregate them into a composite stream. For example, a dinosaur aggregator might filter several natural history museum streams for only those manifests that are about dinosaurs.
```
azaroth42, line 171: Mostly :+1:  The aggregator might not filter at all, but I don't think it makes the description invalid or even unlikely. It's a motivating use case, and the filtering is more probable than not.

source/api/discovery/0.5/index.md
azaroth42: Yeah. Add and Remove came pretty late in the process, so are probably not as well integrated as the rest of the text. I don't think there's a simple fix ... but I'm also not sure that there's a problem.

source/api/discovery/0.5/index.md
azaroth42: :+1:

source/api/discovery/0.5/index.md
azaroth42: :D Fingers are programmed for 30+ years to add double spaces. Hard to revert!

source/api/discovery/0.5/index.md
azaroth42: True enough! It was a precedent that was discussed, but isn't important to understand the spec.  :+1:

source/api/discovery/0.5/index.md
```diff
@@ -373,11 +385,11 @@ Ordered Collections _MAY_ have a `seeAlso` property.  The value _MUST_ be an arr
 
 ##### partOf
 
-This property is used to refer to a larger Ordered Collection, of which this Ordered Collection is part. This would allow a publisher to have thematic or temporal sets of activities, for example to have different collections of activities for their paintings from their sculptures, or their modern content from their archival.
+This property is used to refer to a larger Ordered Collection, of which this Ordered Collection is part. This would allow a publisher to have thematic or temporal sets of activities, for example to have different collections of activities for their paintings from their sculptures, or their modern content from their archival content.
```
azaroth42, line 388: :+1:

source/api/discovery/0.5/index.md
azaroth42: :)

azaroth42: My sentence was influenced by the name of the property, I expect.  :+1: to getting rid of dangling prepositions!

source/api/discovery/0.5/index.md
```diff
@@ -662,39 +676,39 @@ The new location of the IIIF resource, after it was affected by a `Move` activit
...
 
-```
+```json-doc
 { "startTime": "2017-09-20T23:58:00Z" }
 ```
 
 ##### summary
 
-A short textual description of the Activity. This is intended primarily to be used for debugging purposes or explanatory messages, and thus does not have the requirement for internationalization of the content. The intended audience of this field is a software engineer looking at the data, not an end user of the IIIF resource. It is helpful, but not even recommended that the field be present as in most situations, it will never be seen.
+A short textual description of the Activity. This is intended primarily to be used for debugging purposes or explanatory messages, and thus does not have the requirement for internationalization of the content. The intended audience of this field is a human who manages an aggregation service and is looking at the data, not an end user of the IIIF resource. It is not recommended that this field be present as in most situations it will never be seen.
```
azaroth42, line 697: :+1: to the first change. Perhaps just delete the last sentence completely?

source/api/discovery/0.5/index.md
azaroth42: Pretty sure we had `name` and took it out. Can't add back in before 0.9.1 if it is needed, regardless. Should open an issue to track.

source/api/discovery/0.5/index.md
```diff
@@ -705,9 +719,9 @@ Activities _MAY_ have the `actor` property.  The value _MUST_ be a JSON object,
 
 ##### Complete Activity Example
 
-A complete example Activity would thus look like the following example. Note that it does not have a `@context` property, as it is always embedded within a `CollectionPage`.  Please note also that this is a complete example with all of the fields, and most implementations will not need nor expose this level of data.
+A complete example Activity would thus look like the following example. Note that it does not have a `@context` property, as it is always embedded within a `CollectionPage`. Please note also that this is a complete example with all possible fields; most implementations will not need nor expose this level of data.
```
azaroth42, line 722: :+1:

source/api/discovery/0.5/index.md
```diff
@@ -826,12 +840,12 @@ In order to process multiple streams simultaneously, the steps for processing a
...
 
-The algorithm below does not take into account any optimizations or constraints around memory or processing. A more sophisticated algorithm might process pages, relying on each being sorted, to keep an interleaved buffer of activities full, rather than retrieving and processing everything before starting to process the activities. It is not described in as much details as the processing algorithms above for this reason.
+The algorithm below does not take into account any optimizations or constraints around memory or processing. A more sophisticated algorithm might process pages, relying on each being sorted, to keep an interleaved buffer of activities full, rather than retrieving and processing everything before starting to process the activities. It is not described in as much detail as the processing algorithms above for this reason.
```
azaroth42, line 843: :+1:

source/api/discovery/0.5/index.md
azaroth42: It should be consistent, but I seem to have used them interchangeably. Under the Activity section, I say that `type` is the "activity type", but for the `CollectionPage` (for example) I say it is a Class. Happy with the changes :+1:

source/api/discovery/0.5/index.md
```diff
@@ -851,19 +865,19 @@ Given an array (`collections`) of collection URIs as input,
 
 ### 4.1. Activities for Access-Restricted Content
 
-Activities _MAY_ be published about content that has access restrictions. Clients _MUST NOT_ assume that they will be able to access every resource that is the object of an Activity, and _MUST NOT_ assume that if they cannot access it, then it has been deleted and therefore remove it from their indexes.  For example, the content might be protected by an authentication system that is denying access, or there might simply be a temporary network outage preventing the content from being retrieved. An end user might be able to provide the right credentials to gain access.
+Activities _MAY_ be published about content that has access restrictions. Clients _MUST NOT_ assume that they will be able to access every resource that is the object of an Activity, and _MUST NOT_ assume that it has been deleted if it is inaccessible. For example, the content might be protected by an authentication system that is denying access (an end user might be able to provide the right credentials to gain access) or there may be a temporary network outage preventing the content from being retrieved (the network will eventually be restored).
```
azaroth42, line 868: :+1:

source/api/discovery/0.5/index.md
jpstroop: `orderedItems`, right? Anyway, curious what others think; it really stood out to me that the the whole section was about lists of changes but the examples weren't. Maybe others find it less jarring.

source/api/discovery/0.5/index.md
jpstroop: To me it just sounds like it's trying to justify or apologize for the extra structure at level 0, or suggest that 0 isn't actually useful. What actually makes it any easier/more practical to implement? You can do it all with filesystem metadata?

source/api/discovery/0.5/index.md
jpstroop: I don't have an opinion about having it or not, was just noting the divergence from the AS spec examples

source/api/discovery/0.5/index.md
jpstroop: "classes" is only in the spec three times and replacing w/ "types" makes sense. "class" (singular) is only used when talking about a type definition, e.g. 

> The `type` _SHOULD_ be a class defined ...

Which I think make sense. Next push will make the change.

source/api/discovery/0.5/index.md
jpstroop: > and the results of previous attempts to reconcile these standards across domains have not seen widespread adoption

...is what was cut. I think the point is still there and that clause is a little too didactic.

jpstroop: It still says very clearly that inventing new descriptive metadata specs is out of scope

source/api/discovery/0.5/index.md
jpstroop: It makes sense once you've read the whole spec, but the use cases for C U D are elaborated on much more than A and R. A and R could be described up here as well?

jpstroop:
Does anyone (@glenrobson ?) understand why the build is failing? `links.md` is there?

source/api/discovery/0.5/index.md
mikeapp: Also -0.  It didn't strike me as odd, probably because the AS spec does it for examples of an individual, e.g. https://www.w3.org/TR/activitystreams-core/#activities

source/api/discovery/0.5/index.md
azaroth42: It doesn't have to be ordered, and you don't need to know the last time it was changed, just that it exists.  So yes, it is a bit of an apology, as all you're really conveying is an unordered set of URIs, which could have been a plain text file with one URI per line. It's also much less useful than level 1 or 2, as you need to check the entire list every time.  Thus both of those comments are very much warranted, but intentionally so.

source/api/discovery/0.5/index.md
zimeon: I don't think the array works and note that the titles of these examples all say "Example Level 1 Activity" (singular) which doesn't match having a list.

I think the best thing would be to make a forward reference from the first example to  "See [pages of changes](https://iiif.io/api/discovery/0.3/#pages-of-changes) for how these activity objects are assembled into a list." or similar

source/api/discovery/0.5/index.md
zimeon: Rob's para with less apology:

> This minimal level 0 API approach is compatible with the levels 1 and 2 which introduce significant benefits that allow clients to better optimize their interactions.

source/api/discovery/0.5/index.md
zimeon: :+1: I think it still makes the point clearly

source/api/discovery/0.5/index.md
jpstroop: I added this in the latest push

source/api/discovery/0.5/index.md
jpstroop: I removed the array wrappers in the latest push.

jpstroop: If nobody else found it strange to be showing objects while talking about lists, I don't think we need a forward ref; it was probably just me.

source/api/discovery/0.5/index.md
mikeapp: 👍  I agree, makes the point.

jpstroop:
@IIIF/editors What else do you need from me?

zimeon:
In my opinion there is nothing in here that challenges the TRC approval of https://github.com/IIIF/trc/issues/36. I think it is up to @azaroth42 as the primary editor of `discovery-0.5` to either discuss more or merge into the `discovery-0.5` branch. We all then need to review/approve https://github.com/IIIF/api/pull/1926 before final merge

PR #1947, opened 2020-05-13
Add another `color` issue to change log

tomcrane:


zimeon:
:+1: added late since mine was missed

PR #1948, opened 2020-05-13
update 3.0 text for final release

azaroth42:
Update version numbers. Still need final publication date and version names.

PR #1952, opened 2020-06-04
Publish Image API 3.0 and Presentation API 3.0

zimeon:
Per https://github.com/IIIF/trc/issues/37 there is agreement to publish 3.0 final for both Image and Presentation APIs. All necessary metadata and date changes were made in https://github.com/IIIF/api/pull/1948


glenrobson:
Actually I may have been to premature with my approval. Currently this forwards you:

https://preview.iiif.io/api/image_prezi_rc4/api/presentation/

on to 2.1 rather than 3. Same with:

https://preview.iiif.io/api/image_prezi_rc4/api/image

I think they should both forward on to version 3 now this is the latest live version.

zimeon:
:+1: from me, I can't approve as the original PR author

PR #1957, opened 2020-07-04
Cleaning up and updating Jekyll

glenrobson:
* Updating Jekyll to latest
 * Updating Ruby
 * Changing tests so as not to test external links
 * Removing data files used in the website
 * Moving to GitHub Actions for deployment


glenrobson:
Issues:
 * Turned off link checking as there were a number of internal links that were broken.

PR #1963, opened 2020-09-29
Updates for Change Discovery 0.9.1

azaroth42:
Fixes all open change discovery issues:

* https://github.com/IIIF/discovery/issues/48
* https://github.com/IIIF/discovery/issues/61
* https://github.com/IIIF/discovery/issues/74

(other than registry in 74)


zimeon (approved):

source/api/discovery/0.9/index.md
```diff
@@ -121,7 +121,7 @@ Example Level 0 Activity:
...
+When dealing with large sets of resources, it can be useful to work with only those resources that have changed since the last time the list was processed. This is managed by the addition of a time stamp in the `endTime` property, that indicates the time at which the resource was last modified or initially created. Lists with multiple activities _MUST_ be ordered such that the most recent activities occur last. Consumers will then process the list of Activities in reverse order, from last to first, stopping when they encounter an Activity they have already processed.
```
zimeon, line 124: Not a blocker but I wonder whether _"...time at which the resource was last modified or initially created"_ might be better written as _"...time at which the resource was last modified (which could be its creation)"_

azaroth42: I vaguely prefer calling both out separately, as there's `Update` and `Create` activity types which these allude to. Anyone want to provide a tie breaker opinion?

mikeapp: I'd also have a preference for the text in the PR.

jpstroop (dismissed):

mikeapp (approved):

source/api/discovery/0.9/index.md
zimeon: I think the problem I have reading it is that one seems to be given a choice: use the lastmod or creation. When actually we mean use creation only if there have been no subsequent mods. How about: _"......time at which the resource was last modified or initially created (if there have been no subsequent updates)"_

aisaac:
A note that the [last commit](https://github.com/IIIF/api/pull/1963/commits/881ff8ddaccd1b5ce9ff66b2bd4caf6d66141670) is for https://github.com/IIIF/discovery/issues/80

aisaac:
I think there's a missing comma after `"label": { "en": [ "Object Description in Schema.Org" ] } `in the object's `seeAlso`

source/api/discovery/0.9/index.md
```diff
@@ -616,21 +630,34 @@ The IIIF resource that was affected by the Activity. It is an implementation dec
...
-    "id": "http://example.org/iiif/1/manifest",
+    "id": "http://example.org/iiif/v3/1/manifest",
     "type": "Manifest",
+    "canonical": "https://example.org/iiif/1",
     "seeAlso": [
       {
         "id": "https://example.org/dataset/single-item.jsonld",
         "type": "Dataset",
-        "format": "application/ld+json"
+        "label": { "en": [ "Object Description in Schema.Org" ] }
```
glenrobson, line 651: Missing a comma at the end of the line.

tomcrane (dismissed):

zimeon (approved):

jpstroop (approved):

mikeapp (approved):

tomcrane (approved):

Issue #86, opened 2014-05-12
Search results


jpstroop:
When discussing search, talk about Presentation API collection syntax for result.

azaroth42:
Talked about it. Using layers / annotationlists.

PR #1978, opened 2021-03-11
add 0.4 with content-type definition, towards discovery issue #86

azaroth42:
Add content-type towards discovery#86

### DIFF
--- a/source/api/content-state/0.4/index.md
+++ b/source/api/content-state/0.4/index.md
@@ -1,0 +1,701 @@
+---
+title: "IIIF Content State API 0.4"
+title_override: "IIIF Content State API 0.4"
+id: discovery-api-content-state
+layout: spec
+cssversion: 3
+tags: [specifications, presentation-api]
+major: 0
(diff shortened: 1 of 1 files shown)


source/api/content-state/0.4/index.md
```diff
@@ -0,0 +1,701 @@
...
+    <p>...she has often seen a <b>cat</b> without a grin but never a grin without a <b>cat</b></p>
+  </li>
+  <!-- ... more results -->
+</ol>
+```
+
+
+## 5. Media Type of Content State Documents
+
+The format for all content state responses is JSON, as described above. It is good practice for all resources with an HTTP(S) URI to provide their description when the URI is dereferenced.
```
azaroth42, line 662: This should be clearer here or somewhere else that all responses use Presentation API context as param.

mcwhitaker (approved):

zimeon:
Closed in favor of https://github.com/IIIF/api/pull/1986

PR #1976, opened 2021-03-04
Search API --> Content Search API

julsraemy:
To avoid confusion and also to use the full name, I would rename Search API to Content Search API on the API Specifications page.

PR #1980, opened 2021-03-15
Reorder the APIs in alphabetical order

julsraemy:
After the merge of #1976, I noticed that the APIs are no longer in alphabetical order (my bad). 
In this commit, I didn't change the order of the translated APIs in Japanese as it has never been ordered that way (but perhaps it could).

### DIFF
--- a/source/api/index.md
+++ b/source/api/index.md
@@ -11,9 +11,10 @@ cssversion: 2
 | API                | Current Version |
 | ------------------ | --------------- |
 | Authentication API | [{{ site.auth_api.stable.major }}.{{ site.auth_api.stable.minor }}.{{ site.auth_api.stable.patch}}][auth{{ site.auth_api.stable.major }}{{ site.auth_api.stable.minor }}] |
+| Content Search API | [{{ site.search_api.stable.major }}.{{ site.search_api.stable.minor }}.{{ site.search_api.stable.patch }}][search{{ site.search_api.stable.major }}{{ site.search_api.stable.minor }}] |
 | Image API          | [{{ site.image_api.stable.major }}.{{ site.image_api.stable.minor }}.{{ site.image_api.stable.patch}}][image{{ site.image_api.stable.major }}{{ site.image_api.stable.minor }}] |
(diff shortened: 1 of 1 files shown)


mcwhitaker (approved):

zimeon (approved):

mikeapp (approved):

jpstroop:
I'm not sure that I agree with this. I agree that the current order doesn't make sense, but Prezi and Image are really the brand and butter of IIIF, and better, I think, to lead with those than Auth, which is complicated and broken.

kirschbombe:
I agree with @jpstroop that Image and Prezi should come first since these are foundational.

jpstroop:
Maybe we go chronological from birth? (Image, Prezi, Search, Auth, I think it would be?)

mcwhitaker:
I will be honest: I often find unordered lists messy. And suggesting a chronological order implies that the users of the list would know the birth date of each, which is not a good assumption.

Is there another way we could represent the primacy of the "foundational" APIs? Maybe by using a bold type or another visual clue?

azaroth42:
Agree that Image and Prezi should come before Auth and Search. With the upcoming Change Discovery and Content State (both before Image), we could split into two sets: Prezi and Image, and the other 4?

Otherwise, I would suggest:

* Image
* Prezi
* Auth
* Change Discovery
* Content Search
* Content State

Perhaps with one sentence summaries about what they do?

mikeapp:
👍 to the suggestion of having Image and Presentation set apart if we're not going to go with alphabetical order.

kirschbombe:
I'm also in favor of setting image and presi apart

jpstroop:
>Perhaps with one sentence summaries about what they do?

👍🏻

azaroth42:
Editors discussion:

We think it would be good to have two tables, one for current specs, and then the draft specs in progress:

Current table:
  Name / Version (Date) / Description

* Image
* Presentation
* Auth
* Content Search

Draft table:
  Name / Version / Description and link to the responsible TSG 

* Change Discovery
* Content State

Older Versions table:

Ordered as above (Image, Presentation, Auth, Content Search)

azaroth42:
Descriptions (from Rob):

* Image: Access and manipulate image pixels and technical metadata
* Presentation: Just enough description to power a viewer that presents images to users
* Auth: A workflow for allowing users to authenticate with systems to get access to images in different repositories
* Content Search: Search within the annotated content of a Presentation API item, such as a Manifest

azaroth42:
Closing as no longer relevant ... fixed with recent changes.

PR #1984, opened 2021-03-19
Normative "must" formatted as such

mcwhitaker:
On section 3.4 of the Search API spec, a normative "must" was not formatted as a keyword.

https://preview.iiif.io/api/type-must-be-searchHit/api/search/1.0/#search-api-specific-responses


azaroth42:
Agree with the change, but we can't change the published spec. We could issue a 1.0.1 ... but given there's a Search TSG, it should be discussed in that context.

jpstroop:
Doh! I didn't notice that it was against master.

mcwhitaker:
Closing PR in favor of issue #1989

Issue #90, opened 2014-05-12
Feature URIs should be hash URIs that point into the spec


jpstroop:
E.g. size_by_forced_wh should be http://iiif.io/api/image/{version}#size_by_forced_wh (or whatever)

PR #1986, opened 2021-03-22
Content State 0.9 Candidate

tomcrane:
This mostly reproduces the HTTP request/response section from the Presentation API, with some tweaks to make it suitable for content state.

This PR also makes it clear that:

 - content states as HTTP response bodies are JSON-LD, Presentation API responses
 - content states by other mechanisms are base64url-encoded strings

This is I think a stronger assertion than previously.


source/api/content-state/0.3/index.md
```diff
@@ -657,6 +660,38 @@ Firstly, in non-valid, unencoded form to show the annotation:
...
+{: .urltemplate}
+
+If the `Content-Type` header `application/ld+json` cannot be generated due to server configuration details, then the `Content-Type` header _SHOULD_ instead be `application/json` (regular JSON), without a `profile` parameter.
+
+``` none
+Content-Type: application/json
+```
+{: .urltemplate}
+
+The HTTP server _MUST_ follow the [CORS requirements][org-w3c-cors] to enable browser-based clients to retrieve the descriptions. If the server receives a request with one of the content types above in the Accept header, it _SHOULD_ respond with that content type following the rules of [content negotiation][org-rfc-7231-conneg]. Recipes for enabling CORS and conditional Content-Type headers are provided in the [Apache HTTP Server Implementation Notes][notes-apache].
```
zimeon, line 692: This seems to repeat the comment about conneg above?

source/api/content-state/0.3/index.md
```diff
@@ -226,6 +226,9 @@ The destination character set when encoding _MUST_ be UTF-8. For example, when p
 
 When published as inline, base64url-encoded JSON-LD in the full form given in 2.2, the content state Annotation _MAY_ omit the `id` and `@context` properties.
 
+When published on a server for clients to fetch over HTTP, content states _MUST_ be valid JSON-LD documents conforming to the [IIIF Presentation API][prezi-api] and served as described in [Section 5][contentstate-http] below, and _MUST NOT_ be encoded as base64url or any other encoding.
```
zimeon, line 229: The "MOST NOT be encoded" seems strong. Would it be OK to provide a response with `Content-Encoding: gzip` ? I would remove the "or any other encoding" to admit the possibility of existing mechanisms (which don't include b64).

source/api/content-state/0.3/index.md
mcwhitaker, line 692: It does seem to repeat it. If it stays, the "Accept" needs to be back ticked.

azaroth42:
We should close https://github.com/IIIF/api/pull/1978 in favor of this?

tomcrane:
erk... I should have checked first! I could update mine to be a 0.4 too. There is a little extra in it.

source/api/content-state/0.3/index.md
tomcrane: addressed in imminent PR

tomcrane:
Have addresses @zimeon comments and as discussed on Discovery call today, bumped the version to 0.9 in anticipation of TRC process.

tomcrane:
Ignore source/api/content-state/0.3/index.md, the changes are now in source/api/content-state/0.9/index.md

tomcrane:
Preview: https://preview.iiif.io/api/86a-content-type-for-content-state/api/content-state/0.9/

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+
+### 2.2 Form of Annotation
+
+Any Annotation may have one or more [motivations][org-w3c-webanno-motivation], that provide the reason(s) why is was created. For example, `bookmarking`. 
+
+A content state Annotation always has the motivation `contentState`. This motivation is not defined by either the [W3C Web Annotation Data Model][org-w3c-webanno] or the IIIF Presentation API, and is chosen to avoid potential ambiguity when the target of the content state Annotation is itself an Annotation. The content state annotation may also have additional motivations such as `bookmarking`, `identifying` and so on, as defined by the W3C Web Annotation Model, but it is its particular `contentState` motivation that produces the required behavior in compatible software.
+
+A content state annotation can take several formats. All of the four examples below are equivalent to the following annotation, which is the simplest content state: a reference to a IIIF Manifest.
```
azaroth42, line 122: Suggest delete "All of the four examples..." as there aren't four examples below, and the examples (which might have existed in previous versions) span across subsections. There are more than 4 subsections in 2.2, which adds to the confusion.

Possible replacement: A content state annotation can take several formats, described in the following sections.

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+```
+
+Publishers _SHOULD_ provide the content state annotation in one of the following forms.
+A client _SHOULD_ be able to accept and process the content state in all of these forms. 
+
+#### 2.2.1 Full Annotation
+
+The content state _MAY_ be supplied as JSON-LD, as a full Annotation with the motivation `contentState`, as in the example above.
+
+The target of the annotation is in this case a complete IIIF resource (here, a Manifest) but in more complex cases, the target could be a part of a IIIF resource.
```
azaroth42, line 144: Suggest adding parenthetical commas:  ... annotation is, in this case, a ...

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+}
+```
+
+This form is better suited to scenarios where compactness is important, for example a query string parameter, where it is will be combined with base64url encoding as described in Section 2.3.
+
+#### 2.2.4 Target URI
+
+ The content state _MAY_ be supplied as a string whose value is the `id` (the dereferenceable URI) of the `target` property only. This is the simplest form and is just the URI of a resource. For the example above, this would be the URI `https://example.org/iiif/item1/manifest`. The client would simply load this Manifest and display it. 
+
+Examples 2.2.2 and 2.2.4 are both URIs. It is up to the client to recognise that 2.2.4 is a Manifest, whereas 2.2.2 is an Annotation that points to a Manifest. The client inspects the `type` property to determine what the de-referenced resource is.
```
azaroth42, line 167: Suggest adding that if the type is Annotation, then the client must also look at `motivation` to determine if it is a contentState or not.  I think this should also have a normative MUST, otherwise it won't work.

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+    "id": "https://example.org/object1/manifest",
+    "type": "Manifest"
+  }]
+}
+```
+
+
+### 2.3 Protocol and encoding requirements
+
+There are many ways in which the content state data shown in Section 2.2 could be passed to, or exported from, IIIF compatible software such as a viewer. For interoperability, agreeing on the model is not enough. This section defines _Protocols_ for the transfer of this data, so that implementing software can receive or pass a content state without specific knowledge of other participating software. These protocols make use of widely supported features of modern browsers:
```
azaroth42, line 207: Suggest:  ... could be exchanged between IIIF-compatible systems.

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+
+Examples of these protocols are given in Section 3, [Initialization mechanisms][contentstate-initialization].
+
+
+
+#### 2.3.1 Base 64 Encoding with URL and Filename Safe Alphabet
+
+Any content state that is in JSON-LD form, rather than a simple URI string, _MUST_ be encoded as [Base 64 Encoding with URL and Filename Safe Alphabet][org-rfc-4648-5] ("base64url") when declared inline in an HTML document or passed as an HTTP request parameter (e.g., GET or POST), and a client _MUST_ accept it in that form. Note that "base64url" is not the same encoding as "base64", and is used to ensure the integrity of content states passed between web applications.
+
+The destination character set when encoding _MUST_ be UTF-8. For example, when passing the content state as JSON-LD in a query string parameter, base64url is used, as unencoded JSON is vulnerable to corruption. Simple URI forms _SHOULD_ be plain strings.
```
azaroth42, line 225: Do we need to be explicit about what to do with IRIs? e.g. https://en.wiktionary.org/wiki/Ῥόδος
(e.g. encode directly, or transform into %encoded URI form first)?

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+  "id": "https://example.org/object1/canvas7#xywh=1000,2000,1000,2000",
+  "type": "Canvas",
+  "partOf": [{
+    "id": "https://example.org/object1/manifest",
+    "type": "Manifest"
+  }]
+}
+```
+
+The above annotation JSON can be condensed to remove unecessary whitespace:
```
azaroth42, line 245: Typo:  unnecessary

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+```
+
+The string "eyJp...XX0=" is the now URI-safe, encoded form of the content state, suitable for passing to and from web applications.
+
+Not all use cases for providing a content state to a client require base64url encoding, because not all of them are susceptible to HTTP transmission issues. For example, a user interface could allow the user to paste or even type content state JSON-LD directly. Examples are given in section 3. In these scenarios, the client should accept unencoded JSON, too.
+
+
+#### 2.3.4 URI-Encoding
+
+If the content state is a simple URI, it _MUST NOT_ be base64url encoded. It _MAY_ be uri-encoded (percent encoding, as defined by [org-rfc-3986][Generic URI Syntax, RFC 3986]), and a client _MUST_ accept it in that form.
```
azaroth42, line 268: This reference doesn't render in the preview -- is it in _links ?

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+
+Values are passed into web applications via request parameters, or attributes on HTML elements.
+
+If a client is capable of reading the content state from the value of an HTTP GET or POST request parameter, it _MUST_ look for the content state in a request parameter called `iiif-content`.
+
+If a client is capable of reading the content state from the value of an attribute on an HTML element that instantiates the client, it _SHOULD_ look for the content state in an attribute called `data-iiif-content`.
+
+A content state Annotation may be published inline as part of an HTML document, as in the preceding `data-iiif-content` example, or as a query string parameter forming part of the `href` property of an HTML anchor tag. 
+
+If the content state is a simple URI, the client _MUST_ load the resource at that URI and process it. The resource at that URI _MUST_ be the full Annotation as in 2.2.2. above, or a IIIF Resource as in 2.2.4. That is, the de-referenced response _MUST_ be JSON-LD, and _SHOULD_ have a value of `type` taken from `Annotation`, `Collection`, `Manifest`, `Canvas` and `Range`. The response _MUST_ use UTF-8 encoding.
```
azaroth42, line 281: General comment:  Should references to examples by #links within the document?

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+        "type": "Manifest"
+      }
+    ]
+  }
+}
+```
+
+An example of this usage would be a link from search results to a particular page of a digitized book, or a stored bookmark of a particular page (i.e., Canvas).
+
+Without encoding, the link to the viewer would look like this:
```
azaroth42, line 336: Suggest: Make it clearer that this is invalid and NOT to be copied

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+
+The content state _MAY_ be supplied as JSON-LD, as a full Annotation with the motivation `contentState`, as in the example above.
+
+The target of the annotation is in this case a complete IIIF resource (here, a Manifest) but in more complex cases, the target could be a part of a IIIF resource.  
+
+#### 2.2.2 Annotation URI
+
+The content state _MAY_ be supplied as a string whose value is the URI of an Annotation with the motivation `contentState`, that the client must de-reference and process. For the example above, this would be the URI `https://example.org/Annotation-server/bookmarks/b1`. The response from that URI would be the JSON above.
+
+#### 2.2.3 Target body
```
azaroth42, line 150: Title case:  Target Body

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+
+This form is better suited to scenarios where compactness is important, for example a query string parameter, where it is will be combined with base64url encoding as described in Section 2.3.
+
+#### 2.2.4 Target URI
+
+ The content state _MAY_ be supplied as a string whose value is the `id` (the dereferenceable URI) of the `target` property only. This is the simplest form and is just the URI of a resource. For the example above, this would be the URI `https://example.org/iiif/item1/manifest`. The client would simply load this Manifest and display it. 
+
+Examples 2.2.2 and 2.2.4 are both URIs. It is up to the client to recognise that 2.2.4 is a Manifest, whereas 2.2.2 is an Annotation that points to a Manifest. The client inspects the `type` property to determine what the de-referenced resource is.
+
+#### 2.2.5 Limitations of simple URIs
```
azaroth42, line 169: Title Case

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+  "type": "Canvas",
+  "partOf": [{
+    "id": "https://example.org/object1/manifest",
+    "type": "Manifest"
+  }]
+}
+```
+
+
+### 2.3 Protocol and encoding requirements
```
azaroth42, line 205: From here on, headings seem to be in sentence case rather than title case (which is the norm from other specs), should be fixed.

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+    },
+    "selector": {
+        "type": "PointSelector",
+        "t": 14.5
+	  }
+  }
+}
+```
+
+This example should cause a viewer to open Manifest https://example.org/iiif/id1/manifest, navigate to Canvas https://example.org/iiif/id1/canvas1, and start playing at 14.5 seconds into that canvas.
```
azaroth42, line 593: backticks around URLs ?

source/api/content-state/0.9/index.md
```diff
@@ -0,0 +1,713 @@
...
+
+
+## Appendices
+
+### A. Acknowledgements
+{: #acknowledgements}
+
+Many thanks to the members of the [IIIF community][iiif-community] for their continuous engagement, innovative ideas, and feedback.
+
+Many of the changes in this version are due to the work of the [IIIF Discovery Technical Specification Group][groups-discovery], chaired by Antoine Isaac (Europeana), Matthew McGrattan (Digirati) and Rob Sanderson (J. Paul Getty Trust). The IIIF Community thanks them for their leadership, and the members of the group for their tireless work.
```
azaroth42, line 702: Rob Sanderson (Yale University) please :)

azaroth42 (changes requested):

source/api/content-state/0.9/index.md
tomcrane: I have moved the full annotation example into 2.2.1, I think this improves things.

source/api/content-state/0.9/index.md
tomcrane: added

source/api/content-state/0.9/index.md
tomcrane: In this scenario, where you have found a resource that you expect to be a content state (because it's the value of a `iiif-content` attribute or whatever), if the resource is an annotation and the `motivation` is **not** contentState, then the content being addressed is the annotation itself, rather than the target of the annotation.

This is valid (the target is implied, and can be any IIIF resource), but probably very rare. "I want you to look at this tagging annotation". I've now tried to explain it here.

source/api/content-state/0.9/index.md
tomcrane: done

source/api/content-state/0.9/index.md
tomcrane: yes we do...

source/api/content-state/0.9/index.md
tomcrane: oops

source/api/content-state/0.9/index.md
tomcrane: I had the [label] and [ref] the wrong way round

source/api/content-state/0.9/index.md
tomcrane: Yes probably - Presentation 3 does this. https://github.com/IIIF/api/blob/master/source/_includes/links.md has a load of links but they would have to go in now as 0.9 links and then be duplicated as 1.0 links. Maybe one to discuss.

source/api/content-state/0.9/index.md
tomcrane: fixed

source/api/content-state/0.9/index.md
tomcrane: fixed

source/api/content-state/0.9/index.md
tomcrane: fixed

source/api/content-state/0.9/index.md
tomcrane: added

azaroth42 (dismissed):
Modulo #90, which needs discussion, agree this is ready to go.

source/api/content-state/0.9/index.md
```diff
@@ -158,7 +158,7 @@ The content state _MAY_ be supplied as JSON-LD, as the value of the `target` pro
 }
 ```
 
-This form is better suited to scenarios where compactness is important, for example a query string parameter, where it is will be combined with base64url encoding as described in Section 2.3.
+This form is better suited to scenarios where compactness is important, for example a query string parameter, where it is will be combined with the encoding described in Section 2.3.
```
mcwhitaker, line 161: Should it say "it is combined" or "will be combined"?

tomcrane: good catch, have pushed a fix

azaroth42 (dismissed):

zimeon (approved):

mcwhitaker (approved):

mikeapp (dismissed):

kirschbombe (approved):

zimeon (dismissed):

azaroth42 (dismissed):

mcwhitaker (dismissed):

kirschbombe (dismissed):

azaroth42 (approved):

zimeon (approved):

jpstroop (approved):

mixterj (approved):

mikeapp (approved):

PR #2005, opened 2021-06-10
Search v3 alignment

azaroth42:
Force a preview build

### DIFF
--- a/source/api/search/2.0/index.md
+++ b/source/api/search/2.0/index.md
@@ -1,0 +1,779 @@
+---
+title: "IIIF Content Search API 2.0"
+title_override: "IIIF Content Search API 2.0"
+id: content-search-api
+layout: spec
+tags: [specifications, content-search-api]
+major: 2
+minor: 0
+patch: 0
+pre: alpha
+cssversion: 2
+redirect_from:
+  - /api/search/2/index.html
+---
+
+## Status of this Document
+{:.no_toc}
+__This Version:__ {{ page.major }}.{{ page.minor }}.{{ page.patch }}{% if page.pre != 'final' %}-{{ page.pre }}{% endif %}
+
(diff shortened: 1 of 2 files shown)


glenrobson:
This is the preview link: https://preview.iiif.io/api/search-v3-alignment/api/search/2.0/

kirschbombe (dismissed):

mikeapp:
So we don't forget: the motivations registry page will need to be updated with the release date along with the main spec.

kirschbombe (approved):

kirschbombe (dismissed):

mikeapp (approved):
Approving but noticed a minor glitch: in preview the "Latest Stable Version" is showing as `..` however the link works.

mixterj (approved):
noticed the same issue with the latest stable version text in the Preview but as mentioned, the link works

digitaldogsbody:
> Approving but noticed a minor glitch: in preview the "Latest Stable Version" is showing as `..` however the link works.

> noticed the same issue with the latest stable version text in the Preview but as mentioned, the link works

Not sure why this is happening. I updated the stable version numbers which are used to generate that link in https://github.com/IIIF/api/pull/2181/files so it should get pulled through ok.

I noticed that the index page (https://preview.iiif.io/api/search-v3-alignment/) also didn't pick up the change, and is still showing v1. I also updated the main API data file which is used to generated that in the same PR.

Any ideas @glenrobson?

zimeon:
I will :+1: anyway but ideally I think we should tighten the definition of the `date` parameter https://github.com/IIIF/api/pull/2182 (and fix the link already mentioned). Otherwise, looks good!

zimeon (approved):

glenrobson (approved):

digitaldogsbody:
The link issue didn't fix with the deploy to live after merge either

PR #2032, opened 2021-07-27
navPlace Extension Changes

thehabes:
Updates in response to the TRC review and IIIF Maps TSG co-chairs' ongoing conversations.


source/api/extension/navplace/index.md
```diff
@@ -21,14 +21,14 @@ Changes will be tracked within the document.
...
 
-This extension describes a property called `navPlace` which contains geographic coordinates in the form of [GeoJSON-LD](https://geojson.org/geojson-ld/). Clients may use this property to leverage the navigational functionality of web maps such as Google Earth, Leaflet, OpenLayers, etc. giving them the opportunity to enrich data presentation through common web map platforms.
+This extension describes a property called `navPlace` which contains earthbound geographic coordinates in the form of [GeoJSON-LD](https://geojson.org/geojson-ld/). Clients may use this property to leverage the navigational functionality of web maps such as Google Earth, Leaflet, OpenLayers, etc. giving them the opportunity to enrich data presentation through common web map platforms.
```
thehabes, line 31: How in the heck do we justify 'earthbound' in one sentence.

source/api/extension/navplace/index.md
```diff
@@ -261,7 +260,7 @@ Here you can see an example of a IIIF Manifest with the `navPlace` property. It
             "properties":{
                "label":{
                   "en":[
-                     "Photograph from Göttingen, Germany"
+                     "navPlace Extension Specification Example Manifest"
```
thehabes, line 263: If we aren't dodging properties, then our full example should have one.  I am repeating the label property of the Manifest here now (as opposed to the Canvas, b/c navPlace is on the Manifest not the Canvas).  This is b/c of how we worded 2.2 and 3.3 with "MAY be repeated".

source/api/extension/navplace/index.md
```diff
@@ -281,7 +280,9 @@ Here you can see an example of a IIIF Manifest with the `navPlace` property. It
...
+The GeoJSON `properties` object is generic and can be nearly anything. It is used to pass metadata along with the geographic coordinates. Web maps offer common ways for implementers to generate UI(s) for displaying values in `properties` along with the shape made by the geographic coordinates. IIIF encourages interoperability, which is made possible in large part due to Linked Data. Any metadata used in `properties` _SHOULD_ be described by a linked data context. Desired metadata already noted by a property on a IIIF resource _SHOULD_ be gathered by consuming that property for the web map UI and _MAY_ be repeated in GeoJSON `properties` to leverage automated functionality.
```
thehabes, line 283: 2.2 and 3.3 have some similarity.  Is the repetition OK?

source/api/extension/navplace/index.md
thehabes: Technically b/c GeoJSON and WGS84
Philosophically because...TODO

thehabes: Mention how ambiguous it is to use coordinates in space (we don't know those other celestial spheres as well as Earth).  We know how to measure Earth b/c we are on it and worked on it for centuries.  We just don't have this for space (star maps).  Frame it as a matter of 'universally' agreed upon standards.

thehabes: https://lunar.gsfc.nasa.gov/library/LunCoordWhitePaper-10-08.pdf...but there is no GeoJSON built for this (A GeoJSON that accepts other coordinate systems.)

source/api/extension/navplace/index.md
```diff
@@ -139,6 +145,8 @@ The `navPlace` property identifies a single or multiple geographic areas pertine
 *   The `navPlace` property _MUST NOT_ be used on other IIIF resource types.
```
glenrobson, line 145: Do we do this in other specs? I wonder if we just say where it can be and don't mention when we can't. Does this conflict with "Define Success, Not Failure" from https://iiif.io/api/annex/notes/design_patterns/#define-success-not-failure

thehabes: Yes, we do this in other specs.  https://iiif.io/api/presentation/3.0/#navdate

glenrobson: OK fair enough :-)

"Other types of resource must not have the navDate property."

PR #2082, opened 2021-12-01
Georeference Extension

thehabes:
This PR should always be in draft until it is ready for review by TRC.  It exists to review the georeferencing work and trigger the preview build.


thehabes:
Comment from @azaroth42 (paraphrased)

> You have purpose on something other than a TextualBody which is not expected.  In this case, it does become an error, as the FeatureCollection doesn't have a value and TextualBody must have a value

mikeapp (dismissed):

azaroth42:
What are all the examples/ files for? They look like they should be in cookbook entries?

kirschbombe (dismissed):

mikeapp (dismissed):

zimeon (approved):
I approve but note to following minor editorial issues that could perhaps be fixed when merging:
* The term “linked data 1.1” is used several times — I think this is intended to mean “JSON-LD 1.1” which is then specific and can be linked
* typo s/extentd/extend/
* section 2.2 - there is a funny “i in circle” symbol in first bullet
* there is a “georeferencing” in double quotes in the running text that should instead be backticked
* a change log needs to be added (and updated with pub date)

azaroth42 (changes requested):
In the context, there are references to web page sections that should be to a class and predicates. In particular line 44 on:

```
"ImageAPISelector": "https://iiif.io/api/annex/openannotation/#iiif-image-api-selector",
    "region": "https://iiif.io/api/image/3.0/#41-region",
    "size": "https://iiif.io/api/image/3.0/#42-size",
```
These should be the ontology URIs, not specification links.

I don't believe that the contexts at line 31-34 are necessary as they're at the top level, and the URI is oa:body.

Why the need for the georef-terms file? It seems short enough to go into the extension directly?

thehabes:
> ```
> "ImageAPISelector": "https://iiif.io/api/annex/openannotation/#iiif-image-api-selector",
> "region": "https://iiif.io/api/image/3.0/#41-region",
> "size": "https://iiif.io/api/image/3.0/#42-size",
> ```
> These should be the ontology URIs, not specification links.

Yes, good catch that is an oversight.  We should make that change.

> I don't believe that the contexts at line 31-34 are necessary as they're at the top level, and the URI is oa:body.

Here we are attempting a Scoped Context so that when GeoJSON terms are used in `Annotation.body` or `Annotation.resource` they have the GeoJSON context for definition.  GeoJSON terms will only appear in `body` (Presi 3) or `resource` (Presi 2).  I think we should have "https://geojson.org/geojson-ld/geojson-context.jsonld" there to avoid context clashes by accident in any other term (see navPlace's strategy).  The "http://www.w3.org/ns/anno.jsonld" is already defined at the top level, and so is actually duplicated here which is not necessary.  This could be a little shorter by changing to

```
"body": {
      "@context": "https://geojson.org/geojson-ld/geojson-context.jsonld",
      "@id": "oa:body"
},
"resource": {
      "@context":"https://geojson.org/geojson-ld/geojson-context.jsonld",
      "@id": "oa:hasBody"
}
```

> Why the need for the georef-terms file? It seems short enough to go into the extension directly?
We were on the edge about that and went with what we sensed was the general opinion from the review sessions.  We separated them out in favor of a shorter document.  We imagine a future where more terms are added.  However, it would be appropriate for them to be inline instead, and there is a precedent for doing it that way from other IIIF API documents.  Whichever method is favored we can do.

thehabes:
As a second comment to this, maybe if I tell you the LD functionality I am looking for this will make more sense.

Using Presentation API 3 as an example

From the context
```
"iiif_prezi": "http://iiif.io/api/presentation/3#",
 "as": "http://www.w3.org/ns/activitystreams#",
 ...
 ```
 then
 ```
 "AnnotationCollection": {
       "@id": "as:OrderedCollection",
       ...
},
"Manifest": {
     "@id": "iiif_prezi:Manifest",
     ...
 }
 ```
https://www.w3.org/ns/activitystreams#OrderedCollection takes me somewhere direct.  http://iiif.io/api/presentation/3#Manifest and even https://iiif.io/api/presentation/3.0/#Manifest do not.

I want `size`, `region` and `ImageApiSelector` to have valid links to direct vocabulary for those terms.  Perhaps this extension is not the place to try and make that happen and jump ahead of what should be improvements upstream.

**Comment by kirschbombe** [comment]
@thehabes -  the ImageAPISelector context URL is https://iiif.io/api/annex/openannotation/context.json. This is for v2, but there is a branch for v3 that I need to finish up (the URL will be https://iiif.io/api/registry/webannotation/context.json).

**Comment by mikeapp** [comment]
The context has been simplified.  `ImageApiSelector` is defined in the context file pending the merge of #2231.

**Comment by thehabes** [comment]
Thanks @kirschbombe and @mikeapp. 

> The ImageAPISelector context URL is https://iiif.io/api/annex/openannotation/context.json. This is for v2, but there is a branch for v3 that I need to finish up (the URL will likely be https://iiif.io/api/registry/webannotation/context.json)

Ok, looks good!  I even did a test and updated the context I am testing with to be
```JSON
"ImageApiSelector":{
        "@id":"iiif_selector:ImageApiSelector",
        "@context": "https://iiif.io/api/annex/openannotation/context.json"
}
```
This works expected, and I no longer have to embed the terms directly.  It is processes as 
```JSON
"http://www.w3.org/ns/oa#hasSelector": {
      "@type": "http://iiif.io/api/registry/selectors#ImageApiSelector",
      "http://iiif.io/api/image/2#regionValue": {
        "@type": "http://www.w3.org/2001/XMLSchema#string",
        "@value": "810,900,260,370"
      },
      "http://iiif.io/api/image/2#sizeValue": {
        "@type": "http://www.w3.org/2001/XMLSchema#string",
        "@value": "5965,2514"
      }
}
```
I will test the version 3 one as well when it is available.  Good job, thank you.  We can simplify `ImageApiSelector` even further like shown above when https://iiif.io/api/registry/webannotation/context.json is ready.

zimeon (approved):

kirschbombe (approved):

glenrobson (approved):

mikeapp (approved):

azaroth42 (approved):

mixterj (approved):

tomcrane (dismissed):

mikeapp (approved):

zimeon (approved):
IMO, there is no need for another round of +1's for this date change

Issue #2019, opened 2021-07-06
Update API Annex doc references to GeoJSON inclusion, given IIIF NavPlace Extension? 


hadro:
With the potential approval of the NavPlace extension, does the section in the Annex doc on [Linking to External Services](https://iiif.io/api/annex/services/#geojson) with a section talking about geojson-ld need to be revised or updated?

azaroth42:
Yes :+1: I would say deleted, and replaced with a reference to the `navPlace` extension, when it is published.

kirschbombe:
Now that `navPlace` is published, the Annex doc should be updated. Should this go to @thehabes or @mikeapp? I can also take a stab at it.

thehabes:
I don't mind doing it!  I will look at it tomorrow or the 13th.  I can't make the call this Friday 1/14, I will be out 1/14-1/18 but I can have something together for you to look at.

Just to confirm, I am 99% sure we are talking about main/source/annex/services/index.md, Section 3 (Services) and more specifically Section 3.2 (GeoJSON).  

I have a IIIF/API repo fork, I will put a PR together.  I'll post back here when it's up and I will let @kirschbombe and the Cookbook know.

I agree with @azaroth42.  Do we change the icon to ![not allowed] under presentation API, and just have section 3.2 link out to [navPlace ](https://iiif.io/api/extension/navplace/) as the solution for using GeoJSON?  Or are we thinking about a complete removal and actually deleting it from the Section 3 table and deleting 3.2, expecting that people will be routed to [navPlace](https://iiif.io/api/extension/navplace/) when searching for "GeoJSON with IIIF".

kirschbombe:
All good questions @thehabes! Is GeoJSON still the service? It's just used in the API via navPlace? If that's the case, I wonder if we should keep "GeoJSON" in the table and replace 3.2 description with info about using navPlace to use the GeoJSON service.

thehabes:
[...]

PR #2089, opened 2022-01-12
Annex Update for GeoJSON

thehabes:
See https://github.com/IIIF/api/issues/2019.

Now that `navPlace` is released, people should use this property to supply GeoJSON instead of using service objects.  There are Linked Data complications with the service object approach, and navPlace provides a IIIF-safe environment for GeoJSON values that handles these complications.  Since it can be used with any IIIF Resource Type and can be embedded (and therefore still link to a service), it eliminates the need for the service.  Though still possible, it is very much not recommended.  This PR proposes to make the Service Table in Section 3 say that GeoJSON is not allowed, and section 3.2 now explains why and links out to navPlace.

There are other options, like just removing GeoJSON from the table and deleting 3.2, since people will most likely be routed to the navPlace extension when searching "IIIF and GeoJSON".  This PR opens the discussion by starting with the least destructive approach.  It will remain in draft until we come to a consensus.


tomcrane:
Editors call - agree this has been superseded by #2092 - @thehabes OK to close this Draft?

thehabes:
@tomcrane yes, close the draft.  It served its purpose.

PR #2092, opened 2022-01-28
Migrate geojson and physdim services from annex to registry

kirschbombe:
migrate geojson and physdim services from annex to registry


tomcrane:
Icons in the table could maybe reference https://iiif.io/api/presentation/3.0/#a-summary-of-property-requirements rather than redefine the symbols?

glenrobson:
How do you get to the webannotation registry part? Should it be linked from the api/index.md file?

Issue #2052, opened 2021-10-12
find new term for `label` in the Search API Auto Complete Response


mixterj:
The Search API Auto Complete (https://preview.iiif.io/api/search-v3-alignment/api/search/2.0/#autocomplete) Response data (https://preview.iiif.io/api/search-v3-alignment/api/search/2.0/#response) defines the use of `label` as a string which conflicts with its use as a language map. In this situation it is more than likely just a string. To avoid conflict, we need to define a new term for `label` in this context.

mbennett-uoe:
TSG call: Can we update this to use a language map with `@none` when the language is unknown?

digitaldogsbody:
Related to discussion in #2054 about redefining terms used

azaroth42:
Discussion about label of resource being separate for alternate form of the match value.

azaroth42:
Related language issue https://github.com/IIIF/api/issues/513

digitaldogsbody:
Remaining changes:

- Update example to convert string to array for none.
- Update sentence about label to indicate usage of language map pattern (Prezi3 4.4)

PR #2125, opened 2022-04-19
update json-ld context for search

### LINKED ISSUE #2055: Have Search 2 context import Prezi3 context
In the Search alignment work there are many instances where the Search API only uses/imports the Prezi3 context but there are a few situations where it needs both the Prezi3 and Search context. To help standardize the approach to adding the correct context the Search Context could just import the Prezi3 context.

azaroth42:
Fixes #2055, #2052

### DIFF
--- a/source/search/2.0/index.md
+++ b/source/search/2.0/index.md
@@ -798,8 +798,7 @@ The objects in the list of terms are all of `type` "Term", and this _MAY_ be inc
   * The matching term is given as the value of the `match` property, and _MUST_ be present.
   * The link to the search to perform is the value of the `url` property, and this _MUST_ be present.
   * The number of matches for the term is the integer value of the `count` property, and _SHOULD_ be present.
-  * A label to display instead of the match can be given as the value of the `label` property, and _MAY_ be present.  There may be more than one label given, to allow for internationalization.
-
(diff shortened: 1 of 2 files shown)


mikeapp (approved):

digitaldogsbody:
Do we also need to change [Section 3.4](https://preview.iiif.io/api/search-v3-alignment/search/2.0/#search-api-specific-responses)?

Since the Search context will now include the Prezi3 context, is the following sentence still accurate/correct?
> As these responses include Search specific information, the value of @context must be an array with both the Presentation API and the Search API context URIs included, in that order. This allows the two APIs to develop separately and yet remain as synchronized as possible.

If this change is made, then I think the examples following in the subsections would also need to be updated to remove the Prezi3 context.

digitaldogsbody:
Also, I've noticed that all of the Prezi context links in the examples (which would still be needed for the examples in 3.3.X even if removed from 3.4.X) are rendering incorrectly, missing the version number (and thus gaining a double slash and leading to a 404): `http://iiif.io/api/presentation//context.json`. I think this is because the `{{ site.presentation_api.stable.major }}` variable is not getting substituted by the compiler?

kirschbombe (approved):

digitaldogsbody (approved):

PR #2169, opened 2022-11-15
Adding the Europeana Data Model to the Registry of Profiles

nfreire:
The Discovery Technical Group  intends to write a note about crawling/indexing the metadata on IIIF resources using the IIIF Change Discovery API, and also change the recipe for Linking to Structured Metadata via seeAlso in manifests.
With these objectives in mind, we'd like to add the Europeana Data Model to the Registry of Profiles.


glenrobson:
Note this pull request has been deleted but is now replaced by this one: https://github.com/IIIF/api/pull/2235

PR #2182, opened 2022-12-06
Be more specific about date ranges

zimeon:
Tighten definition of date ranges


zimeon:
I did wonder whether we should more closely parallel the Prezi 3 description of `navDate` https://iiif.io/api/presentation/3.0/#navdate by introducing reference to XSD dateTime literals, but that would be a bigger edit

digitaldogsbody (approved):

mikeapp (approved):

mixterj (approved):

PR #2183, opened 2022-12-06
Setting latest search version to v2

glenrobson:
Updating theme for the `/api/` page and using the correct variables in the search page.


digitaldogsbody (approved):
Looks fixed in the preview deployment build

mikeapp (approved):

PR #2187, opened 2022-12-20
Fixing redirects

glenrobson:
According to the [jekyll-redirect-from](https://github.com/jekyll/jekyll-redirect-from) you can specify redirects as:

```
redirect_from:
  - /post/123456789/my-amazing-post/
```

or 

```
redirect_from: /post/123456798/
```

but not:

```
redirect_from:
   /post/123456789/my-amazing-post/
```

So this fixes all of the redirects I could find. This causes problems on Windows as it tries to create directories with spaces in the filename and fails. On Mac it seems to just not create many of the redirect files.

### DIFF
--- a/source/annex/notes/design_principles.md
+++ b/source/annex/notes/design_principles.md
@@ -4,7 +4,7 @@ layout: spec
 tags: [annex, presentation-api, image-api]
 cssversion: 3
 redirect_from:
-  /annex/notes/design_patterns/
+ - /annex/notes/design_patterns/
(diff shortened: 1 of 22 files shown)


zimeon (approved):
Thanks Glen!

mikeapp (approved):

kirschbombe (approved):

mixterj (approved):

kirschbombe (approved):

PR #2220, opened 2023-03-02
Auth2 seq diagram

mikeapp:
Authorization Flow section 2.2:
* adds diagram of service interactions
* edit notes about interactions and group them into bulleted list

The language is 2.2 was very rough and the opening two sentences need further improvement.


mikeapp:
@tomcrane I assumed the `1` in your diagram label was the number of the diagram, so I followed that pattern and bumped yours to `2`

zimeon (approved):

zimeon:
I'm not sure the numbers in the diagram labels are very useful. I'd be inclined to remove them both

tomcrane:
That number is in the HTML not the diagram - 
https://github.com/IIIF/api/blob/092ff796c5cb669dd81ea2353b2a4ce67cc26a5c/source/auth/2.0/index.md?plain=1#L906

But yes, I don't see a need for the number.

tomcrane:
In fact that HTML table is a hangover from the previous version, there may be a better way to include images in the new IIIF site design.

mixterj (approved):

kirschbombe (approved):

tomcrane (approved):

PR #2221, opened 2023-03-07
Proposal for adding 'timed-text' in the Text Granularity Extension

nfreire:
Captions and subtitles for video objects can be made available via the IIIF Presentation API by using annotations on the canvas that contains the media file. For reference, see the Section "Captions and Subtitles" for the recipe "[Transcripts, Captions, and Subtitles - General Considerations](https://iiif.io/api/cookbook/recipe/0231-transcript-meta-recipe/)".

Just like the OCRed text of a newspaper can be provided via annotations with spatial coordinates on a images, captions and subtitles may be provided as annotations with temporal coordinates (a cookbook [recipe for providing captions and subtitles as annotations](https://github.com/IIIF/cookbook-recipes/blob/0079-webvtt-to-annotations/recipe/0079-webvtt-to-annotations/index.md) is planned for the IIIF cookbook). 

The Text Granularity Extension allows one to indicate the level of text granularity for an annotation (block, line, etc.), but currently it does not contain a suitable granularity value for the case of captions and subtitles, whose text granularity is neither paragraphs of sentences. The text granularity of captions/subtitles follows standard subtitling guidelines in terms of reading speed, number of lines in each subtitle, line length (number of characters), minimum and maximum subtitle duration, and minimum interval between two consecutive subtitles. 

In conclusion, the text granularity of captions and subtitles is specific to these resources, and this proposal consists in adding the text granularity level 'timed-text' to the levels defined by the extension.


azaroth42 (dismissed):
I'm okay with it as it stands, but slightly reluctant as it somehow still seems to be orthogonal to the granularity. I don't have a good argument against it, however.

azaroth42:
Closing -- rejected by TRC

PR #2222, opened 2023-03-07
Updates Based On Review Period Comments

thehabes:


thehabes:
Let's take this over to iiif:georef to get the preview with all these changes.

PR #2227, opened 2023-03-28
Add algorithm for auth 2 flow

tomcrane:
Updates to the spec as discussed on Auth call 28 Feb 2023

 - Introduce required order in which access services _SHOULD_ be tried (`external`, `kiosk`, `active`), before we get to the flow diagram.
 - Add an _algorithm_ - while this is still in review I haven't formatted the markdown into a nested list, so it only makes sense when you view the markdown source. I'll decorate this with nested lists markup (probably, rather than markdown) once we agree it looks OK.
 - Fix a couple of JSON errors
 - Remove some unnecessary repetitive text


zimeon (changes requested):
Content is good, two things:
* There are many instances "Capitalized service" (e..g Access service, Probe service) where in the rest of the document we (mostly) say "uncapitalized service"
* What is the rule for whether there should be a colon at the line end in pseudocode? Not sure I see that pattern -- if indents are key then are they required at all?

mikeapp (approved):

zimeon (approved):
I see your pain with markup, but I think this is good to go

zimeon:
Given the need to get this out for TRC and that we are only merging to a preview branch, I'm going to merge

PR #2231, opened 2023-04-25
add Selectors  section to Registry; move/update imageAPISelector context

kirschbombe:
This PR:
- creates a new folder/section in the registry for "Selectors"
- edits the old openannotation index page to focus on selectors, not web anno extensions
- copies the current imageAPISelector context from annex to registry/selectors
- updates the imageAPISelector context to point to the current version of image api instead of specific version
- change the ImageApiSelector context to include all selectors - so now it's the Selectors context
- adds note to ImageApiSelector documentation that the default values in the table are for Image 3 and that implementers should consult previous versions for their default values
- includes the selectors context in presentation 3 context so it doesn't need to be imported separately
- update link to selectors registry in presi 3 documentation
- add mention of webanno selectors and FragmentSelector to Selectors registry
- updates change log

Preview: https://preview.iiif.io/api/image-api-selector-update/registry/selectors/
Context preview: https://preview.iiif.io/api/image-api-selector-update/registry/selectors/context.json


source/registry/selectors/index.md
```diff
@@ -0,0 +1,160 @@
...
+
+In order to make this as easy as possible for the situations when a IIIF Image API endpoint exists, we introduce a new Selector class called [`ImageApiSelector`][registry-image-api-selector].  It has properties that give the parameter values needed to fill out the URL structure in the request.  If the property is not given, then a default should be used.
+
+One use of this is within the [IIIF Presentation API][prezi-api], when a Canvas is being painted by part of an image, or an image that requires rotation before display.  
+
+| Property | Default   | Description                                            |
+| -------- | --------- | -----------------------------------------------------  |
+| type     |           | Required.  Must be the value "ImageApiSelector".       |
+| region   | "full"    | The string to put in the region parameter of the URI.  |
```
mikeapp, line 61: Would be `max` in v3

source/registry/selectors/index.md
```diff
@@ -0,0 +1,160 @@
...
+| size     | "full"    | The string to put in the size parameter of the URI.    |
```
mikeapp, line 62: Would be `max` in v3

zimeon (changes requested):

source/registry/selectors/index.md
```diff
@@ -0,0 +1,160 @@
...
+### B. Change Log
+
+| Date       | Description                                        |
+| ---------- | -------------------------------------------------- |
+| 2023-04-27 | Version 1.2 (Change document to Registry of Selectors from Open/Web Annotation Extensions)
+| 2018-03-14 | Version 1.1 (Update ImageApi, add Point Selectors) |
+| 2014-07-01 | Version 1.0 (RFC)                                  |
+{:.api-table}
+
+[prezi-rot]: {{ site.api_url | absolute_url }}/presentation/2.0/#rotation "Rotation in Presentation API"
```
zimeon, line 157: I think \[prezi-rot\] should be in `links.md` and we should also reference prezi 3

azaroth42:
Context should use scoped contexts to limit the definition of the properties to only the class.... and define the class.

thehabes:
Hi all, just sharing some research and results from the JSON LD playground today.

We get `jsonld.SyntaxError: Invalid JSON-LD syntax; invalid scoped context.` from this context as it is now.  We have narrowed it down to two issues

1. This context needs a `"iiif_selectors":"http://iiif.io/api/__TODO__/#"` entry
2. Under `ImageAPISelector` in the `@context` array it tries to scope in "http://iiif.io/api/image/#".  This needs to be the context.json and so should be "https://iiif.io/api/image/3/context.json" if you need that context in there.

I am happy to report that we produced a test object which combined this context, the IIIF Presentation API 3 context, and the Goereference Extension context which was processed correctly in the JSON-LD playground.  It raised a couple other questions that need researched but it is a promising result.  Below is what we found out.

[Go Here ](https://store.rerum.io/v1/id/6453edc3907395217254aefe)to see the test object that we brute force embedded the contexts into.
[See It Process](https://json-ld.org/playground/#startTab=tab-compacted&json-ld=https%3A%2F%2Fstore.rerum.io%2Fv1%2Fid%2F6453edc3907395217254aefe&context=%7B%7D) in the JSON-LD playground.

kirschbombe:
I added a new `iiif_selectors` context definition to the context file. For the image API context, all other IIIF contexts use the URI pattern `http://iiif.io/api/image/#`, so I left that for now. I did change the https to http.

I also added a note about the ImageApiSelector default values (v3 vs v2).

source/registry/selectors/context.json
```diff
@@ -0,0 +1,60 @@
...
+  "@context": {
+    "@version":1.1,
+    "iiif_selectors": "http://iiif.io/api/registry/selectors#",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+
+
+    "ImageApiSelector": {
+      "@id": "iiif_selectors:ImageApiSelector",
+      "@context": [
+        "http://iiif.io/api/image/#",
```
mikeapp, line 11: I think this should be defined above as `"iiif_image": "http://iiif.io/api/image/#"` and not listed in this `@context`

zimeon (changes requested):
Minor issue noted with duplicate defn. of `[registry-selectors]` but otherwise looks good

mikeapp (dismissed):

kirschbombe:
The president 3 context is updated to include the selectors context. A big thanks to @thehabes for testing the context and helping me simplify it.

tomcrane (approved):

azaroth42 (approved):

mikeapp (approved):

mixterj (approved):

JulieWinchester (approved):

PR #2225, opened 2023-03-27
Auth 2.0 release

tomcrane:
Preview: https://preview.iiif.io/api/auth-2-milestone/auth/2.0/ 

Changelog: https://preview.iiif.io/api/auth-2-milestone/auth/2.0/change-log/

Issue #2112, opened 2022-03-15
Required use of labels and other UI strings in Auth Services


tomcrane:
From @ksclarke in Slack:

From the properties table at https://iiif.io/api/auth/1.0/#service-description it looks like label is required. But once one gets down to the Kiosk and External patterns we see:

> There is no user interaction before opening the access cookie service URI, and therefore any of the label, header, description and confirmLabel properties are ignored if present.

The JSON examples have label and the other recommended but ignored properties. What's the point of making label required (and the others recommended) if the clients are instructed to ignore it/them? 

`@id` for instance has a description that prescribes use for particular types of services: It is required with the Login, Clickthrough, or Kiosk patterns, in which the client opens the URI in order to obtain an access cookie.

tomcrane:
For Auth 2, the presence of these fields in the examples is now confined to the `interactive` profile example.

Further, in https://github.com/IIIF/api/commit/b0e1101b64a7c5cb0de0e14316b9857191cdf0da, text has been added to the table to qualify use of these properties.

tomcrane:
Included in https://github.com/IIIF/api/pull/2127

zimeon:
Resolved. In Auth 2.0 the additional labels are defined only for the `active` interaction pattern: https://iiif.io/api/auth/2.0/#active-interaction-pattern

Issue #2108, opened 2022-03-01
Support IP-based and related forms of authentication


tomcrane:
Auth 1.0 assumes that the user's relationship with access-controlled resources is decided by credentials carried by cookies.

Even in Auth 1.0 this is not an absolute dependency; the server is basing an access control decision on some aspect of a _simple_ web request, such as that caused by an image tag, or a GET from opening a resource in a new tab. Usually that aspect is the credentials carried by cookies. But it could be any aspect of the request, such as origin, IP range, or some other information, that can be seen by the protected resource, and sometimes also the token service.

The auth interaction pattern doesn't need to change to accommodate this (although it might change for other reasons) - and it could help generalise the applicability of the spec.

A flow based on IP range would look the same to the client, but the probe aspect of the request could return a 200 immediately (where probe includes the current info.json request, see #1290). Or, a _clickthrough_ pattern could show the user something in a clickthrough tab (its "cookie" service), but not set a cookie. The _external_ pattern could allow token acquisition based on any aspect of the request, not just its cookie content.

Other patterns could emerge that involve a "cookie" service that doesn't set a cookie. The "cookie service" could be renamed to... something else. It's still most likely to be cookies, but a change in language (without being too obscure - we still assume cookies are the most likely use) could help people understand different mechanisms, including custom clients (access based on user agent, for example).

If it's IP-based, why is it even in scope for IIIF Auth at all?
 - Because this allows for people who aren't in the permitted IP range to see some useful messages, rather than broken images. IIIF Auth is about managing the user experience.

tomcrane:
Auth call - this is about changing the language of the spec to support more scenarios

tomcrane:
See https://github.com/IIIF/api/commit/b4cae0ac9ed5954895f7d490e9e56a6a48363e0e and later commits.
The spec language in https://github.com/IIIF/api/pull/2127 uses the term "aspect of the request" rather than cookies; the presence of a cookie is an aspect of the request, but so is the origin IP, or the time of day. Similarly the "cookie service" is renamed "access service" - the place you go to acquire the _**aspect of the request**_.

You might acquire this aspect of the request by logging in, and thereby gaining a cookie.
You might already have this aspect of the request (you're coming from an IP address that suggests Australia, or "it's Thursday").
The flow doesn't distinguish between these two scenarios and still takes you to the Access Service.

This may be desirable, to not leak implementation details.

[...]

PR #2234, opened 2023-05-23
first pass at Auth 2 changelog

tomcrane:


zimeon (changes requested):
Overall I think this is looking great, a few nits:
* We are rather inconsistent about reference to other APIs, we have a mix of “IIIF Image API“, “Presentation 3.0“, “Presentation API 3.0”. Looking through prezi 3.0 we are mostly consistent about using “IIIF Whatever API [optional version]” so I suggest we standardize on that.
* I think section “1. Breaking changes” should have an initial first sentence along the lines of: _“The “clickthrough” and “login” interaction patterns have been replaced with a new “active” pattern.”_ because this seems really key.
* In 2.2 the link “See #2112” seems wrong, should this be https://github.com/IIIF/api/issues/2108 ?
* In 2.3 suggest changing “learned the user’s access to IIIF API Image responses” to “learned [about] the user’s [ability] to access to IIIF API Image responses”
* In 2.4 the string “Probe Service response” links to the probe service https://preview.iiif.io/api/auth2-change-log/auth/2.0/#probe-service whereas I think it should link to the response specifically https://preview.iiif.io/api/auth2-change-log/auth/2.0/#52-probe-service-response (or, if not, only “Probe Service” should be linked, not response). Also remove “(see below)” at the end.

zimeon:
Editors' meeting -- Decided to merge this into the https://github.com/IIIF/api/pull/2225 Auth 2.0 branch so that we are working with everything together to check before final release

PR #2248, opened 2023-06-22
Rights/license registry

kirschbombe:
- Updates the existing registry/rights stub page to include the existing CC and RightsStatement.org URIs.
- Adds the registry/rights page to the Registry index page.
- Adds registry/rights to links.md

### DIFF
--- a/source/_includes/links.md
+++ b/source/_includes/links.md
@@ -285,7 +285,9 @@
 [registry-profiles]: {{ site.api_url | absolute_url }}/registry/profiles/
 [registry-timeModes]: {{ site.api_url | absolute_url }}/registry/timeModes/
 [registry-types]: {{ site.api_url | absolute_url }}/registry/types/
-[registry-viewingDirections]: {{ site.api_url | absolute_url }}/registry/viewingDirections/
+[registry-viewingDirections]: {{ site.api_url | absolute_url }}/registry/
+viewingDirections/
+[registry-rights]: {{ site.api_url | absolute_url }}/registry/rights/
 [search-stable-version]: {{ site.api_url | absolute_url }}/search/{{ site.search_api.stable.major }}.{{ site.search_api.stable.minor }}/ "Latest Content Search API Stable Version"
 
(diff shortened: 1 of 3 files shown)


source/_includes/links.md
```diff
@@ -285,7 +285,9 @@
 [registry-profiles]: {{ site.api_url | absolute_url }}/registry/profiles/
 [registry-timeModes]: {{ site.api_url | absolute_url }}/registry/timeModes/
 [registry-types]: {{ site.api_url | absolute_url }}/registry/types/
-[registry-viewingDirections]: {{ site.api_url | absolute_url }}/registry/viewingDirections/
+[registry-viewingDirections]: {{ site.api_url | absolute_url }}/registry/
+viewingDirections/
```
kirschbombe, line 289: I'm not sure how this line break got in here. It's not a new line in my local version

kirschbombe: fixed

source/registry/rights/index.md
```diff
@@ -58,11 +56,10 @@ The process for having a new entry added to this registry is [described here][re
 
 This table summarizes the known rights statements and licenses available for use with the IIIF APIs.
 
-The URIs defined by [Creative Commons][org-cc-licenses] and [RightsStatements.org][org-rs-terms] are not listed here, but should be considered to be included.
```
azaroth42, line 61: This one is important. Either we need to list all of the CC / RSs licenses or not list them and keep this assertion. The URIs in the PR aren't themselves licenses/statements.

azaroth42 (changes requested):
Need to list license URIs or include CC/RSs by the text

source/registry/rights/index.md
kirschbombe: I like the idea of listing these two standards (esp. since we have no other to list at the moment), but could we have a column for each that says "for a full list of the CC license URIs, see...." rather than listing each license/statement URI? Then we don't have to worry about updating if the specific licenses/statements change over time -- and don't have to list them all here when they are listed elsewhere. Or we could list a few of the most common, then send folks to CC and RS for more.

mikeapp (approved):

tomcrane (approved):

tomcrane (approved):

mikeapp (approved):

JulieWinchester (approved):

Issue #2322, opened 2024-11-19
Should partOf of a AnnotationPage be an array or object?


glenrobson:
Original raised by @jptmoore on slack: https://iiif.slack.com/archives/C18FHBKAQ/p1731497378271299?thread_ts=1731491354.710519&cid=C18FHBKAQ

One of the recipes going to the TRC tomorrow is on [Annotation Collections](https://github.com/IIIF/trc/issues/130). In this receipe the partOf of an annotation page looks as follows:

```
"partOf": [
            {
              "id": "https://preview.iiif.io/cookbook/309-annotation-collections/recipe/0309-annotation-collection/anno_coll.json",
              "type": "AnnotationCollection",
              "label": {
                "en": [
                  "Newspaper layout markup"
                ]
              },
              "total": 8,
              "first": "https://preview.iiif.io/cookbook/309-annotation-collections/recipe/0309-annotation-collection/anno_p1.json",
              "last": "https://preview.iiif.io/cookbook/309-annotation-collections/recipe/0309-annotation-collection/anno_p2.json"
            }
          ],
```

and follows the IIIF guidance on `partOf` (https://iiif.io/api/presentation/3.0/#partof):

"The value must be an array of JSON objects. "

but the Web Annotation Model says (https://www.w3.org/TR/annotation-model/#model-35):

"The relationship between the Page and the Annotation Collection that it is part of.
Each Page SHOULD have a exactly 1 partOf relationship, with the value being either the IRI of the Collection or an object with some or all of the Collections properties, including at least its id."

As this is an AnnotationPage linking to a AnnotationCollection, which standard should we follow?

azaroth42:
The challenge:
* ActivityStreams defines `partOf` as functional, meaning per AS: "Properties marked as being "Functional" can have only one value."
* Web Annotation follows that model for pages/collections.
* DC makes no cardinality assertions about isPartOf
* IIIF Prezi3 says it must always be an array

We use `partOf` in places that probably should always be singular, especially annotation-page within annotation-collection -- it makes no sense for the page to be part of multiple collections as they're a side effect of partitioning a single conceptual list. But also in places where multiple is essential, such as the same manifest being partOf multiple collections.

There's no good option, but it seems to come down to:

* Ignore the AS/Anno cardinality and always use an array for consistency (bad for interop outside of IIIF)
* Acknowledge the inconsistency with partOf (bad for developers)
[...]

PR #2325, opened 2024-12-03
Motivation should be Array, not String

### LINKED ISSUE #2324: Motivation should be Array, not String
https://iiif.io/api/content-state/1.0/#53-multiple-targets-for-a-comparison-view

https://github.com/IIIF/api/blob/17629e12c6fa0327b6043bcdffa8c7f804d48edb/source/content-state/1.0/index.md?plain=1#L581

cubap:
Fixes #2324


cubap:
Thanks @markpbaggett for pointing this out

azaroth42:
It's not incorrect for it to be a string, per https://www.w3.org/TR/annotation-model/#motivation-and-purpose 

Related also to #2322

PR #2326, opened 2024-12-05
fix typo

julsraemy:
dereferencable --> dereferenceable


azaroth42:
Good catch. I think we'd need to have a 3.0.1 to merge however.

stweil (approved):

stweil:
PR #2343 mentions some more typos. I expect that several or even most of them are also in 3.0.0.

tomcrane:
Have applied this to Presentation 4.0

PR #2340, opened 2025-04-01
Lights, Audio, UnitValue and associated properties

mikeapp:
Questions:
* `intensity` and `volume` - restrict from use on other resources?
* Note about Spot Audio requiring a transform.  I changed to default to -y direction, like Spot Light.

### DIFF
--- a/source/presentation/4.0/model.md
+++ b/source/presentation/4.0/model.md
@@ -523,7 +523,7 @@ Ambient Light evenly illuminates all objects in the Scene, and does not have a d
 
 Directional Light emits in a specific direction as if it is infinitely far away and the rays produced from it are all parallel. It does not have a specific position.  
 
-The light emits in the negative Y (-y) direction by default, but the orientation of the light can be altered by subsequent transforms.
+The light is emitted in the negative Y (-y) direction by default, but the orientation of the light can be altered by subsequent transforms.
 
 ##### Point Light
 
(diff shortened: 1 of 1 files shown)


JulieWinchester:
I wonder if maybe -Z makes more sense as a default direction, like a camera, rather than -Y like a light? I'm sort of analogizing a spotaudio source to kind of like a directed speaker, which I think might be more natural to have in the same XZ "horizontal" plane by default, pointing "back" at me, rather than pointing "down" at me? Interested in what others think, though.

mikeapp:
I'm assuming that for Audio `source` we'd want to allow at least Timeline, and possibly any Container (i.e., an implicit selection of the audio track from a Canvas or Scene).   I assumed this was covered by "Containers may be treated as content resources for the purposes of annotating on to other Containers."  but it relies on a loose reading of "annotating on to".  This might need to be called out in property definition.

source/presentation/4.0/model.md
```diff
@@ -509,89 +509,94 @@ Properties...
...
-
-
+ 
 ##### Ambient Light
 
 `"type": "AmbientLight"`
 
-Ambient Lights evenly illuminates all objects in the scene, and does not have a direction or position.
-
+Ambient Light evenly illuminates all objects in the Scene, and does not have a direction or position.
```
tomcrane, line 518: Are these "An Ambient Light", "A Point Light", etc? Not sure which reads better.

source/presentation/4.0/model.md
```diff
@@ -1886,6 +1924,18 @@ The value _MUST_ be a string.
 { "viewingDirection": "left-to-right" }
 ```
 
+### volume
+{: #volume}
+
+The volume propery represents the relative volume of an audio source. The `value` of the specified UnitValue represents the desired volume on a linear scale from 0.0 (silence) to 1.0 (maximum volume).  If this property is not specified, then the default volume value is client-dependent.
```
tomcrane, line 1930: typo "property"

tomcrane (approved):

source/presentation/4.0/model.md
mikeapp: I followed other examples in the document.  I'm sure we'll need to review for style later.

Issue #2426, opened 2026-02-16
Requirements for core Annotation properties


azaroth42:
In 4.0, we are importing the annotation properties to IIIF as the 3d work requires a lot of extensions and processing rules.

This means that we need to define the requirements for relationships like `body`, `target`, `motivation`, `source`, and so on. Some are easy, but these four are very inconsistent in the way they're used in Annotation documents.

The consistent route for their values would be:

`body` and `target` **MUST** be an array of JSON objects. Each item **MUST** have the `type` property. 
`source` **MUST** be a JSON object with `id` and `type`.
`motivation` **MUST** be an array of strings.

Following the "if it can ever have multiple values then it is always an array" rule, and similarly that if it can ever be a JSON object then it should always be a JSON object rather than a shortcut to the object's `id` as a string.

However this will break almost all 3.0 Manifests, especially for `target`.

There are several routes forward:

1. We can remain completely lax with the definitions and allow whatever comes (string, object, array of strings and/or objects)
2. We can make the consistent way **SHOULD** with a note that a future version will make it **MUST**
3. We can make it **MUST** with a note to client implementers to please be lenient in accepting other values
4. We can just make a clean break and say **MUST**

PR #2428, opened 2026-02-17
Use Composite instead of Lists

azaroth42:
Per #2426 discussion, change `body` and `target` to be a single JSON object. If there's multiple, then wrap in an aggregate construction (Composite, List, Independents, Choice). 

I think this is significantly cleaner than singleton arrays 99% of the time.

(Thanks to @stephenwf!)

### DIFF
--- a/source/presentation/4.0/index.md
+++ b/source/presentation/4.0/index.md
@@ -1259,7 +1259,7 @@ There are two types of Camera, [`PerspectiveCamera`][prezi-40-model-PerspectiveC
 
 ### Lights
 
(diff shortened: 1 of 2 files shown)


source/presentation/4.0/index.md
```diff
@@ -1845,13 +1859,19 @@ This example is a Manifest with a Canvas that represents two images displayed si
...
+                        {"id": "https://example.org/iiif/presentation/examples/manifest-composite-two-canvases/canvas/c2",
+                          "type": "Canvas"
+                        }
                       },
                     ]
                   }
                 ]
               },
-              "target": "https://example.org/iiif/presentation/examples/manifest-composite-two-canvases/canvas/c1#xywh=0,0,300,300"
+              "target": {}
```
mikeapp: remove `}`

mikeapp (approved):

PR #2444, opened 2026-03-12
Protocol  changes

mixterj:
moved Protocol to the Model and added a reference to the Index

### DIFF
--- a/source/presentation/4.0/index.md
+++ b/source/presentation/4.0/index.md
@@ -1417,58 +1417,9 @@ An image might not be correctly aligned with the Canvas, and require rotation as
 ```
 
 
-
-
-
-
 # Protocol
 
-This section outlines recommendations and requirements related to URIs, HTTP requests, and authentication for IIIF resources.
-
-## URI Recommendations
-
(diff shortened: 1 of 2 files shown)


source/presentation/4.0/index.md
```diff
@@ -1417,58 +1417,9 @@ An image might not be correctly aligned with the Canvas, and require rotation as
...
-Content-Type: application/json
-```
-{: .urltemplate}
-
-The HTTP server _MUST_ follow the [CORS requirements][org-w3c-cors] to enable browser-based clients to retrieve the descriptions. If the server receives a request with one of the content types above in the Accept header, it _SHOULD_ respond with that content type following the rules of [content negotiation][org-rfc-7231-conneg]. Recipes for enabling CORS and conditional Content-Type headers are provided in the [Apache HTTP Server Implementation Notes][notes-apache].
-
-Responses _SHOULD_ be compressed by the server as there are significant performance gains to be made for very repetitive data structures.
-
-
+IIIF uses a variety of standard Web protocals. See the [Prototco](model/#Protocol) for more detail.
```
JulieWinchester:
```suggestion
IIIF uses a variety of standard Web protocols. See [Protocol](model/#Protocol) for more detail.
```

PR #2446, opened 2026-03-12
Add scope as activating annotation shortcut to model

JulieWinchester:


JulieWinchester:
Ok, I've updated the paragraph in response to the comments, it now reads:

> In a Scene, `scope` would typically be applied to a Content Resource or commenting Annotation and would reference a Camera that should be activated and selected to set the correct viewpoint in response to user interaction with that resource. This specification does not establish specific expectations for references to other classes. Clients _MUST_ process `scope` as a short-hand expression of an Annotation where `motivation` is "activating"; where `action` values are `show`, `enable`, and `select`; where the `source` of the activating Annotation `body` is the Resource referenced by `scope`; and where the `target` is the Annotation or Specific Resource on which `scope` is specified. 

Re: the third "where" statement about `body`, I think we have to have the slightly convoluted language around `source` because it specifically is the source, since the `body` of an activating Annotation is always a SpecificResource but the thing referenced by scope in many cases won't be.

PR #2453, opened 2026-03-25
Adding new v4 validator action 

glenrobson:


glenrobson:
Currently failing on the duplicate id check. I need to add body to the list of allowed places for a duplicate id. 

But should the first sub-scene be embedded:

https://preview.iiif.io/api/prezi-4/presentation/4.0/example/uc06_scene_in_scene.json

glenrobson:
I think this is ready to go but I would like someone to confirm the changes I've made to scene in scene make sense:

The first sub scene is embedded in full:

https://github.com/IIIF/api/blob/31421a417e9320e13b8e9e33875f2ebb47dc3f41/source/presentation/4.0/example/uc06_scene_in_scene.json#L21

but the second instance of the scene is only a reference:

https://github.com/IIIF/api/blob/31421a417e9320e13b8e9e33875f2ebb47dc3f41/source/presentation/4.0/example/uc06_scene_in_scene.json#L73

Is that right? It would mean a lot of duplication to duplicated the sub scene twice.

glenrobson:
Update example with a range with behaviour sequence

PR #2460, opened 2026-03-30
Update nesting to be real UC, and move above 3D

azaroth42:
Nesting Change. Please approve/deny first before merging mar30_edits


JulieWinchester:
From editor's meeting April 8th, group is in favor of the reorganization here and it significantly cleans up ordering and language of things like the mentions of nesting in the 3D section, but maybe it would be good if the use case were further explained or maybe even modified? Like, there was some concern that we don't want users to think that canvas-in-canvas nesting is a preferred way to do things that could be handled by just putting two images side by side or using a 2-up viewer, so maybe if we could elaborate on use case (or add additional annotations to each canvas?), remove the concept of "reconstruction," or in some other way match this to a more real-world use case in which someone might use canvas-in-canvas nesting.

mikeapp (approved):
Assuming "merge then improve in another PR"

JulieWinchester (approved):

